PR #1601, opened 2012-01-26
Send confirmation email after_commit to avoid background jobs from trying to send email before record is created

macmartine:
When delegating DeviseMailer to something like Resque, sometimes the email tries to get sent before the record has fully been created. So, changing after_create to after_commit which fixes this problem.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -31,9 +31,9 @@ module Devise
 
       included do
         before_create :generate_confirmation_token, :if => :confirmation_required?
-        after_create  :send_confirmation_instructions, :if => :confirmation_required?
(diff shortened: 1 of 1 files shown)


gingerlime:
I know it's an old one, but just bumped into the same issue as well since all our emails are automatically sent using sidekiq.

Occasionally, there would be a race condition and the mailer will try to access a user id that doesn't exist... I saw this problem explained / described very well on http://www.justinweiss.com/blog/2015/03/10/a-couple-callback-gotchas-and-a-rails-5-fix/ (including how to avoid test issues)

I'm wondering if devise should by default try to use `after_commit` if it can, like @josevalim described, otherwise fallback to `after_create`? This should make things work for both synchronous / asynchronous cases.

PR #2022, opened 2012-08-14
Allow after_sign_out_path_for to use current_user

mikepack:
When `#after_sign_out_path_for` is defined in ApplicationController, it cannot use `current_user` to determine the path. From within `#after_sign_out_path_for`, `current_user` will be the user prior to signing out. As the name of this method implies, this should be called **after** the user has been signed out.

One of my projects does this in the following way:

``` ruby
def after_sign_out_path_for(user)
  if can?(:manage, :users)
    admin_dashboard_path
  else
    root_path
  end
end
```

The `can?` method (of CanCan) returns based on `current_ability` which is derived from `current_user`.

My scenario is as follows:
1. Log in as user A (who can :manage :users)
2. Ghost log in as user B (who can **not** :manage :users)
3. Log out of user B
4. (FAIL) Redirect to root_path instead of admin_dashboard_path

After step 3, the `current_user` should be user A from within `#after_sign_out_path_for`. However, because `#after_sign_out_path_for` is currently being called _before_ the user is signed out ([link](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/sessions_controller.rb#L23-24)), `current_user` is user B.

This regression was introduced in [this commit](https://github.com/plataformatec/devise/commit/9ea724936869e094a10f6c7ee8a718428926d19e).

### DIFF
--- a/app/controllers/devise/sessions_controller.rb
+++ b/app/controllers/devise/sessions_controller.rb
@@ -20,8 +20,8 @@ class Devise::SessionsController < DeviseController
 
   # DELETE /resource/sign_out
   def destroy
-    redirect_path = after_sign_out_path_for(resource_name)
     signed_out = (Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name))
+    redirect_path = after_sign_out_path_for(resource_name)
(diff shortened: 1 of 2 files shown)


Franci200 (commented):
Ddf

Issue #2821, opened 2014-01-04
Do not keep the token in devise initializer file if the application has a config/secrets.yml file


josevalim:
I.e. let's use the same convention as Rails 4.1.

PR #2835, opened 2014-01-13
Add support for Rails 4.1 secrets.

lucasmazza:
This is a first spike for that @josevalim proposed on #2821. Upgrade existing apps will be simple - developers just need to move their keys to the `secrets.yml` - but on fresh apps its required to add the `secret_key` key manually. We could still generate a token inside the initializer as a fallback (so the app can boot without this step) or figure out an automatic way of placing the token on the application configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+### Unreleased
+
+* enhancements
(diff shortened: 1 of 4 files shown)


arthurnn:
:heart:

Issue #2823, opened 2014-01-06
Use inet column type for ip columns when generating migrations for posgresql


adrianpacala:
Inet column type can be set from string (as it uses IPAddr class internally) so no changes in Devise code base should be required. I can provide a PR if it's ok with the maintainers.

Any opinions on this?

PR #3017, opened 2014-05-07
Use inet column type when generating migrations under PostgreSQL

adrianpacala:
Hi, this is a PR for #2823 which I suggested some time ago.

I'm not sure how to properly test the output for postgresql since the test environment is hardcoded to sqlite. Any ideas?

### DIFF
--- a/lib/generators/active_record/devise_generator.rb
+++ b/lib/generators/active_record/devise_generator.rb
@@ -71,7 +71,11 @@ RUBY
 
       def ip_column
         # Padded with spaces so it aligns nicely with the rest of the columns.
-        "%-8s" % (postgresql? ? "inet" : "string")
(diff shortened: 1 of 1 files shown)


ahnbizcad:
When using mysql, what would be the appropriate column type / method to use? 
inet is failing. mysql doesn't seem to support it.

sigvei:
Late to the party here, but was this really a good idea? There is no guarantee that the development environment is identical to the production and test environments, and IMO the generators should stick to the least common multiple. Developers who are sure they will use postgres for both testing, development, staging and production could be adviced to change this to inet in a comment in the generated migration.

I bumped into this as an issue when converting a single-developer development environment with Postgres to a lighter and more replicable sqlite3 environment.

PR #3045, opened 2014-05-28
Set default sender only when not set in the application

jrz:
Only set the default sender if it's not set in the application config.
This reduces duplicate effort when changing the sender. (I've forgotten it in a few projects)

### DIFF
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -21,7 +21,7 @@ Devise.setup do |config|
     config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
   end
 <% else -%>
-  # ActionMailer::Base.default :sender => "email@example.com"
+  #ActionMailer::Base.default({sender: "email@example.com"})
   unless ActionMailer::Base.default.try(:[], :sender)
     config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
   end


halilim:
I'd be interested in this too for the reasons @jrz originally mentioned. Note that in Rails 4 the config key is [`:from` instead of `:sender`](http://guides.rubyonrails.org/v4.0.0/action_mailer_basics.html#example-action-mailer-configuration).

My Rails 4 only suggestion: How about;

- Commenting out that line in the generator,
- Adding a comment pointing to the relevant part of the Rails guides/docs,
- Modifying [`Devise::Mailers::Helpers.mailer_sender`](https://github.com/plataformatec/devise/blob/v3.5.2/lib/devise/mailers/helpers.rb#L50) like this:
  
  ```ruby
  def mailer_sender(mapping, sender = :from)
    default_sender = default_params[sender]
    if default_sender.present?
      default_sender.respond_to?(:to_proc) ? instance_eval(&default_sender) : default_sender
    elsif Devise.mailer_sender.present?
      if Devise.mailer_sender.is_a?(Proc)
        Devise.mailer_sender.call(mapping.name)
      else
        Devise.mailer_sender
      end
    else
      Rails.configuration.try(:action_mailer).try(:default_options).try(:[], :from)
    end
  end
  ```

PR #3126, opened 2014-07-23
Non-hash resource param throws exception

arrtchiu:
Saw some kind of malicious user/bot posting invalid params, which was triggering 5xx errors.

Instead, I think it should be returning kind of 4xx error without an exception, since it's bad data coming from a user.

Haven't patched it yet, but here's a start - a failing test case.


josevalim:
Users can send all sort of bad parameters and make your application crash. I wouldn't worry about handling all of those cases.

PR #3153, opened 2014-08-17
Rails 4.2 compatibility

lucasmazza:
This Pull Request aims to bring back compatibility with Rails 4.2/master since rails/rails@dc3f25c8a5aa64de9225f11498a389a2d31e880a has changed the `@scope` data structure from the router, without losing compatibility with existing versions of Rails where this object is a `Hash`.

In the matter of Rails 4.2, we also have this deprecation warning:

```
DEPRECATION WARNING: defining a route where `to` is a controller without an action is deprecated.  Please change "to: :users/omniauth_callbacks" to "controller: :users/omniauth_callbacks". (called from devise_omniauth_callback at /Users/lucas/code/ptec/devise/lib/devise/rails/routes.rb:436)
```

Changing the router code to use `controller: controllers[:omniauth_callbacks]` leads to the following error:

```
/opt/rubies/2.1.2/lib/ruby/gems/2.1.0/bundler/gems/rails-d5be08347fb7/actionpack/lib/action_dispatch/routing/mapper.rb:263:in `block (2 levels) in check_controller_and_action': 'users/auth/:action' is not a supported controller name. This can lead to potential routing problems. See http://guides.rubyonrails.org/routing.html#specifying-a-controller-to-use (ArgumentError)
```
- [x] Fix router monkeypatch.
- [x] Remove OmniAuth route warning.
- [x] Deal with `respond_with` removal.
- [x] Reorganize the project `Gemfile`s.
- [ ] :shipit:

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -78,7 +78,7 @@ module Devise
         options[:resource_name] = scope
         options[:scope] = "devise.failure"
         options[:default] = [message]
-        options[:authentication_keys] = Devise.authentication_keys.join(", ")
+        options[:authentication_keys] = scope_class.authentication_keys.join(I18n.translate(:"support.array.words_connector"))
         options = i18n_options(options)
 
         I18n.t(:"#{scope}.#{message}", options)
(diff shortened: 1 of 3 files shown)


promiseverma:
replace 
gem 'devise' 
to  
gem 'devise', :git => 'https://github.com/plataformatec/devise.git'

joshdance:
I was still having errors. This SO answer fixed it for me - http://stackoverflow.com/questions/27611947/devise-raises-error-with-rails-4-2-upgrade

amritdeep:
Use devise 3.4.1.

PR #3208, opened 2014-09-15
Modularize controllers

jensljungblad:
This is a proof of concept PR that supports the scenario I asked about on the mailing list: https://groups.google.com/forum/#!topic/plataformatec-devise/1ZPdAPpjGAY

To reiterate:

We have a scenario in a project where we want to to use devise both in the main app, and inside an engine. Using the information found here: https://github.com/plataformatec/devise/wiki/How-To:-Use-devise-inside-a-mountable-engine this works reasonably well:

```ruby
  devise_for :users
  mount Foo::Engine, at: "foo"
end

Foo::Engine.routes.draw do
  devise_for :foo_users, router_name: "foo"
end
```

However, all Devise controllers inherit from whatever is specified in the shared initializer, i.e. ``::ApplicationController``. 

This means code from the main app's ApplicationController is inherited by all the engine DeviseControllers. Code that does things that are not meant to be done in the engine, for instance.

```ruby
class ApplicationController < ActionController::Base
  # code only applicable to the application
end

class Foo::ApplicationController < ActionController::Base
  # code only applicable to the engine
end
```

By extracting the devise controller code into modules, it is possible to choose inclusion over inheritance. That is, we can now do the following:

```ruby
class SessionsController < Devise::SessionsController # Business as usual
end

class Foo::SessionsController < Foo::ApplicationController # Not inheriting from a devise controller
  include Devise::Controllers::Base
  include Devise::Controllers::Sessions
end
```

What do you think about this approach? @lucasmazza @josevalim


josevalim:
Thank you! Sorry this sat here for so long but we are still not comfortable this is the best way to go.

jensljungblad:
Let me know if you ever find a way to solve it!

PR #3218, opened 2014-09-23
use authentication_keys in flash message (fix #3201)

### LINKED ISSUE #3201: wrong error message in case of custom authentication_keys
Setup:

``` ruby
Devise.setup do |config|
  config.authentication_keys = [ :username ]
end
```

Bug:
If the user types in wrong credentials, devise shows the I18n key `'devise.failure.invalid' => 'Invalid email or password.'`. It should be `Invalid credentials.` or `Invalid username and password.`

timoschilling:


lib/devise/failure_app.rb
```diff
@@ -78,6 +78,7 @@ def i18n_message(default = nil)
         options[:resource_name] = scope
         options[:scope] = "devise.failure"
         options[:default] = [message]
+        options[:authentication_keys] = Devise.authentication_keys.join(", ")
```
timoschilling: Sorry, but I don't have the time to work on that

PR #3223, opened 2014-09-25
Add a seconds call to timeout_in calculation

nbibler:
Currently, by simply calling `timeout_in.ago`, a warning is raised when the user provides a native, Numeric type to the timeout configuration ("Using 5.ago is deprecated," for example).

By calling `.seconds` on the configured value, it allows users to provide a native, Numeric to the timeout configuration value, rather than requiring an ActiveSupport::Duration instance, to avoid these warnings.


josevalim:
A pull request that checks for .ago and emits a proper otherwise would be welcome!

PR #3241, opened 2014-10-03
Revert "Removes reloading of routes when eager loading ..."

lucasmazza:
Reverts plataformatec/devise#3195, that proved to be create an regression @ https://github.com/plataformatec/devise/pull/3153#issuecomment-57259282. We need to eager load the routes to ensure that the our mappings and helpers are defined before loading the rest of the app.

### DIFF
--- a/lib/devise/rails.rb
+++ b/lib/devise/rails.rb
@@ -10,6 +10,9 @@ module Devise
       Devise.warden_config = config
     end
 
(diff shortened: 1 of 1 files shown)


csmuc:
Consequently routes are loaded twice at Rails startup, which is not good for Rails startup time if there are many routes. Any other ideas how to avoid this?

PR #3250, opened 2014-10-09
Fix no confirmation is allowed.

### LINKED ISSUE #3244: :confirmable lets login unconfirmed
Hello 

I've found out that `````:confirmable````` may work incorrect.

I have following config
`````config.allow_unconfirmed_access_for = 2.days`````

And it appears that after 2 days it doesn't let user in, but if user request new confirmation email it allows him to login without confirmation. So, in theory it's possible to use unconfirmed account for infinite time, just requesting new confirmation before login.

Maybe a bug. Any suggestions?

Thanks.

JuanitoFatas:
when `allow_unconfirmed_access_for` is set to `0.days`
and `confirmation_sent_at` is any day after today (include today)
the `confirmation_sent_at && confirmation_sent_at.utc >=
self.class.allow_unconfirmed_access_for.ago`
will always return true.

This patch check for 0.days case and return false immediately.

fixes #3244

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -186,6 +186,8 @@ module Devise
         #   confirmation_period_valid?   # will always return true
         #
         def confirmation_period_valid?
+          return false if self.class.allow_unconfirmed_access_for == 0.days
+
           self.class.allow_unconfirmed_access_for.nil? || (confirmation_sent_at && confirmation_sent_at.utc >= self.class.allow_unconfirmed_access_for.ago)
         end
 


josevalim:
The build is still green without this pull request and it includes Ruby 2.2. What are we missing? :)

JuanitoFatas:
Hmmm. Sorry I must be wrong. Also as per https://github.com/plataformatec/devise/issues/3244#issuecomment-74223085, the correct fix of this should be the solution of this pull request. Thanks for checking again.

PR #3251, opened 2014-10-10
Translate authentication_keys

znz:
I think `authentication_keys` should translate in log-in failure messages.


josevalim:
Ping. :)

PR #3254, opened 2014-10-15
Running views generator with only mailer fails

runephilosof:
Either make this work or document it in the generator help message.


lucasmazza:
I managed to write a patch and a more specific test case for this on eb091b8 - sorry for taking a bit longer to come back to this and thanks for the report.

PR #3281, opened 2014-10-30
Unable to reset password, due to false error about 'wrong reset_password_token'

esofomuso:
The line former 127:-
reset_password_token = Devise.token_generator.digest(self, :reset_password_token, original_token)

looks like it was counter productive. It sticks user in a deadlock when they try to reset their password.

The link is clicked with the reset token. But when user enters password information they get an error saying invalid token even though token is still valid and exists. I found out that this line which I am requesting to take out, was generating a new token, then using that new token to fetch for the user, which it wouldn't find and therefore the user will not be able to reset their password, even though they have a valid token.

Thanks.


josevalim:
Yup, you need to update your views!

PR #3304, opened 2014-11-10
Add ActiveSupport::Notification instruments

BM5k:
I've started adding support for ActiveSupport::Notification instruments to devise. Before I get too carried away, I'd like to know if this approach is satisfactory, and if there are any specific places that should emit notifications.

Also, test_unit isn't exactly my cup of tea, so I could use some pointers in how to test these instruments, if that's necessary.

This is still a work in progress.

### DIFF
--- a/lib/devise/models.rb
+++ b/lib/devise/models.rb
@@ -83,7 +83,7 @@ module Devise
 
       devise_modules_hook! do
         include Devise::Models::Authenticatable
-        include Devise::Models::Notifiable
+        include Devise::Models::Instrumentation
 
         selected_modules.each do |m|
           mod = Devise::Models.const_get(m.to_s.classify)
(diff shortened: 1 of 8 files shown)


josevalim:
Ping. :)

BM5k:
Oops! I’ve been super distracted with other stuff and kind of forgot about this. I’ll set some time aside for it this week and see what I can get finished.

BM5k:
It seems like another good time for a sanity check. Am I still headed in the right direction with this?

Does anyone else have thoughts about where instrumentation should occur? I’ve been mainly just reading through the code and guestimating.

ulissesalmeida:
@lucasmazza  @josevalim  Do we have to add any new instrumentations? It will launch after `4.0.0` release right?

PR #3347, opened 2014-11-28
Prevent CookieOverflow errors for large urls

danielfone:
This occurs if an unauthenticated user requests an excessively long url. This was a problem I encountered with large forms that were submitted via GET and user sessions timing out. Granted, it's an edge case, but the fix is quite simple.

Thoughts?

### DIFF
--- a/test/integration/authenticatable_test.rb
+++ b/test/integration/authenticatable_test.rb
@@ -306,7 +306,7 @@ class AuthenticationRedirectTest < ActionDispatch::IntegrationTest
   end
 
   test 'does not store return_to if it would cause a CookieOverflow error' do
-    get users_path, x: 'x' * ActionDispatch::Cookies::MAX_COOKIE_SIZE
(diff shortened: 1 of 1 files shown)


gaganawhad:
Thanks, @danielfone, for your work on this, and @josevalim for maintaining this project. I appreciate you guys' work!

@josevalim can @danielfone's solution be improved at all? It seems like a rails app implementing devise would 5xx with `ActionDispatch::Cookies::CookieOverflow` if anyone submits at any random path that's too long for the cookie store. (Let me know if I am missing anything!)

gaganawhad:
@danielfone Interesting! That makes sense. I am getting that on all endpoints, even the ones that don't require authentication. I'll investigate further. Maybe something with how we have configured devise. 

Thanks for your help 👍

marckohlbrugge:
The suggested monkeypatch by @danielfone doesn't seem to work anymore.

Does anyone have an updated version?

PR #3365, opened 2014-12-10
Add support for KeyGenerator iterations variable (resolves #3363)

### LINKED ISSUE #3363: Add support for setting KeyGenerator iterations
I've noticed during tests, that Devise is taking long time for generating new confirmation token for the first time (took around 20 seconds), then it was cached. Not sure why there is no way to define it except redefining whole Devise.token_generator

This pull request adds an option to configure number of iterations for generating Tokens (with default staying the same as previous).

koleksiuk:


josevalim:
Thank you, the plan is to rely on the Rails generator soon enough.

PR #3405, opened 2015-01-07
Update find_message to accept scope

Dinuz:
This should still keep the devise lookup in the case that a customed scope is not passed as option, but if instead the custom scope is passed, then the find_message method will use it.

This is kind of useful, if i don't want overwrite the devise locale, and use different locale files, but keeping still the fallback of my devise locale.

### DIFF
--- a/app/controllers/devise_controller.rb
+++ b/app/controllers/devise_controller.rb
@@ -167,7 +167,7 @@ MESSAGE
 
   # Get message for given
   def find_message(kind, options = {})
-    options[:scope] = "devise.#{controller_name}" unless options[:scope]
+    options[:scope] ||= "devise.#{controller_name}"
     options[:default] = Array(options[:default]).unshift(kind.to_sym)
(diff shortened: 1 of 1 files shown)


app/controllers/devise_controller.rb
```diff
@@ -167,7 +167,7 @@ def devise_i18n_options(options)
 
   # Get message for given
   def find_message(kind, options = {})
-    options[:scope] = "devise.#{controller_name}"
+    options[:scope] = "devise.#{controller_name}" unless options[:scope]
```
josevalim: Maybe we should use `options[:scope] ||=` then? :)

Dinuz: @josevalim you are right that is more elegant:)

I am always kind of scare in using the monad style (there are different opinion of what really ||= means, but in this case there is no possibility of been wrong):

I am going to change the pull request and resubmit:)

josevalim:

PR #3407, opened 2015-01-07
Use Devise translations when inheriting from core controllers

### LINKED ISSUE #3367: Overriding a controller shouldn't mess with Devise translation rules
Devise's behavior for an attempt to load the sign-in path while already authenticated is to forward the request to the main (or configured) page. But when the session times out in our application we redirect to a static page explaining how to re-authenticate (this is done in a third-party application) rather than redirecting to the log-in page. So, if the user navigates to the log-in page he will be redirected to the main page since the user is authenticated but then it will redirect again to the timed-out page since the session timed out. In order to avoid this behavior that some users have complained we decided to always show the log-in page to the user when requested instead of redirecting it when authenticated by skipping one of the filters for SessionsController in a inherited controller from our own. This problem happened in a moment when there was a problem with the third-party authenticator and they asked us to allow users to use a shared account until they fixed the authenticator. I've written about this before in the mailing list:

https://groups.google.com/forum/#!topic/plataformatec-devise/w5SJcBGI_Ec

As you can see in the solution we used another controller inheriting from Devise::SessionsController.

Now I noted that some translations are missing in some flash messages set by this new inherited controller. Then I found how `find_message` works:

https://github.com/plataformatec/devise/blob/master/app/controllers/devise_controller.rb#L163

It uses the controller's name for building the translation scope. It would help already if that line was replaced by another method like "devise_scope" so that we could easily override it in the overridden controller but it would be even better if overriding a Devise's controller worked more seamless without having to worry about translations at all.

Maybe a guide on the Wiki explaining how to override Devise's controller would be helpful. Maybe the scope could be passed as a parameter to find_message so that overriding a Devise's controller wouldn't have such side effects.

What do you think?

rosenfeld:
closes #3367

I will now start working in the tests but you can already review the changes. I've tested it with my application manually already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
   * `RegistrationsController#new` and `SessionsController#new` now yields the
     current resource (by @mtarnovan, @deivid-rodriguez)
   * Password length validation is now limited to 72 characters for newer apps (by @lleger)
+  * Controllers inheriting from any Devise core controller will now use appropriate translations.
+    The i18n scope can be overridden in `translation_scope`.
 
 ### 3.4.1 - 2014-10-29
 
--- a/app/controllers/devise/confirmations_controller.rb
+++ b/app/controllers/devise/confirmations_controller.rb
@@ -44,4 +44,8 @@ class Devise::ConfirmationsController < DeviseController
         new_session_path(resource_name)
       end
     end
+
+    def translation_scope
+      'devise.confirmations'
+    end
 end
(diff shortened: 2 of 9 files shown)


rosenfeld:
@josevalim I've found out this is handled by `mocha` and tried this approach in a new test file. Please let me know if there's a better way to test this.

lucasmazza:
@rosenfeld can you please rebase your branch? Looks like we got some conflicts from #3405 here.

rosenfeld:
@lucasmazza done

lucasmazza:

rosenfeld:
Thanks :-)

PR #3410, opened 2015-01-09
Update Timestamps Syntax

Dinuz:
@josevalim this should be update in order to follow the rails directives:)

What do you think?


josevalim:
We still need to keep it explicit since we support multiple rails versions. :)

Dinuz:
I understand:)
But the rails message is pretty clear, this is a deprecate thing. I am sure you know better than me when will be the right time to change it.
I saw for the first time this message with rails 4.2, and considering that other things are been updated, I thought this was supposed to be included too!
Do you think this can break other rails apps?

josevalim:
Are you getting a warning? In earlier versions, the default was `null: true`, that's why we had to be explicit here. :S

PR #3417, opened 2015-01-13
add mm.yml

dreamingblackcat:
added translations for Burmese language(Myanmar country)


carlosantoniodasilva:
Thank you, but we do not include translations directly on the devise repo. There are, however, a couple of external solutions/links that you can look at on the wiki: https://github.com/plataformatec/devise/wiki/I18n.

Hope that helps.

dreamingblackcat:
Oh thanks. So basically I just need to put it in a gist and link to it in wiki?

carlosantoniodasilva:
That's one way indeed, another is to try a pull request to https://github.com/tigrish/devise-i18n (or you could do both :)

PR #3425, opened 2015-01-16
get rid of config.serve_static_assets deprecation warning from production.rb

kbullaughey:
apply the same change to production.rb that was put into test.rb to get rid of the deprecation warning related to the config.serve_static_assets -> config.serve_static_files in Rails 4.2.0


josevalim:

Issue #3429, opened 2015-01-18
Confirmation token invalid shown if used twice misleading


ronalchn:
I just had a new user tell me that it wasn't letting him confirm his email.

When I checked, the email was already confirmed. Perhaps the page was loaded twice or something, and after the first use, the token is no longer valid.

I suggest that after confirming, the token is left in the database, so that the page can tell the user that it has already been confirmed.

I believe no extra DB columns are required, because the confirmed_at field is used to check whether an account is confirmed, and unconfirmed_email is used for whether any reconfirmation is required.

josevalim:
I believe this has been fixed in latest Devise versions. Which Devise version are you using?

ronalchn:
devise (3.2.2)

josevalim:
Ok, I was wrong, it seems we do not have this feature. Its addition is probably welcome!

ronalchn:
Just read some of the code. Would taking out the line setting the token to nil make it work? Once set to nil, the token cannot be looked up anymore. So you never check "pending_any_confirmation" again.

```ruby
      def confirm!(args={})
        pending_any_confirmation do
          ...
          self.confirmation_token = nil // take this line out???
          ...
        end
      end

        def pending_any_confirmation
          if (!confirmed? || pending_reconfirmation?)
            yield
          else
            self.errors.add(:email, :already_confirmed)
            false
          end
        end
```

josevalim:
That sounds like it! And a test case too. :)

PR #3432, opened 2015-01-19
Keep used confirmation tokens for more user friendly error message (issue #3429)

ronalchn:
Removes line setting token to nil as mentioned in https://github.com/plataformatec/devise/issues/3429.

By saving it, a message (self.errors.add(:email, :already_confirmed)) can be shown to the user. If set to nil, the potentially misleading message "Confirmation token is invalid" is given. In such a case, if the user does not log in, they may never know that their account's email has already been confirmed.

Test cases: 1 added (checks for :already_confirmed error), 1 removed (checks token set to nil), 3 changed.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -64,7 +64,6 @@ module Devise
             return false
           end
 
-          self.confirmation_token = nil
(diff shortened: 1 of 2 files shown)


josevalim:

PR #3443, opened 2015-01-30
Added validation check to registrations#create

wazery:
I have added a validation check before saving the resource, because in my applications I noticed that there are huge ID gapes that happens regularly because some users not passing the validation checks and the save is called in which it increments the ID but the validation failure rolls back the transaction leaving the incremented ID. Please review that scenario and let me know if my PR can be merged.

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -14,22 +14,24 @@ class Devise::RegistrationsController < DeviseController
   def create
     build_resource(sign_up_params)
 
-    resource.save
-    yield resource if block_given?
-    if resource.persisted?
-      if resource.active_for_authentication?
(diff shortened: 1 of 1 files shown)


josevalim:
save should validate your resource. We should not need to call `valid?` directly.

josevalim:
And thanks for the pull request! :D Which ORM are you using btw?

wazery:
ActiveRecord

josevalim:
Yup, so save should definitely do it, we
Shouldn't check it ourselves!


-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Lead Developer*

wazery:
So any idea why the ID gaps happen?

josevalim:
No, I don't know. But I wouldn't worry about it either. It may be a callback aborting the operation?

PR #3447, opened 2015-02-03
removing deprecation warning from rails 5

lucianosousa:
removing warning from gererated migrations

```
DEPRECATION WARNING: `#timestamp` was called without specifying an option for `null`. In Rails 5, this behavior will change to `null: false`. You should manually specify `null: true` to prevent the behavior of your existing migrations from changing. (called from block in change at /Users/luciano/Dropbox/projects/Luciano-Sousa/db/migrate/20150203012635_devise_create_users.rb:36)

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,7 +460,7 @@ Devise allows you to set up as many Devise models as you want. If you want to ha
 create_table :admins do |t|
   t.string :email
   t.string :encrypted_password
-  t.timestamps, null: false
+  t.timestamps null: false
 end
 
 # Inside your Admin model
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -460,7 +460,7 @@ Devise allows you to set up as many Devise models as you want. If you want to ha
 create_table :admins do |t|
   t.string :email
   t.string :encrypted_password
-  t.timestamps
+  t.timestamps, null: false
```
rafaelfranca: I think the comma is not needed

lucianosousa: Yes, it is!
Improved. Tks for the tip

PR #3449, opened 2015-02-04
Add comments re: secret_key on Rails 4+ apps

gaganawhad:
Adding a couple comments in initializer that explain that Devise will use `secret_key_base` on Rails 4+ applications as its `secret_key` by default.


lucasmazza:
Thanks!

PR #3451, opened 2015-02-05
Fixed bug in config.secret_key code in devise.rb

redbar0n:
The <% tags are not needed, and would cause a bug:

/.rvm/gems/ruby-2.0.0-p247/gems/activesupport-3.2.13/lib/active_support/dependencies.rb:245:in `load': /Workspace/projectname/config/initializers/devise.rb:9: syntax error, unexpected '<' (SyntaxError)
<% if rails_4? -%>
 ^
/projectname/config/initializers/devise.rb:9: syntax error, unexpected tFID, expecting keyword_end
<% if rails_4? -%>
              ^

Also, rails_4 is in the  `devise/lib/generators/devise/install_generator.rb` file, which is not automatically included in a project when using the latest devise gem. So people will probably not have this file. Therefore it's better to use `Rails::VERSION::MAJOR == 4` directly instead.

PS: I'm on rails 3.2.13 and devise 3.4.1.

### DIFF
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -6,11 +6,11 @@ Devise.setup do |config|
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` on Rails 4+ applications as its `secret_key`
   # by default. You can change it below and use your own secret key.
-<% if rails_4? -%>
-  # config.secret_key = '<%= SecureRandom.hex(64) %>'
-<% else -%>
-  config.secret_key = '<%= SecureRandom.hex(64) %>'
-<% end -%>
+  if Rails::VERSION::MAJOR == 4
+    # config.secret_key = '<%= SecureRandom.hex(64) %>'
+  else
+    config.secret_key = '<%= SecureRandom.hex(64) %>'
+  end
 
   # ==> Mailer Configuration
   # Configure the e-mail address which will be shown in Devise::Mailer,


josevalim:
There is something definitely weird here. The template is meant to be evaluated inside the generator. So `<%` should be removed and `rails_4` to be available.  How are you getting this error? How can we reproduce it?

lucasmazza:
If you copy the file from the repo directly to your app you will get this error. You should use the generator instead.

redbar0n:
Yeah, I just copied the file from the repo, since I didn't have the latest file in my app, even though I was using the latest devise version. I didn't use any generator to generate that file, since I already had some config from a previous version that I wanted to preserve, so I just manually diffed the files instead. I presume any generator will overwrite previous config?

Maybe this is an issue when upgrading?

rafaelfranca:
We recommend you to use the generator. If you want to copy the file from repository you should make sure you are coping the code that was generated not the template file.

> I presume any generator will overwrite previous config?

Yes, it will, but you can do the diff using git and put your old configuration that you still need.

redbar0n:
Ok, thanks for the suggestion. You might want to update the docs with this advice, so other people don't unassumingly run into the same problem.

rafaelfranca:
Could you point the doc that advice people to copy the template directly in the application? It should be updated to not give this advice.

redbar0n:
No such advise was in the doc. And no advice is given on the procedure to upgrade Devise either. Just for the first install. It seems reasonable that there should be a section about upgrading.

PR #3452, opened 2015-02-05
workaround for mountable Engines with Rails 4.2

jaume-prat:
By Using devise on mountable Engines with Rails 4.2, the scope_url method from FailureApp will return a wrong url/path. Concrete in this line: https://github.com/plataformatec/devise/blob/a93edc72fd9f6cc5839dd74107b215a81c16dc37/lib/devise/failure_app.rb#L119

By setting the script_name option to nil, it overrides the engine's prefix set on the routes.rb from the main_app. Example:

````ruby
  mount Admin::Engine, at: 'admin'
end
`````

However until Rails 4.1 the returned path was correct. After this commit: https://github.com/rails/rails/commit/33d6e3be623d4dbcec077e62699272cbdcd23f37 by overriding the script_name option (it should be the engine_name), action_dispatch returns the path without the engine prefix and this ist wrong

In FailureApp this condition is always "true", although the option is not set:
**Rails.application.config.respond_to?(:relative_url_root))** # true
````ruby
# (...)
def scope_url
 # (...)
 opts[:script_name] = (config.relative_url_root if config.respond_to?(:relative_url_root))
 #(...)
end
````

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -111,7 +111,10 @@ module Devise
       opts[:format] = request_format unless skip_format?
 
       config = Rails.application.config
-      opts[:script_name] = (config.relative_url_root if config.respond_to?(:relative_url_root))
+
+      if config.respond_to?(:relative_url_root) && config.relative_url_root.present?
+        opts[:script_name] = config.relative_url_root
+      end
 
       context = send(Devise.available_router_name)
 


maeve:
We have run into this issue as well when mounting devise inside an isolated engine under Rails 4.2.0, and would appreciate this fix being merged into devise (in the meantime, we've registered a custom failure app just to override the `scope_url` method).

Not to be overly nitpicky, but the commit that introduced the changed behavior in rails was actually rails/rails@8cbcd19d7079db1b5df6ddb3b813fea57cd5cc38, which introduced an explicit check for the presence of a `:script_name` key in the options hash, instead of just testing for the value.

josevalim:
Great, thank you!

jaimerson:
When is the release including this fix planned to happen? I need to use devise on a 4.2 rails mountable engine, and I can't use github dependencies in the gemspec =/ (sorry for bringing the issue back from the dead)

PR #3462, opened 2015-02-11
Normalise Sing-in/Log-in language

whmii:
A simple edit. One of the session links refers to “Log in” when everything else refers to “Sign in”


josevalim:
We need to update the remaining of the docs. We have lately decided to rename it to Log in because Sign up and Sign in was confusing to users. :) Thank you!

PR #3465, opened 2015-02-13
Add warning about logged tokens to the README

mattbrictson:
Rails now defaults to `config.log_level = :debug` in production. Since the entire contents of outgoing emails are logged at the debug level, this means plaintext password reset tokens are exposed in log files. Active Job also logs all arguments to enqueued jobs at the info level, so `deliver_later` will leak tokens as well, even if the log level is set to `:info`.

I suggest updating the Devise README to make developers aware of the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,20 @@ def send_devise_notification(notification, *args)
 end
 ```
 
+### Password reset tokens and Rails logs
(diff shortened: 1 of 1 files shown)


josevalim:

Issue #3466, opened 2015-02-13
Automatically logging a user in after a password reset


borski:
It's fine for this to remain the default, but there should be a configurable option to turn this off. For some, this is a security issue, and they end up having to override the controller and simply remove the one line. Is there a reason this isn't a configurable option? :)

josevalim:
Users that want to customize this behaviour can just provide their own controllers. Devise already has too many options and we cannot provide an option around every behaviour that an action goes about. :)

borski:
josevalim, while normally I would agree with you, I would pose the argument that options for secure defaults should be prioritized over other options. Given that this can be insecure (for example, if using two-factor-authentication) I would argue this is an important option to have. Thoughts?

josevalim:
Good point, I forgot about two-factor auth. A pull request would be welcome! :D Thank you!

borski:
No problem - we wrote the gem and are about to modify the readme with [this](https://github.com/tinfoil/devise-two-factor/pull/20/files) but it's ugly and shouldn't be necessary, which was why this was important to me. :)

Mind re-opening the issue so I can be reminded to pull req?

borski:
Pull req merged, so closing this issue now.

PR #3467, opened 2015-02-13
Added an option to not automatically sign in a user after a password res...

borski:
...et. This is useful for cases where additional strategies might be needed (such as two-factor authentication, e.g.), or generally if it is considered a security risk to automatically log in a user after a password is reset.

This was discussed in #3466

### DIFF
--- a/app/controllers/devise/passwords_controller.rb
+++ b/app/controllers/devise/passwords_controller.rb
@@ -34,10 +34,15 @@ class Devise::PasswordsController < DeviseController
 
     if resource.errors.empty?
       resource.unlock_access! if unlockable?(resource)
-      flash_message = resource.active_for_authentication? ? :updated : :updated_not_active
-      set_flash_message(:notice, flash_message) if is_flashing_format?
-      sign_in(resource_name, resource)
-      respond_with resource, location: after_resetting_password_path_for(resource)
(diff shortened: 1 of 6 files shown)


app/controllers/devise/passwords_controller.rb
```diff
@@ -34,10 +34,15 @@ def update
...
-      sign_in(resource_name, resource)
-      respond_with resource, location: after_resetting_password_path_for(resource)
+      if Devise.sign_in_after_reset_password
+        flash_message = resource.active_for_authentication? ? :updated : :updated_not_active
+        set_flash_message(:notice, flash_message) if is_flashing_format?
+        sign_in(resource_name, resource)
+        respond_with resource, location: after_resetting_password_path_for(resource)
+      else
+        set_flash_message(:notice, :updated_not_active) if is_flashing_format?
+        redirect_to new_session_path(resource_name)
```
josevalim: I think you still want respond_with resource, just change the location, no?

josevalim:
Thank you, this looks great, I have added just one tiny comment!

borski:
Fixed. Not sure why I made it a redirect anyway. :)

josevalim:
Waiting for travis and then merging, thank you!

josevalim:
:heart:

Issue #3457, opened 2015-02-07
BUG: Resend confirmation page shows old email address


redbar0n:
If the user does not click the confirmation link within the confirm_within time period, he is taken to a resend confirmation page. On that page, actually the old email address is shown, and not the new one.

Here's a screenshot:
![screenshot 2015-02-07 15 56 14](https://cloud.githubusercontent.com/assets/459282/6092494/e9fd15fa-aee1-11e4-9e91-048f55b381c7.png)

It is a potential security risk, since it will leak the user's old email address to any new user that might have gotten the confirmation email when the original user was trying to change his email (and did a typo).

Besides, the error message devise shows makes no sense in that context, because it presumes the new email is the one being shown.

josevalim:
Thanks. Which Devise version? Also, are you using custom views?

redbar0n:
Never mind.. I just found out that the error was in a custom view we had used. To my defense I didn't code it. I will double double check before posting next time. Thanks for quick reply though. For future reference:

We had used:

    <div class="form-inputs">
      <%= f.input :email, :required => true %>
    </div>

Instead of the correct one:

    <div class="form-inputs">
      <%= f.input :unconfirmed_email, :required => true %>
    </div>

redbar0n:
Actually, I just found that the devise error message isn't being shown when I changed to displaying :unconfirmed_email instead of :email . Any ideas why?

redbar0n:
And actually, I just discovered the original bug persisted even after deleting our custom view, so I guess it is in fact a bug in devise.

redbar0n:
My current workaround, setting the email field explicitly in code, and disallowing the user from editing it:

```ruby

<%= simple_form_for(resource, :as => resource_name, :url => confirmation_path(resource_name), :html => { :method => :post }) do |f| %>
  <%= f.error_notification %>

  <div class="form-inputs">
    <%= f.input :email, :required => true, readonly: true, input_html: { value: resource.unconfirmed_email } %>
  </div>

  <div class="form-actions">
    <%= f.button :submit, t("devise.confirmations.new.resend_confirmation_instructions") %>
  </div>
<% end %>
```

```css
.form-inputs .readonly {
	color: grey;
}
```

PS: I'm using the simple_form gem.

redbar0n:
I'm on devise version:     devise (3.4.1)

josevalim:
[...]

PR #3468, opened 2015-02-13
Fix potential security leak in email reconfirmation flow

redbar0n:
Ref. bug/issue: https://github.com/plataformatec/devise/issues/3457

If the user does not click the confirmation link within the confirm_within time period, he is taken to a resend confirmation page. On that page, actually the old email address is shown, and not the new one.

It is a potential security risk, since it will leak the user's old email address to any new user that might have gotten the confirmation email when the original user was trying to change his email (and did a typo).

`:autofocus` removed, and `:readonly` added, since the user shouldn't be able to change the confirmation email address, when asking for a new confirmation email. Allowing the user to change the email address is a another potential security risk:
- A malicious user getting hold of an old confirmation link could change the email to his own, and resend confirmation instructions to that instead.

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -5,7 +5,7 @@
 
   <div class="field">
     <%= f.label :email %><br />
-    <%= f.email_field :email, autofocus: true %>
+    <%= f.email_field :email, required: true, readonly: true, input_html: { value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) } %> 
   </div>
 
   <div class="actions">


app/views/devise/confirmations/new.html.erb
```diff
@@ -5,7 +5,7 @@
 
   <div class="field">
     <%= f.label :email %><br />
-    <%= f.email_field :email, autofocus: true %>
+    <%= f.email_field :email, required: true, readonly: true, input_html: { value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) } %>
```
josevalim: I think it shuold  be:

```
<%= f.email_field :email, autofocus: true, value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) %>
```

If you don't want to amend, just tell me and I will go ahead and do the change myself and push!

redbar0n: That seems to work as well, but see my notes in the commit description on why :autofocus should be removed, and :readonly added. I also presume :required is nice since it will HTML5 validate the form client-side, so that it's sure to contain an email before submitting.

josevalim: Well, we don't do that for any of the other forms I think. So if we are doing those changes, it needs to be a more complete overhaul (or users can just use simple form :D).

PR #3480, opened 2015-02-17
Remove spaces inside square brackets

Kosmas:


carlosantoniodasilva:
Thanks for your contribution @Kosmas, but I feel this is more a cosmetic change so I'm gonna give it a no go.

Kosmas:
@carlosantoniodasilva yes you are right that is a small style change.

But the reason for this PR, is that every time I (and suppose other people too), use a code review tool, I get warnings for the spaces, and I have to remove them manually.
See here for the reason the warnings come up: https://github.com/bbatsov/ruby-style-guide#no-spaces-braces

So by making this change, I believe many people would be happier if these warnings don't come up.

And further than this, inside the same file there are places that square brackets are used without spaces. So it would be more consistent too.

carlosantoniodasilva:
I'm sorry for the delay @Kosmas.

These code review tools make me somewhat :disappointed:. In any case, there are lots of other places in the devise source code that have the space convention, so if we'd ever change it, it'd have to be global, in a single commit.

I might take care of that in the future, thanks for your contribution.

Kosmas:
@carlosantoniodasilva no worries. 

And it's only about convention not religion ;-)

These two files are the ones that people see after using devise, since they are generated by devise. So maybe this is a special case.

But will try and have a look at the whole code and do a different PR, if you prefer that.

Thanks for replying

carlosantoniodasilva:
@Kosmas I merged a squashed version of your commits and did a final pass: 18b6064d74726147eccd69b24812000074261bbb, hopefully that should keep everyone happy now. Thanks!

Kosmas:
@carlosantoniodasilva that is great ! Thank you very much for doing that. Sorry I did not get the time to do it myself. Thanks again!

PR #3493, opened 2015-02-24
allow Devise::Models::Authenticatable to be loaded before Rails

betesh:
I am using some Devise modules in a Sinatra app.  Because Devise::Models::Authenticatable expects Rails to be defined, I need to add the following hack:

    require 'active_model'
    module Rails
      VERSION ||= ActiveModel::VERSION # Needed in order to load devise/authenticatable
    end

However, while this solves the immediate problem of making it possible to `include Devise::Models::Authenticatable`, it creates a new set of problems--other gems ([annotate](https://github.com/ctran/annotate_models/blob/v2.6.5/lib/annotate.rb#L130) and recent versions of [paperclip](https://github.com/thoughtbot/paperclip/issues/1739), [delayed_job](https://github.com/collectiveidea/delayed_job/pull/634) and [ActiveRecord](https://github.com/rails/rails/blob/v4.2.0/activerecord/lib/active_record/connection_handling.rb#L3)) use `defined?(Rails)` to determine whether some component of Rails is loaded.

Since the Rails version is being checked to determine whether `serializable_hash` is defined, it would be more appropriate to check the version of ActiveModel anyway, since that is where `serializable_hash` lives.

### DIFF
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -1,3 +1,4 @@
+require 'active_model/version'
 require 'devise/hooks/activatable'
 require 'devise/hooks/csrf_cleaner'
 
@@ -97,7 +98,7 @@ module Devise
 
       array = %w(serializable_hash)
       # to_xml does not call serializable_hash on 3.1
-      array << "to_xml" if Rails::VERSION::STRING[0,3] == "3.1"
+      array << "to_xml" if ActiveModel::VERSION::STRING[0,3] == "3.1"
 
       array.each do |method|
         class_eval <<-RUBY, __FILE__, __LINE__


betesh:
By the way, there are a lot of other tweaks that would make it easier to use devise with Sinatra.  Let me know if you'd consider a PR to make this a little easier.  I'd include documentation and keep it fully backwards-compatible.

georgeguimaraes:
Your suggestion about the version check makes sense. I'm going to merge it.

However, we don't support using Devise on top of Sinatra since almost all devise features are Rails specific.

georgeguimaraes:
The build is failing. Can you look at it? 
https://travis-ci.org/plataformatec/devise/jobs/52027539

betesh:
Fixed.  I had to explicitly `require "active_model/version"` for Rails 4.  Thank you for agreeing to merge so quickly!

> almost all devise features are Rails specific.

I disagree.  Several Devise::Model modules have been very helpful in implementing secure authentication in Sinatra.  This ticket is probably the wrong place for this discussion, so where would be an appropriate place to continue it?

rafaelfranca:
I know you can use some Devise modules in general Ruby applications not
just Rails, but I think we as maintainers, don't want to support anything
besides Rails right now.

betesh:
OK Fair enough.

betesh:
You're still planning to merge this, correct?

rafaelfranca:
This one yes

PR #3502, opened 2015-03-01
remove use of hide_action in favor of protected

davetron5000:
Rails 5 will [not have `hide_action` any longer](https://github.com/rails/rails/pull/18371/files), as the Rails convention is to not expose private or protected methods as actions, thus obviating the need for `hide_action`.

Presumably, there is code inheriting from `DeviseController` that is calling these helpers, so they cannot be private, so `protected` seems to be the only way to get Devise working with Rails 5.  At least to me :)

I also considered back-porting `hide_action` into this gem, but that felt like the wrong way to go.


josevalim:

davetron5000:
Wow that was fast. Thanks!

PR #3503, opened 2015-03-03
Remove redundant `remember_created_at.nil?`

arnvald:
`remember_expired?` already calls this method


carlosantoniodasilva:
Thanks.

PR #3513, opened 2015-03-13
remove deprecated use of prepend_before_filter

davetron5000:
replace all `prepend_before_filter`s with `prepend_before_action`, as `prepend_before_filter` is deprecated in Rails 5 and will be removed in Rails 5.1.

I believe this will make this gem require Rails 4.  If we want this to be compatible with versions of Rails before `prepend_before_action` was added, what is the best way to do that?  Check `Rails.version`?

### DIFF
--- a/app/controllers/devise/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise/omniauth_callbacks_controller.rb
@@ -1,5 +1,5 @@
 class Devise::OmniauthCallbacksController < DeviseController
-  prepend_before_filter { request.env["devise.skip_timeout"] = true }
+  prepend_before_action { request.env["devise.skip_timeout"] = true }
 
   def passthru
     render status: 404, text: "Not found. Authentication passthru."
--- a/app/controllers/devise/passwords_controller.rb
+++ b/app/controllers/devise/passwords_controller.rb
@@ -1,5 +1,5 @@
 class Devise::PasswordsController < DeviseController
-  prepend_before_filter :require_no_authentication
+  prepend_before_action :require_no_authentication
   # Render the #edit only if coming from a reset password email link
   append_before_filter :assert_reset_token_passed, only: :edit
 
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -1,6 +1,6 @@
 class Devise::RegistrationsController < DeviseController
-  prepend_before_filter :require_no_authentication, only: [ :new, :create, :cancel ]
-  prepend_before_filter :authenticate_scope!, only: [:edit, :update, :destroy]
+  prepend_before_action :require_no_authentication, only: [ :new, :create, :cancel ]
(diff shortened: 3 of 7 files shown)


carlosantoniodasilva:
Devise is still compatible with Rails 3.2, so this change won't work with it, unfortunately.

I'm not sure yet we're going to target Rails 5 right now, it seems too soon for doing so, so I'd have to hold on on your change. Are you already trying Devise with Rails master?

davetron5000:
Yeah, I'm working on a book about Rails and am keeping my example app running on both 4.2 and master, since I'm expecting Rails 5 to ship about when my book is ready, so I figured anything useful I find can be given as feedback.  Understood if we want devise to keep working with Rails 3.2 as long as we can.  This aspect of Rails won't change until 5.1 (at the earliest).

carlosantoniodasilva:
Right, I see. Well thanks for testing Devise with Rails master then :smile:

I believe we might remove 3.2 support when Rails 5 ships, but as I said, it seems to soon for doing so, and I'm not sure adding conditionals to the code right now is the best solution. Maybe we can provide a small shim for devise controllers with the *action we use. Or, unless the deprecations are super annoying, we can keep it like that for now. Wdyt?

davetron5000:
Yeah, I'm fine with not merging this for a while.  Not sure if you'd rather have this hanging around open, or re-open it later or what.  Just happened to have it handy, so I'm game for however you want to handle it.

teknull:
So those of us that are running apps still in 3.2 will have to stick with a gem version compatible?  No more support for 3.2?  I have several production apps in 3.2.x and it's going to be a big overhaul to bring them to 4.2 much less 5.

teknull:
I'm assuming if 3.2 Rails users wanted the latest gem once this merge happens, we could probably monkey patch ourselves?

rafaelfranca:
> So those of us that are running apps still in 3.2 will have to stick with a gem version compatible? No more support for 3.2?

Yes. We may support a version compatible with 3.2 but just for security fixes, but we don't plan to support a version that is not supported for the Rails Core for much longer.

teknull:
@rafaelfranca Thanks for the clarification.  I'm hoping to upgrade my apps to 4.2 but it's a major overhaul for much of what I've written.  I'll keep a close eye on this, so far devise is happy nestled in both of my 3.2 apps with little to no problems.  Do you know what gem version will be the last officially supported version for 3.2 branch?  Btw, you guys are the best.  I've rolled my own authentication before but it's no where near what Devise provides.  I love it.  Keep up the excellent work, the community thanks you.

WojtekKruszewski:
I'm too using Devise with Rails master too, on two new projects. I did `alias_method :prepend_before_filter, :prepend_before_action` to silence the warning.

twalpole:
An equivalent change was part of the big rails 5 compatibility PR that was recently merged into master - this PR can probably be closed

lucasmazza:
@twalpole yeah. @davetron5000 thanks for the Pull Request :)

PR #3530, opened 2015-03-23
Update validatable.rb

beviz:
Before validate password length should check whether the password is required, otherwise, if `database_authenticatable` was not set, it will throw an exception "undefined method password..".


josevalim:
Thank you but this commit breaks a test which is checking exactly the scenario you want to avoid. In your case, the best scenario is to not use `validatable`.

beviz:
How about `if: -> { respond_to?(:password) }`

josevalim:
There is no need, thanks. One shuold really not use validatable if you are not using database authentication.

PR #3531, opened 2015-03-23
Use the currently logged-in resource if available

jlecour:
When displaying the new confirmation form, if the user is already logged-in and is still in the confirmation time window, why not keeping the resource instead of resetting it to a blank one ?

I've ran the test suite before and after and I didn't break anything.
**But** I don't know devise enough to know how to add a good test for that.

Maybe we should decide if this is a good idea, then figure out how to properly test this.


josevalim:
I am closing this because I don't have any evidence of why this would be a good idea so why change? :)

PR #3547, opened 2015-04-01
Don't error if resource doesn't need confirming

andygeers:
I've overridden #confirmation_required? in my application so that only certain classes of users require it, and as a result resetting passwords has broken. This small tweak fixes it.


josevalim:
We need a test. Can you please add one?

josevalim:
Closing this. I will gladly reopen when we have tests. :)

andygeers:
Sorry, have been on holiday - I'll try to write a test later next week.

PR #3549, opened 2015-04-02
Increase default stretches to 12

jetheredge:
The current default number of stretches for devise should probably be increased from 10 to at least 12. While 10 is still the default in many bcrypt libraries, it is no longer slow enough on modern CPUs to provide good protection against brute forcing. As an example, a factor of 12 still only takes a quarter of a second on a several year old laptop:

```
Benchmark.measure do
  BCrypt::Password.create('JGQuufAzwRtxgA8Ph7eqJYgHVJGQ6s4oeLZ7txjj', cost: 12)  
end 
```

Even increasing it to 14 only takes a bit over a second on the same 2.5 year old laptop.

This has been getting some attention recently with the publishing of this post http://chargen.matasano.com/chargen/2015/3/26/enough-with-the-salts-updates-on-secure-password-schemes.html

It should also be noted that this value is stored in the bcrypt digest, and therefore can be updated without breaking any previously computed digests.

### DIFF
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -92,15 +92,15 @@ Devise.setup do |config|
   # config.clean_up_csrf_token_on_authentication = true
 
   # ==> Configuration for :database_authenticatable
-  # For bcrypt, this is the cost for hashing the password and defaults to 10. If
+  # For bcrypt, this is the cost for hashing the password and defaults to 12. If
   # using other encryptors, it sets how many times you want the password re-encrypted.
   #
   # Limiting the stretches to just one in testing will increase the performance of
   # your test suite dramatically. However, it is STRONGLY RECOMMENDED to not use
-  # a value less than 10 in other environments. Note that, for bcrypt (the default
+  # a value less than 12 in other environments. Note that, for bcrypt (the default
   # encryptor), the cost increases exponentially with the number of stretches (e.g.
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks. I would like to with 11 which is the recommendation from that blog post. Other than that, thanks for reporting @jetheredge!

georgeguimaraes:
Thank you @jetheredge! I've merged your PR with stretches being 11.

georgeguimaraes:
https://github.com/plataformatec/devise/commit/dc7e7d66c99c77621ac6073259db4666251e6475

aried3r:
On that note, is there a way devise can migrate old bcrypt digests or do I have to manually migrate (as users login again)?

I read

> It should also be noted that this value is stored in the bcrypt digest, and therefore can be updated without breaking any previously computed digests.

but from what I've gathered, devise, more precisely bcrypt, will handle both, old and new, so I can update the factor without worrying, but won't migrate old ones to new ones, correct?

sergey-alekseev:
Regarding @aried3r's question:
>devise, more precisely bcrypt, will handle both, old and new, so I can update the factor without worrying, but won't migrate old ones to new ones, correct?

Correct.  
Re "I can update the factor without worrying" – https://github.com/plataformatec/devise/pull/5074#issuecomment-492687076.
Re "won't migrate old ones to new ones" – https://github.com/plataformatec/devise/pull/5074#issuecomment-492696360.

PR #3552, opened 2015-04-07
Added missing require_no_authentication filter in ConfirmationsController

msx2:
Hi guys!

I noticed that authenticated user still has a possibility to access confirmations module. :worried:
Should it really be the case? I mean, authenticated user cannot access any other module, e.g. unlocks, passwords.

Here's a missing filter applied


josevalim:
It is fine for the user to do that. Many sites allow you to be signed in without confirming the account. We don't really want users to sign out just so they confirm it.

PR #3563, opened 2015-04-17
Add a replacement for the last character of a confirmation token.

naturtrunken:
A user could not activate his account through confirmable in Devise 3.4.1. He sent me the email and I saw that the URL was probably not recognized by his email client. The hyphen as a last character is not interpreted as belonging to the URL (see image). Therefore, the code was incompletely transmitted to the browser.

I know that this is very, very, very, unlikely to be happen, but I don't want that users resign only because they are not able to confirm their account. (Of course, they could demand a new token, but who will do that..?) So, I simply made a patch which makes sure that the last character of the token is definitely not a hyphen.

This should be sufficient, because the other characters which urlsafe_base64 can generate are correctly detected (by Thunderbird).

bildschirmfoto 2015-04-17 um 18 10 42

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -466,7 +466,7 @@ module Devise
 
   # Generate a friendly string randomly to be used as token.
   def self.friendly_token
-    SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz')
+    SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz').last.tr('-', '4')
   end
 
   # constant-time comparison algorithm to prevent timing attacks
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -466,7 +466,8 @@ module Devise
 
   # Generate a friendly string randomly to be used as token.
   def self.friendly_token
-    SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz').last.tr('-', '4')
+    token = SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz')
+    (token.last == '-') ? token.chop + '4' : token
   end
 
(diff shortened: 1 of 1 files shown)


josevalim:
Isn't this going to return a single character string?

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Lead Developer

naturtrunken:
D'oh, of course, thanks! I updated it with chop.

josevalim:
Chop changes the size of the token. Maybe it is easier to just add - to the
list of tr values. :)

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Lead Developer*

lucasmazza:
:+1: for moving it to the `tr` call that we already have.

naturtrunken:
That's right, but only by one character. (Because urlsafe_base64 cannot create non-printable characters.) So I delete one and add one.

But your idea is better indeed. I tried to change the format of the string so little as possible, but when I think about it, it doesn't matter (at least for my applications) if there are hyphens in the string or not.

Thanks for your help!

josevalim:
I am having mixed feelings about this. Why not send an HTML e-mail where everything is properly marked as a link? Why are you relying on the email client ability to autolink? :)

naturtrunken:
Because many people are still using text emails. And I don't like the idea that Devise would only work with HTML emails correctly. The fix doesn't effect anything in a negative way and is compatible with autolink and HTML links.

josevalim:
You can send multipart e-mails where the text part should not be autolinked with a proper html part too. Projects like markerb should make it easy to send both: https://github.com/plataformatec/markerb

Thanks!

PR #3570, opened 2015-04-19
Removes the bang in confirm! and reset_password!

### LINKED ISSUE #3412: Confirmable API
Hi @josevalim,

As I was looking at the `confirm!` method in confirmable: https://github.com/plataformatec/devise/blob/85238014593443fb66cd0b1387f7ec8103f0fb91/lib/devise/models/confirmable.rb#L59-L84

And find the API a bit odd.

`confirm!` expects users to check for its return value, true or false, to detect possible errors.
This might not be a reasonable assumption. The name of the method implies that it would raise, similarly to the `save` vs `save!` behavior.

Here's some data to backup my claims:
- GitLab, a rails app made by fairly advanced coders exhibits the mistake: https://github.com/gitlabhq/gitlabhq/blob/0991d19c0d157025db8e9bf01d216771a4dbfe31/app/controllers/admin/users_controller.rb#L76
- Lots of small applications contain this mistake: https://github.com/search?l=ruby&o=desc&p=1&q=%22user.confirm%21%22+filename%3Acontroller+NOT+test+NOT+Factory+NOT+colchonet&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93

So I'd say that `confirm!` should use `save!` instead of returning true/false. Or be renamed to just `confirm?`, or `try_confirm`, to make it explicit to the user should check for the return value.

What do you think?

Side note: I have personal interest to go with the `save!` route, because NoBrainer's `save!` method has the same semantics as other ORMs, unlike `save` which behaves like `save!`.

nviennot:
Fixes #3412

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
   * Controllers inheriting from any Devise core controller will now use appropriate translations.
     The i18n scope can be overridden in `translation_scope`.
 
+* deprecations
+  * `confirm!` has been deprecated in favor of `confirm`.
+  * `reset_password!` has been deprecated in favor of `reset_password`.
+
 ### 3.4.1 - 2014-10-29
 
 * enhancements
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -24,7 +24,7 @@ module Devise
     #
     # == Examples
     #
-    #   User.find(1).confirm!      # returns true unless it's already confirmed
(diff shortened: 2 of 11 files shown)


nviennot:
edit: added the deprecation methods

josevalim:

nviennot:
sorry, just pushed a new version (the reset_password!() arguments wasn't matching the one of reset_password()).

josevalim:

PR #3575, opened 2015-04-21
Conditionally use flashes in failure_app

### LINKED ISSUE #3574: Flash support in warden failures
I'm using json as the only endpoint and rails-api gem.
I'm getting exception `undefined method 'flash'` in login error responses, looking at the code like https://github.com/plataformatec/devise/blob/master/lib/devise/failure_app.rb#L52 it seems it's not conditionally use flash as in controllers https://github.com/plataformatec/devise/blob/master/app/controllers/devise/sessions_controller.rb#L18 only when the format is navigational, maybe it should be checked too in the failure_app?

alex88:
Fixes #3574

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -49,17 +49,19 @@ module Devise
 
     def recall
       env["PATH_INFO"]  = attempted_path
-      flash.now[:alert] = i18n_message(:invalid)
+      flash.now[:alert] = i18n_message(:invalid) if is_flashing_format?
       self.response = recall_app(warden_options[:recall]).call(env)
     end
 
     def redirect
       store_location!
-      if flash[:timedout] && flash[:alert]
-        flash.keep(:timedout)
-        flash.keep(:alert)
-      else
-        flash[:alert] = i18n_message
+      if is_flashing_format?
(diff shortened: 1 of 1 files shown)


josevalim:

serene:
@alex88 Would we then be able to get error messages from somewhere else, if flash messages are no longer being populated in flash?

alex88:
@serene you already get the error In the response afaik

serene:
@alex88 Would this be the case even in the event of a remote/JS sign in call?

alex88:
@serene I think so, I don't remember the case, but I think I was exactly calling the API to sign in via ajax

lib/devise/failure_app.rb
```diff
@@ -49,17 +49,19 @@ def http_auth
 
     def recall
       env["PATH_INFO"]  = attempted_path
-      flash.now[:alert] = i18n_message(:invalid)
+      flash.now[:alert] = i18n_message(:invalid) if is_flashing_format?
```
serene: I think the errors are getting lost for ajax calls. They never get set anywhere else for a non-navigational format.

alex88: Have you tried? What's the response of the ajax call?

PR #3579, opened 2015-04-23
refactor Devise.secure_compare

bolshakov:
method becomes a bit clearer


josevalim:
I do not necessarily agree it is clearer and I would prefer to avoid "style" changes on code that can be sensitive in terms of security. Thank you though!

PR #3587, opened 2015-05-02
Allow save() to be overridden by ORM adapters

nviennot:
Ref. #3412 

Needs https://github.com/ianwhite/orm_adapter/pull/45


nviennot:
Nevermind, I no longer need this.

PR #3594, opened 2015-05-06
Split out BCrypt hashing to make it reusable

magnusvk:
This logic is generic and reusable -- hash a secret; and take an unhashed secret and compare it to a hashed secret. This breaks this out to make it reusable in other places. Specifically, we use this in our own token auth at Bonobos that we plan to split out as a Devise extension. This will make that possible without copy & pasting this code.

### DIFF
--- a/lib/devise/encryptor.rb
+++ b/lib/devise/encryptor.rb
@@ -1,0 +1,22 @@
+require 'bcrypt'
+
+module Devise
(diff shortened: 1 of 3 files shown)


josevalim:

magnusvk:
Thanks for the quick merge.

Issue #3505, opened 2015-03-04
Possible bug in threaded apps: ActionView::Template::Error: undefined method `url_options' for #<Module:xxx>


IAMRYO:
There is an issue in puma/puma that talks about this error, which we just encountered.  Here's our stack trace: https://gist.github.com/IAMRYO/e8bee5a8e6710ad4b970

Here's the issue on puma/puma: https://github.com/puma/puma/issues/647

Here's the bug that was reported to MRI: https://bugs.ruby-lang.org/issues/10871

and here's a way to reproduce the issue: https://gist.github.com/evanphx/6eef92f2c40662a4171b

josevalim:
Thanks for letting us know!

glittershark:
So is this fixed?

glittershark:
Requesting a reopen of this, since it essentially invalidates the "Devise is guaranteed to be thread-safe" claim

josevalim:
@glittershark Is this a Devise issue though? Can you reproduce it or not in a regular Rails app? Maybe we should claim to be as thread-safe as Rails? :P

Afaik, Devise is not doing anything in particular here. The url helpers code that you showed is run on initialization, before we serve requests, and it should not be the source of the concurrency issues:

https://github.com/plataformatec/devise/blob/master/lib/devise/rails.rb#L16-L18

I merged your PR because it leads to cleaner code, not because it would directly solve this issue. :)

Thoughts?

glittershark:
:+1: for cleaner code :smile:

As far as the issue goes - the problem arises not on the actual running of
the initialization code, but when the methods defined there are called -
specifically, due to [this MRI issue][1] (which afaik is fixed in ruby
unstable, ie what will eventually be 2.2.3) the `class_eval`, or now
`define_method`, is only successfully run on some threads. It's possible
that this might be able to be solved by fiddling with Puma's app preloading
- I'll do some more testing when I'm not afk.

josevalim:
I would expect Puma to start threads only after the app is initialized. :( Is that not the case with Puma & Rails?

PR #3597, opened 2015-05-12
Use `define_method`, not `class_eval`, for url_helpers

glittershark:
I'm not hugely familiar with ruby internals vis-a-vis sclass semantics,
but this *probably* serves as a workaround for the MRI thread-safety
bug mentioned in #3505.
Beyond that, and even if this doesn't fix the thread-safety issue, per
[this blog post][1], `define_method` is recommended over `class_eval`
for performance (and, fwiw, readability) reasons anyway.

[1]: http://tenderlovemaking.com/2013/03/03/dynamic_method_definitions.html


glittershark:
:sob: I was able to pound away at my Puma server enough to confirm that this *doesn't* fix the thread-safety issue in #3505.

PR #3603, opened 2015-05-21
add config for setting layout in devise

BenMorganIO:
While working on a project, I needed the layout of Devise to have a different layout than what was being inherited.


josevalim:
You can customize this in other ways, using conditional layouts in your application controller, having a custom parent controller for Devise, and so on. It won't be as concise as this patch but it isn't such a common use case and we already have waaaaay to many configuration options. Thanks for the patch anyway! :)

PR #3615, opened 2015-06-03
fixed typo referring to routes.rb file

Dbz:
When creating devise controllers there is a message presented:

     Some setup you must do manually if you haven't yet:

      Ensure you have overridden routes for generated controllers in your route.rb.
      For example:

      devise_for :users, controllers: {
        sessions: 'users/sessions'
      }
    end

However, the file "Route.rb" should be "Routes.rb"


carlosantoniodasilva:
Thanks!

PR #3617, opened 2015-06-03
Refactoring `build_resource`

yui-knk:
To set default args (empty hash)

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -96,8 +96,8 @@ class Devise::RegistrationsController < DeviseController
 
   # Build a devise resource passing in the session. Useful to move
   # temporary session data to the newly created user.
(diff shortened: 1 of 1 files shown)


app/controllers/devise/registrations_controller.rb
```diff
@@ -96,8 +96,8 @@ def update_resource(resource, params)
 
   # Build a devise resource passing in the session. Useful to move
   # temporary session data to the newly created user.
-  def build_resource(hash=nil)
-    self.resource = resource_class.new_with_session(hash || {}, session)
+  def build_resource(hash = {})
```
josevalim: This changes the behaviour. Now, if someone passes `build_resource(nil)`, we will send nil to the model, which is not what we want.

yui-knk:
OK, I see. Thanks for your review!

PR #3618, opened 2015-06-03
Doesn't send reset password instructions if access_locked? && !unlockable?(resource) 

AlexKlim:
I think, devise should not send reset password instructions if 'unlock strategy' set as :none and 'access locked' is true for some user.  
The first moment: 
If devise sends the instructions, user goes to the link from email, he fills in a field with a new password, but, after it, he will see "Your account is locked." error message. It is confuses the user. 
The second moment: 
We should not load any email service, if it is not necessary. It is small performance.

What do you think about it ?

### DIFF
--- a/app/controllers/devise/passwords_controller.rb
+++ b/app/controllers/devise/passwords_controller.rb
@@ -10,8 +10,16 @@ class Devise::PasswordsController < DeviseController
 
   # POST /resource/password
   def create
-    self.resource = resource_class.send_reset_password_instructions(resource_params)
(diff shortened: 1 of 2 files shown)


josevalim:
I don't consider this is an improvement. The user can still be able to reset their password even if their account is locked or perform other actions that do not require signing in. Of course, you can always customize it in your applications! Thank you!

AlexKlim:
Ok, thanks.. But, I can't understand why I should be able to reset password, If my account can't be unlocked anyway. And I can't to login. It is a little bit a confuse me.

PR #3619, opened 2015-06-04
Validating the presence of password and confirmation in recoverable model

ojash:
In the process of recovering, though the password_confirmation field was mandatory, it was not properly validating so password was being reset when the password_confirmation was set nil.
It fixes that issue and returns proper validation message when password or confirmation_password are nil or blank.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,14 +1,3 @@
-PATH
-  remote: .
-  specs:
-    devise (3.5.1)
-      bcrypt (~> 3.0)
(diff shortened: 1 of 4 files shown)


josevalim:
That's how Rails validations works. We are not changing it. :) If you want to be sure there is a validation, just include it in your forms and it should be validated.

josevalim:
In any case, thank you for the PR! :heart:

ojash:
:v:

PR #3630, opened 2015-06-12
Autoload Encryptor

NAndreasson:
Spring started complaining about Devise::Encryptor being uninitialized so I figured that it should be autoloaded.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -8,6 +8,7 @@ require 'responders'
 
 module Devise
   autoload :Delegator,          'devise/delegator'
+  autoload :Encryptor,          'devise/encryptor'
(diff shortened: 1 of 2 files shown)


carlosantoniodasilva:
It should not be a problem in a normal workflow because we require it in database authenticatable, but I also don't see any harm in doing the autoload. This probably means we can [remove the require](https://github.com/plataformatec/devise/blob/aa49dd53b5578c7f2d840c9e171998f7b99fd398/lib/devise/models/database_authenticatable.rb#L2), then, would you like to do that as well? Thanks.

NAndreasson:
@carlosantoniodasilva Ok, perfect. Just removed the require statement.

carlosantoniodasilva:
Thanks.

PR #3641, opened 2015-06-17
Update rails gem for security

camelmasa:
See. http://weblog.rubyonrails.org/2015/6/16/Rails-3-2-22-4-1-11-and-4-2-2-have-been-released-and-more/


lucasmazza:
Thanks!

PR #3645, opened 2015-06-21
Maybe it's time to create rails5-support branch?

hedgesky:
Rails 5.0 will be released in autumn, and it has several changes impacting devise, including:
  - Requiring Ruby 2.2.2 or newer;
  - Deprecating `alias_method_chain` and `before_filter` methods.

So I know it's impossible to merge these commits into `master` but maybe it's time to create v4 branch with Rails 5 support?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,16 @@
 language: ruby
 
 rvm:
-  - 1.9.3
-  - 2.0.0
-  - 2.1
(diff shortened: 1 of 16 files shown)


hedgesky:
After some thinking I've decided that removing one or two deprecation warnings doesn't worth dropping support for several Ruby versions. Sorry for the noise.

PR #3662, opened 2015-06-30
Just use regular 4.x mongoid gem

vincentwoo:
Hopefully this PR makes tests green again.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,5 +25,5 @@ platforms :ruby do
 end
 
 group :mongoid do
-  gem "mongoid", github: "mongoid/mongoid", branch: "master"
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
I suppose we could just add the mongo repo as well, but that works too, at least for now. Thanks!

yogeshjain999:
@carlosantoniodasilva tests are failing for Mongoid 5.0 because, `test/support/mongoid.yml` is expecting `clients` instead of `sessions`. Why can't we replace it ?

josevalim:
We can likely define both eys?

yogeshjain999:
Yes, we can do that too, by checking Mongoid version.

PR #3672, opened 2015-07-11
Remove obsolete unused `expire_auth_token_on_timeout` config

deivid-rodriguez:
Hi!

I was going to add `timeoutable` to my app so I had a look at the config and saw `expire_auth_token_on_timeout`. So asked myself "what is this?" and "do I need it?". I had a look at the code and saw it calls the `reset_authentication_token!` method that it's not defined anywhere. Then I did a bit of git digging and I think all of this was probably meant to be removed together with dff7891.

Regards and thanks for this gem!


josevalim:
:heart:

PR #3673, opened 2015-07-13
allow setting of reset_password_token on an unpersisted record, in a way that matches the functionality of setting the confirmation_token

betesh:
It's possible to set the confirmation_token before saving the record and then wait until after the record is saved to email the confirmation link.  This PR adds the same functionality for the reset_password_token.

The motivation is a use case where the model is not `devise :registerable`.  Instead, an Administrator creates the user with just the authentication key (email) and the new user immediately receives an email with a link to set his password.

Devise::Models::Validatable adds this validation:

```
validates_presence_of     :password, if: :password_required?
validates_confirmation_of :password, if: :password_required?
```

So there's a built-in way to ignore a missing password when desired, and it would be a shame to have to re-implement all validation when instead I should just be able to override `password_required?`

So I tried overriding it as follows:

```
def password_required?
  reset_password_token.blank? && super
end
```

However, there was no way to set the reset_password_token until the record is saved.  This PR adds the code to make that possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * enhancements
   * Upon setting `Devise.send_password_change_notification = true` a user will receive notification when their password has been changed.
+  * reset_password_token can now be set before saving a new record. (@betesh, #3673)
 
 ### 3.5.2 - 2015-08-10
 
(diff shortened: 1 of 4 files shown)


betesh:
I discovered a bug in this PR this morning and fixed it

tegon:

Thank you!

PR #3675, opened 2015-07-14
Replace all instances of Log in to Sign In

foxlance:
This makes sense as there are no instances of log out on the site. Methods and URL also uses sign_in


josevalim:
We have renamed those on purpose because "sign in", "sign out" and "sign up" is confusing. The idea is to use three different words. However, we can't rename Devise internals, as it is backwards inompatible, so only the text mentions "log in" and "register" and similar.

PR #3661, opened 2015-06-30
Do not use digests for confirmation tokens

vincentwoo:
Addresses https://github.com/plataformatec/devise/issues/3640.

I did this by changing as little code as possible. It's definitely possible to strip out `raw_confirmation_token` entirely, but you'd have to change a LOT of specs. Please advise if this is up to par for your repository.

PR #3676, opened 2015-07-16
Token overwrite

urbas:
This PR changes the behaviour:

- The method `Confirmable#confirm_by_token(non_digested_token)` sets the `confirmation_token` to the non-digested one

to this behaviour:

- The `confirmation_token` is not modified.

Also, an unrelated test was failing on my Windows machine. It tests that the user is not active for authentication when the resource is unconfirmed for exactly the timeout period. I fixed it by pushing the `confirmation_sent_at` date one second further into the past.


carlosantoniodasilva:
Considering closed by #3661, let us know if there's anything missing here. Thanks.

PR #3679, opened 2015-07-17
User unconfirmed access should not be extended by re-sending confirmation email

dimonzozo:
I think there is an issue with how `confirmable#confirmation_period_valid?` method logic was implemented.

User is able to access website during period specified in `allow_unconfirmed_access_fo` setting, but she can extend this period by just requesting re-sending confirmation email again and again.

I found [discussion](https://groups.google.com/forum/#!topic/plataformatec-devise/4ab91Vr1b9k) in Google Groups, but it seems no one give attention to this problem.

What do you think about?

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -198,7 +198,7 @@ module Devise
         #   confirmation_period_valid?   # will always return true
         #
         def confirmation_period_valid?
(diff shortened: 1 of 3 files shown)


dimonzozo:
If #3661 will be merged we can stop worrying about this issue, because token will not be regenerated on re-sending confirmation instructions.

carlosantoniodasilva:
Considering closed by #3661, please let us know if you still find any issue. Thanks!

Issue #2058, opened 2012-09-12
I18n Views


emaiax:
Hello guys, I'm sorry if it's a duplicated issue about this subject, but I looked everywhere and didn't found anything.

Almost every project, I need to re-write Devise views to use Rails' I18n feature at the same time I put a translation into config/locale folder.

Preventing that, I already started a patch to send you a pull request changing all those hard-coded strings to I18n format using locales, but I prefer to know what you think about it first.

Issue #1642, opened 2012-02-12
Why no i18n in views?


tmaier:
Is there any reason why the views at 
- https://github.com/plataformatec/devise/tree/master/app/views/devise
- https://github.com/plataformatec/devise/tree/master/lib/generators/templates/simple_form_for

do not use any calls to i18n?

Else I would sit down and add all of them to `devise.en.yml`.

PR #3684, opened 2015-07-24
Add translations to view

Kendos-Kenlen:
Hi,

I added translations to view to avoid having to edit them.

I replaced all double quoted strings when unnecessary by simple quoted strings to follow Ruby Style Guide.


rafaelfranca:
Thank you so much for the work on this pull request but we'll not accept this change.

See #2058 #1642 #1983 #1989 and many others similar issues/pull requests for the reasons.

Issue #3685, opened 2015-07-26
Return of error "undefined method `mimes_for_respond_to'" on devise 3.5.1 rails 4.2.3


cefigueiredo:
Hi guys,

I noticed an error on Devise that was supposed to be solved by #3153 and I want to report it.

As mentioned on the title, the problem is that an error `undefined method 'mimes_for_respond_to'` is being raised when I try access any route.

I setup Devise properly, following the Getting Starter, setting on my ApplicationController a `before_filter :authenticate_user!` properly (Off course User is my model for authorization).

When I try to access any action I get this result:

```
Started GET "/pages" for ::1 at 2015-07-26 19:47:53 -0300
Processing by PagesController#index as HTML
Completed 500 Internal Server Error in 9ms (ActiveRecord: 0.0ms)

NameError (undefined local variable or method `mimes_for_respond_to' for DeviseController:Class):
  devise (3.5.1) app/controllers/devise_controller.rb:12:in `<class:DeviseController>'
  devise (3.5.1) app/controllers/devise_controller.rb:2:in `<top (required)>'
  activesupport (4.2.3) lib/active_support/dependencies.rb:457:in `load'
  activesupport (4.2.3) lib/active_support/dependencies.rb:457:in `block in load_file'
  activesupport (4.2.3) lib/active_support/dependencies.rb:647:in `new_constants_in'
  activesupport (4.2.3) lib/active_support/dependencies.rb:456:in `load_file'
  activesupport (4.2.3) lib/active_support/dependencies.rb:354:in `require_or_load'
  activesupport (4.2.3) lib/active_support/dependencies.rb:494:in `load_missing_constant'
  activesupport (4.2.3) lib/active_support/dependencies.rb:184:in `const_missing'
  devise (3.5.1) lib/devise/controllers/helpers.rb:144:in `devise_controller?'
  devise (3.5.1) lib/devise/controllers/helpers.rb:112:in `authenticate_user!'
  activesupport (4.2.3) lib/active_support/callbacks.rb:430:in `block in make_lambda'
  activesupport (4.2.3) lib/active_support/callbacks.rb:162:in `call'
  activesupport (4.2.3) lib/active_support/callbacks.rb:162:in `block in halting'
  activesupport (4.2.3) lib/active_support/callbacks.rb:502:in `call'
  activesupport (4.2.3) lib/active_support/callbacks.rb:502:in `block in call'
  activesupport (4.2.3) lib/active_support/callbacks.rb:502:in `each'
[...]

PR #3686, opened 2015-07-27
Set minimum version 2.0 to responders

cefigueiredo:
On environment that uses just bundler to manage gems and has installed an old
version of responders (prior 2.0.0) that old version will be installed ,
but Devise will raise errors trying to call methods there exists only on
responders 2.0+

It solves #3685

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,7 +5,7 @@ PATH
       bcrypt (~> 3.0)
       orm_adapter (~> 0.1)
       railties (>= 3.2.6, < 5)
-      responders
+      responders (~> 2.0)
       thread_safe (~> 0.1)
       warden (~> 1.2.3)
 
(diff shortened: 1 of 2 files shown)


josevalim:
Unfortunately we cannot do this because Responders 2.0.0 is Rails 4.2. Maybe we could release a new Devise version with Rails 4.2 only support, in preparation to Rails 5, but we will need to decide that before moving on.

cefigueiredo:
On Responders Readme for versions 2.0.0 .. 2.0.x says it is Rails 3.2+ compatible.
Is it wrong? If don't, I think it would be ok fix responders version to '~> 2.0.2'

cefigueiredo:
Nevermind, I read the Changelog, and there really says that 2.0.x is rails 4.2 compatible only :(

PR #3687, opened 2015-07-27
Error on submit

pauloschiavon:
With the url confirmation_path(resource_name) generates a post on "/confirmation.user" generating the error "No route matches"

### DIFF
--- a/lib/generators/templates/simple_form_for/confirmations/new.html.erb
+++ b/lib/generators/templates/simple_form_for/confirmations/new.html.erb
@@ -1,6 +1,6 @@
 <h2>Resend confirmation instructions</h2>
 
-<%= simple_form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: { method: :post }) do |f| %>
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
@pauloschiavon thanks, but we can't just rely on `user_confirmation_path` since your scope might not be `:user`, it could be anything. That's why the generic helpers that receive the `resource_name` argument exists, so they are "polymorphic" based on the scope they receive.

Anyway, for the `/confirmation.user` route to be generated, it seems it's interpreting the `:user` resource name argument as the format instead, so I'm thinking something might be off with your routes. Please try sending a question to stack overflow or the mailing list, posting your current devise routes definition and the related output of `rake routes`.

I'm closing since it doesn't seem to be an actual devise issue. Thanks!

pauloschiavon:
@carlosantoniodasilva Really you are correct! After opening the pull request i thought the polymorphism.
I will reassess my code and if the problem still, I'll make an issue.

Sorry for the inconvenience.

carlosantoniodasilva:
No problem! :) Please feel free to use the mailing list or ping us in case you find anything. Thanks.

Issue #3639, opened 2015-06-16
Email regexp failing on invalid address


solanoluz:
!!('teste@gmail..com' =~ Devise.email_regexp)
> true

carlosantoniodasilva:
The email regexp is not supposed to cover everyone's cases / validations, it's a simple/basic filter. Please check #3537 and #3376 for more information (there might be other related issues as well). Thanks.

solanoluz:
Thanks!


The alternative regexp posted at issue #3537 seems to cover this case.



2015-06-16 13:39 GMT-03:00 Carlos Antonio da Silva <notifications@github.com

Issue #3537, opened 2015-03-25
Should not the default email regex be the same as Rails 


ybart:
I have one user which was able to register with an email that looks like the following (we don't use email confirmation on our application) :

```
pxxx@oxxx.xxx,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,à
```

(I've substituted some characters with x for obvious confidentiality reasons)

I assume that Rails EmailValidator is not used for compatibility reasons.

In EmailValidator the regexp used by Rails looks like this :

```
/\A[^@\s]+@([^@\s]+\.)+[^@\s]+\z/
```

The one from Rails is the following :

```
/\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i
```

After looking at #3376, I understand that the user part might need to be loose, but I don't see any problem having a more strict validation for the domain part.

If nobody disagrees with this, I would be happy to publish a pull request for this.

carlosantoniodasilva:
As far as I remember, Rails does not provide an email validator, it contains one used on Active Model tests and some examples on how to do that with format validation on the docs if I'm not wrong, but that's about it.

In any case, email validation is somewhat controversial, it may vary from app to app, so if Devise does not seem to provide enough confidence for you in this validation, you can just [change the default regexp in your devise config](https://github.com/plataformatec/devise/blob/v3.4.1/lib/devise.rb#L120-L124).

Unfortunately I don't think Devise is changing it's default email regexp, also because it may also break other applications when they upgrade. Thanks!

ybart:
I've done that. Thanks you !

Issue #3376, opened 2014-12-17
"bad_email$ %@^&" passes email validation


mhuggins:
While writing an rspec test, we were ensuring the failure path of a controller worked as expected.  Unfortunately, validation did _not_ fail when using the string "bad_email$ %@^&".  If the "@" is removed from the string, validation fails as expected.  It looks like the validation rule is not strict enough on ensuring an email address is formatted properly.

PR #3697, opened 2015-08-03
Change email_regexp to reject e.g. foo@bar..com

mdemanett:


georgeguimaraes:
Hi. Thanks for your contribution but Devise's regexp is not supposed to cover all possible email cases. You can always change that in your app.

See #3537, #3639, #3376 and others for more info.

PR #3703, opened 2015-08-11
Delete of current password in update_without_password

diogorosa:
Deletion of current password key when updating without password.
This enables a hybrid model on the form that needs current password
for update specific fields, some with pass and others without.

Signed-off-by: Diogo Rosa <diogo.m.c.rosa@gmail.com>

### DIFF
--- a/lib/devise/models/database_authenticatable.rb
+++ b/lib/devise/models/database_authenticatable.rb
@@ -92,6 +92,7 @@ module Devise
       #   end
       #
       def update_without_password(params, *options)
+        params.delete(:current_password)
         params.delete(:password)
         params.delete(:password_confirmation)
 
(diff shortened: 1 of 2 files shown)


josevalim:
I don't understand the need. Don't you need to choose between calling update_with_password and update_without_password anyway? Can't you just delete the parameter then? Also, even if you pass it, nothing should happen, should it?

diogorosa:
Currently I'm deleting it, because if you pass it to the update_without_password it'll raise an error. Yes I need to choose, but it would be simpler not to be worried when you have an hybrid model to delete the current_password before calling the update_without_password.

josevalim:
We raise to avoid calling the wrong method accidentally so what you have today is the ideal. Thanks for you contribution though!

PR #3712, opened 2015-08-18
confirmation_sent_at updated when confirmation token is re-sent

randoum:
Since PR #3661 `confirmation_sent_at` is not updated when a non-expired confirmation token is re-sent.


josevalim:
I am not sure we should. Maybe confirmation_sent_at is a bad name now. I would like it to be kept though as when the token as generated. Otherwise the token can last forever by simply renewing it over and over again and it does not sound safe.

randoum:
It does not sound as a security problem either. 
Anyhow I can do differently.

Issue #3701, opened 2015-08-09
block in with_devise_exclusive_scope': undefined method `[]=' for #<ActionDispatch::Routing::Mapper::Scope:0x007f8b29ea4938> (NoMethodError)


aantix:
master broken with commit https://github.com/rails/rails/commit/5e81e6ca31aa0d2a38f8db5c8fe9bcbce98d9689 from @tenderlove.

Devise relies upon that setter for setting scope route attributes
https://github.com/plataformatec/devise/blob/master/lib/devise/rails/routes.rb#L446

```
/Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/devise-48d82853449b/lib/devise/rails/routes.rb:446:in `block in with_devise_exclusive_scope': undefined method `[]=' for #<ActionDispatch::Routing::Mapper::Scope:0x007f8b29ea4938> (NoMethodError)
	from /Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/devise-48d82853449b/lib/devise/rails/routes.rb:446:in `each_pair'
	from /Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/devise-48d82853449b/lib/devise/rails/routes.rb:446:in `with_devise_exclusive_scope'
	from /Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/devise-48d82853449b/lib/devise/rails/routes.rb:248:in `block (2 levels) in devise_for'
	from /Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/devise-48d82853449b/lib/devise/rails/routes.rb:351:in `block in devise_scope'
	from /Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/rails-5e81e6ca31aa/actionpack/lib/action_dispatch/routing/mapper.rb:939:in `block in constraints'
	from /Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/rails-5e81e6ca31aa/actionpack/lib/action_dispatch/routing/mapper.rb:815:in `scope'
	from /Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/rails-5e81e6ca31aa/actionpack/lib/action_dispatch/routing/mapper.rb:939:in `constraints'
	from /Users/aantix/.rvm/gems/ruby-2.2.2/bundler/gems/devise-48d82853449b/lib/devise/rails/routes.rb:350:in `devise_scope'
[...]

PR #3714, opened 2015-08-19
Rails 5 Support

twalpole:
This PR adds updates/changes to pass all tests with the current Rails 5 master branch.  ~~Note: the tests are pretty noisy with a lot of deprecation warnings.~~


lucasmazza:
@twalpole woah, thanks! This is a great start for Devise to support Rails 5 :heart:

lib/devise/models/authenticatable.rb
```diff
@@ -253,6 +253,7 @@ def find_or_initialize_with_error_by(attribute, value, error=:invalid) #:nodoc:
 
         # Find or initialize a record with group of attributes based on a list of required attributes.
         def find_or_initialize_with_errors(required_attributes, attributes, error=:invalid) #:nodoc:
+          attributes = attributes.to_unsafe_h.with_indifferent_access if attributes.respond_to? :to_unsafe_h
```
lucasmazza: This is due some change on `ActionController::Parameters`? I wonder if would be better to change this on the controller/parameter sanitizer level to always return a Hash from up there.

lib/devise/models/database_authenticatable.rb
```diff
@@ -37,6 +37,7 @@ def self.required_fields(klass)
 
       # Generates password encryption based on the given value.
       def password=(new_password)
+        self.encrypted_password_will_change! #trigger validations of the newly set password
```
lucasmazza: should we push this inside the `if @password.present?` from the `encrypted_password` assignment?

lib/devise/rails/routes.rb
```diff
@@ -415,7 +415,12 @@ def devise_omniauth_callback(mapping, controllers) #:nodoc:
 ERROR
         end
 
-        path, @scope[:path] = @scope[:path], nil
+        path = @scope[:path]
```
lucasmazza: Using `@scope.frame` seems to fix #3701, but I wonder if we should try something with `@scope.new` and `@scope.parent` from @tenderlove patch from that thread.

lib/devise/models/authenticatable.rb
twalpole: ActionController::Parameters is no longer derived from Hash and tests pass ActionController::Parameters directly to this method so I assumed it was required to accept them here

lib/devise/models/database_authenticatable.rb
twalpole: This is outside the @password.present? so that if a blank password is set it still triggers validation, without it here tests fail.

lib/devise/rails/routes.rb
twalpole: I missed #3701 - will take a look, although using frame may be cleaner code for backwards compatability

twalpole:
Ok - taking hints from @tenderlove posting in issue #3701 I moved away from using Scope#frame.  Passing path: nil to match as suggested in #3701 doesn't seem to make any difference so its not exactly what was suggested there - all tests passing for all tested versions of rails

FanaHOVA:
@twalpole I just tried your fork on a clean install of Rails 5 and it doesn't work, am I missing something? I have

    gem 'devise', github: 'twalpole/devise'

In my gemfile, I've ran bundle update and then install, ran the generators but I still get the same

    /Users/fana/.rvm/gems/ruby-2.2.2/bundler/gems/devise-213aa51126d5/lib/devise/rails/routes.rb:446:in `block in with_devise_exclusive_scope': undefined method `[]='

blabla.

twalpole:
@FanaHOVA you need to use the rails5 branch

    gem 'devise', github: 'twalpole/devise', branch: 'rails5'

FanaHOVA:
@twalpole Oh right, everything works fine for now, thank you!

jasnow:
Should we fix this too?

DEPRECATION WARNING: alias_method_chain is deprecated. Please,
use Module#prepend instead. From module, you can access the
original method using super. (called from <class:RouteSet> at
$HOMEx/.rvm/gems/ruby-2.2.3@2.2.3/bundler/gems/devise-af405f0c1544/
lib/devise/rails/routes.rb:27)

REF: https://github.com/rails/rails/pull/19434

jasnow:
Is this include too?

DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from <class:RouteSet> at
$HOMEx/.rvm/gems/ruby-2.2.3@2.2.3/bundler/gems/devise-af405f0c1544/lib/devise/rails/routes.rb:27)

REF: https://github.com/rails/rails/pull/19434

twalpole:
@jasnow  I've only done the work to make devise work with Rails 5 while maintaining backwards compatibility with all other currently supported rails versions and ruby versions.  Fixing that deprecation warning cleanly means moving to ruby 2.0 + only so nothing has been done about it yet.

audionerd:
Thanks for all the work on this branch to get Devise on Rails 5 up and running!

I'm using this branch, but I ran into an issue where, when attempting to sign in, there was no redirect, and instead the error message: `No template found for Users::SessionsController#create, rendering head :no_content`.

The solution is to use the latest master of `responders`, which has Rails 5 Support:

```ruby
gem 'responders', github: 'plataformatec/responders'
```

Reference:
- plataformatec/responders#131
- plataformatec/responders#126
- rails/rails#20276

twalpole:
@audionerd yep -- please look at https://github.com/twalpole/devise/blob/rails5/gemfiles/Gemfile.rails-5.0-alpha to see the gems that have to be used for the tests to pass with rails5

audionerd:
@twalpole Nice! I didn't think to reference that that. Thanks!

lucasmazza:
Just a head's up to anyone who might drop by here and believe this thread to be abandoned, we intent to
revisit this PR back in november (once we get some spare time to focus on Devise for a couple of weeks) and try to get a deprecation free codebase with support for Rails 5, probably dropping support for older Rails versions. There is a lot of work to do and I don't think we might be able to do on small sprints of different PRs, as a lot of the changes are related to one another :see_no_evil:. 

I would like to avoid any new Pull Requests on small changes and fixes related to Rails 5 without the Travis matrix update that we have here or more version checks on the attempt to support both Rails 4 and 5. We can't promise constant compatibility with rails HEAD, but hopefully we might be able to have a compatible branch once the first rails alpha/beta comes out.

jasnow:
@lucasmazza - Thanks for the update.

maratgaliev:
@lucasmazza Thanks!

twalpole:
@lucasmazza I have been occasionally rebasing off master, and making sure all the tests still pass using rails 5 master -  Are you saying you would prefer me to stop doing that in this PR ?   I had also left the deprecations due to maintaining compatibility with older rails/rubies -- If it's acceptable to drop some prior rails or ruby versions in this PR (< ruby 2.0.0 for instance) I can relatively easily cleanup most of the deprecation notices

twalpole:
As of now this PR should pass all tests on all supported versions of rails/ruby with no deprecation warnings -  if rails/rubies can be removed from the supported list then it could simplify some things a bit

lucasmazza:
@twalpole woah, I had no idea you were rebasing the branch, thank you so much. When I cut a stable branch for Rails 4/3 we can work on removing the compatibility code here.

README.md
```diff
@@ -406,7 +406,7 @@ Caution: Devise Controllers inherit from ApplicationController. If your app uses
 Devise includes some test helpers for functional specs. In order to use them, you need to include Devise in your functional tests by adding the following to the bottom of your `test/test_helper.rb` file:
 
 ```ruby
-class ActionController::TestCase
+class Devise::ControllerTestCase
```
lucasmazza: I guess this was an automatic search & replace that needs to be reverted? `Devise::ControllerTestCase` is for Devise tests only.

lib/devise/test_helpers.rb
```diff
@@ -1,6 +1,6 @@
 module Devise
   # Devise::TestHelpers provides a facility to test controllers in isolation
-  # when using ActionController::TestCase allowing you to quickly sign_in or
+  # when using Devise::ControllerTestCase allowing you to quickly sign_in or
```
lucasmazza: Same thing here regarding the `Devise::ControllerTestCase` replacement.

lib/devise/test_helpers.rb
```diff
@@ -15,8 +15,11 @@ def self.included(base)
 
     # Override process to consider warden.
     def process(*)
-      # Make sure we always return @response, a la ActionController::TestCase::Behaviour#process, even if warden interrupts
-      _catch_warden { super } || @response
+      # Make sure we always return @response, a la Devise::ControllerTestCase::Behaviour#process, even if warden interrupts
```
lucasmazza: and here :see_no_evil:

test/support/http_method_compatibility.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  class ControllerTestCase < ActionController::TestCase
+    # %w( get post patch put head delete xml_http_request
+    #           xhr get_via_redirect post_via_redirect
+    #         ).each do |method|
+    %w( get post put ).each do |method|
+      define_method(method) do |action, options={}|
+        if Rails.version >= '5.0.0'
```
lucasmazza: we could move this version check above the `define_method` so the version check is only called when defining the methods and not on each invocation.

lucasmazza:
> When I cut a stable branch for Rails 4/3 we can work on removing the compatibility code here.

Done - we have a [`3-stable`](https://github.com/plataformatec/devise/tree/3-stable) branch for the 3.5.x releases and we can drop support for Rails 3.2/4.0 (and older rubies like 1.9.3 and 2.0.0) and move forward with the Rails 5 support on Devise 4.

@twalpole if you want to work on removing support for those versions feel free, otherwise we can merge this branch with just the existing changes. I think we could squash this PR into a single commit to remove the intermediate commits from the overall git history.

twalpole:
@lucasmazza I'll rebase off current master, and then remove the support for older rails/rubies - just to clarify -- Is it going rails 5 only?  or rails 4.1, 4.2, 5.0  - and ruby  2.2 only or 2.1, 2.2?

lucasmazza:
@twalpole Rails 4.1, 4.2 and 5.0, and Ruby 2.1 and 2.2 (if we come to go Ruby 2.2+ only we can do it later).

twalpole:
@lucasmazza Ok -- I think thats probably good -- If it looks ok to you I can squash it down to one commit

devise.gemspec
```diff
@@ -16,12 +16,12 @@ Gem::Specification.new do |s|
...
   s.require_paths = ["lib"]
-  s.required_ruby_version = '>= 1.9.3'
+  s.required_ruby_version = '>= 2.1.0'
 
   s.add_dependency("warden", "~> 1.2.3")
   s.add_dependency("orm_adapter", "~> 0.1")
   s.add_dependency("bcrypt", "~> 3.0")
   s.add_dependency("thread_safe", "~> 0.1")
-  s.add_dependency("railties", ">= 3.2.6", "< 5")
+  s.add_dependency("railties", ">= 4.1.0", "< 5")
```
lucasmazza: the version constraint should be `< 5.1`, otherwise the dependency won't work with the final `5.0.0` release.

lib/generators/active_record/devise_generator.rb
```diff
@@ -75,13 +75,17 @@ def ip_column
       end
 
       def inet?
-        rails4? && postgresql?
+        postgresql?
       end
 
       def rails4?
```
lucasmazza: I guess we can finally :scissors: these `rails4?` methods, right?

lib/devise/models/authenticatable.rb
lucasmazza: I'll try to debug this one out and see if we can avoid to pass a `Parameters` object down the models in the builtin controllers - now that they aren't inheriting from `Hash` anymore there are a few subtle API differences that can be pretty annoying to debug and we could save the developers from having to deal with that in their apps.

lib/devise/test_helpers.rb
```diff
@@ -122,7 +126,8 @@ def _process_unauthenticated(env, options = {})
         # ensure the controller response is set to our response.
         @controller.response ||= @response
         @response.status = ret.first
-        @response.headers = ret.second
+        @response.headers.clear
+        ret.second.each { |k,v| @response[k]=v }
```
lucasmazza: add whitespces around the `=` attribution here.

test/integration/authenticatable_test.rb
```diff
@@ -559,7 +559,8 @@ class AuthenticationOthersTest < ActionDispatch::IntegrationTest
   test 'sign out with non-navigational format via XHR does not redirect' do
     swap Devise, navigational_formats: ['*/*', :html] do
       sign_in_as_user
-      xml_http_request :get, destroy_user_session_path, {}, { "HTTP_ACCEPT" => "application/json,text/javascript,*/*" } # NOTE: Bug is triggered by combination of XHR and */*.
+      get destroy_user_session_path, xhr: true, headers: { "HTTP_ACCEPT" => "application/json,text/javascript,*/*" } # NOTE: Bug is triggered by combination of XHR and */*.
+      # xml_http_request :get, destroy_user_session_path, {}, { "HTTP_ACCEPT" => "application/json,text/javascript,*/*" } # NOTE:Bug is triggered by combination of XHR and */*.
```
lucasmazza: :scissors: the commented `xml_http_request` lines?

test/rails_app/app/controllers/home_controller.rb
```diff
@@ -20,6 +20,8 @@ def set
   end
 
   def unauthenticated
-    render text: "unauthenticated", status: :unauthorized
+    r_opts = { status: :unauthorized}
```
lucasmazza: we are missing a whitespace before the `}` here.

devise.gemspec
twalpole: This was intentional until rails 5 is released and a test can be run against it -- I can change it if you still want me to?

twalpole:
@lucasmazza Ok -- fixed the minor style comments and squashed down to one commit

djsegal:
@twalpole: you're awesome :+1:

lib/devise/models/authenticatable.rb
lucasmazza: I'm figuring it out on how to stop passing an `ActionController::Parameters` to this method, but I'll work on it on a new PR later so we can remove this.

twalpole: I think the issue might be that the tests are passing strong parameters where the required parameters aren't actually permitted - looking into that currently

lucasmazza: What I've found is that both `DeviseController#resource_params` and the `ParameterSanitizer` are passing down instances of `ActionController::Parameters` to the model methods, and we operate on them like if they were regular HWIA - which they aren't anymore. 

I think it might be worth some refactor on the sanitizer class, but I want to deal with that after the PR, and I'm not sure if we should change the `resource_params` method as discussed back in #3162.

twalpole: I think resource_params definitely needs to be sanitized - otherwise it requires calling the to_unsafe_h on the parameters which just doesn't sound like a great idea.  If the relevant fields are permitted in resource_params (have the relevant controllers call the correct sanitizer) then it's just a matter of calling .to_h  on the ActionController::Parameters which returns the filtered HWA

lucasmazza:

marczking:
:+1: was having issues with rails5. This PR solved it :tada:

thank you!

bbnnt:
Hi sorry to bother;

`gem "devise", :github => 'plataformatec/devise', :branch => '3-stable'` still fails on my side :/ (https://gist.github.com/bbnnt/178e932ff082643f2ecf), same error (rack delegation)

`gem "devise", :github => 'twalpole/devise'`'s okay

I might not have discerned the hints you give in the comments; thx a lot for clarifying

lucasmazza:
@bbnnt `3-stable` isn't compatible with Rails 5 - you should use the `master` branch instead.

bbnnt:
@lucasmazza  thx for the quick reply!
`gem "devise", :github => 'plataformatec/devise', :branch => 'master'`

nthj:
:+1: Thank you for all your work!

davidpaulhunt:
thanks everyone, `master` is great

PR #3715, opened 2015-08-20
Use default_url_options from parent_controller

milgner:
URLs are now generated using the configured `Devise.parent_controller`, not the hardcoded `ApplicationController`, as in the rest of the library.

Also it now uses the instance method as recommended to implement in the Rails guides, not only the class-level hash. 

Only thing I wasn't sure of is how to test this. I couldn't find a test for the old implementation but when I tried implementing one, I couldn't find a way to modify the environment in a way that changed the parent controller without leaving a big mess behind. Changing `Devise.parent_controller` or modifying `ApplicationWithFakeEngine` has no effect at the time the tests are run as all classes and hierarchy have already been set up.

I don't like contributing code w/o tests so I'm open to suggestions and will amend my code accordingly.
As usual, comments are very welcome.

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -22,9 +22,12 @@ module Devise
       @respond.call(env)
     end
 
+    # Try retrieving the URL options from the parent controller (usually 
+    # ApplicationController). Instance methods are not supported at the moment,
+    # so only the class-level attribute is used.
     def self.default_url_options(*args)
-      if defined?(ApplicationController)
-        ApplicationController.default_url_options(*args)
+      if defined?(Devise.parent_controller.constantize)
+        Devise.parent_controller.constantize.try(:default_url_options) || {}
       else
         {}
       end


josevalim:
I would love to get the change that uses parent_controller but I believe I don't want to add the instsance feature because it will be a "fake" instance, that won't have the proper values one would expect to be set in the controller. Thank you.

milgner:
Good point! I see a couple of solutions: 

1. Let FailureApp inherit from parent_controller, too. Most straightforward but I'm not sure whether there may be undesired side-effects to that.
2. Provide the controller instance with all the state from the Devise controller (probably via `set_request!` et al). Could be somewhat slow and a bit hacky.
3. Don't support instance methods. Also sub-optimal since the Rails guides refer to them in a couple of places.

Any preferences?

josevalim:
1. Hrm, better not since it is not really a controller
2. Yes, definitely hacky, I don't even think it is a public API in Rails
3. Keeping as is today would be my vote. Also because we can call the failure app without even invoking a rails controllers.

milgner:
Ok, I adapted my code as discussed.

josevalim:

PR #3723, opened 2015-08-23
Replace redundant sign_in after creating new session

MarkMT:
The current code calls `sign_in` after authenticating a new user session. This stores the user in the session if it isn't already present. However `warden.authenticate!` already stores the user, so there is no need to do so again. The only other function performed by `sign_in`  that is needed after authentication is to remove old session data by calling `expire_data_after_sign_in!`.

### DIFF
--- a/app/controllers/devise/sessions_controller.rb
+++ b/app/controllers/devise/sessions_controller.rb
@@ -16,7 +16,8 @@ class Devise::SessionsController < DeviseController
   def create
     self.resource = warden.authenticate!(auth_options)
     set_flash_message(:notice, :signed_in) if is_flashing_format?
-    sign_in(resource_name, resource)
+    # remove any previous data stored in the session with keys starting with 'devise.'
+    expire_data_after_sign_in!
     yield resource if block_given?
     respond_with resource, location: after_sign_in_path_for(resource)
   end


josevalim:
Thank you but I don't agree the code is redundant. In the future, we may add new operations to sign_in and we will certainly forget to update `create`. It is best if sign_in is still called to execute all properties related to sign in. And that is not considering users themselves can override it.

MarkMT:
No problem. Thanks for giving consideration to the suggestion. I think my broader concern was that the semantics of the sign in process seem kind of fuzzy. When I first studied the code I saw `sign_in` in the `create` action and discovered that it stores the user in the session, so I figured that that was essentially what 'signing in' meant here. But then I realized that storing the user actually happens elsewhere, so I was left feeling unsure conceptually what this thing is that is happening after authentication takes place (of course on sign-up it makes more sense). 

Since this functionality is exposed in the application, it would be good if there was a clearer way of expressing this. At the moment `sign_in` seems like kind of a hold-all for operations performed on the session after *either* signing in or signing up. *Personally* I think I would favor some refactoring so that `set_user` does not appear in the method called by the controller after authenticating a new session, but only on signing up a new user. Just my $0.02.

MarkMT:
In case anyone else stumbles across this thread, it may be worth mentioning that the simple_token_authentication gem is an example of a third party library that utilizes the `sign_in` method to store the user in the session *without* having first authenticated the user via Warden's Proxy#authenticate! method:

https://github.com/gonzalo-bulnes/simple_token_authentication/blob/master/lib/simple_token_authentication/sign_in_handler.rb

PR #3730, opened 2015-08-28
use OmniAuth::Utils.camelize

timoschilling:
This will display the provider names with the original camelization:
```
facebook => Facebook
github => GitHub
linkedin => LinkedIn
```
We do the same already in the [`OmniauthCallbacksController `](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/omniauth_callbacks_controller.rb#L9).

### DIFF
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -20,6 +20,6 @@
 
 <%- if devise_mapping.omniauthable? %>
   <%- resource_class.omniauth_providers.each do |provider| %>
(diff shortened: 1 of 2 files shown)


reedloden:
The comment here says Facebook, but the commit uses FaceBook (which is incorrect).

timoschilling:
@reedloden that's just on example. The description on top shows the right camelizations. The tests are using a facke camelization, otherwise I had to rewrite all test from 'Facebook' to 'GitHub', while 'Facebook' don't have a custom camelization.

PR #3732, opened 2015-08-29
Wrap helper_method calls in respond_to?(:helper_method)

posgarou:
Currently `Devise::Controllers::Helpers` depends on `#helper_method`, which is defined inside `AbstractController::Helpers`.  This module is included in `ActionController::Base`, but it is not included by default in `ActionController::Metal`.

`#helper_method` is not essential to the functionality of the methods inside `Devise::Controllers::Helpers`, so this PR wraps all Devise `#helper_method` calls with `respond_to?(:helper_method)` checks.

### DIFF
--- a/test/controllers/helper_methods_test.rb
+++ b/test/controllers/helper_methods_test.rb
@@ -8,14 +8,14 @@ class HelperMethodsTest < ActionController::TestCase
   tests ApiController
 
   test 'includes Devise::Controllers::Helpers' do
-    assert @controller.class.ancestors.include?(Devise::Controllers::Helpers)
+    assert_includes @controller.class.ancestors, Devise::Controllers::Helpers
   end
 
   test 'does not respond_to helper_method' do
-    refute @controller.respond_to?(:helper_method)
+    refute_respond_to @controller.class, :helper_method
   end
 
   test 'defines methods like current_user' do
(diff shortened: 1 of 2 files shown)


test/controllers/helper_methods_test.rb
```diff
@@ -0,0 +1,21 @@
...
+    assert @controller.class.ancestors.include?(Devise::Controllers::Helpers)
+  end
+
+  test 'does not respond_to helper_method' do
+    refute @controller.respond_to?(:helper_method)
+  end
+
+  test 'defines methods like current_user' do
+    assert @controller.respond_to?(:current_user)
+  end
```
carlosantoniodasilva: Please use the builtin methods: `assert_includes`, `assert_respond_to` and `refute_respond_to`. They provide nicer error messages in case of test failures.

posgarou: You're right, thank you.

betesh:
+1

ballPointPenguin:
This seems like a good solution, rather than adding mock support for this method in Rails 5 api code.

posgarou:
Any updates on this PR?  Thanks!

lucasmazza:
Sorry, for the delay, thanks for the fix :heart:

posgarou:
Thanks!

PR #3739, opened 2015-09-06
Account for relative_url_root in FailureApp's recall method

### LINKED ISSUE #3738: `failure_app::recall` doesn't take into account relative_url_root
When a login failure occurs, Warden sends back a full path in `attempted_path`, but devise's `failure_app` assumes the path does not contain a relative URL root and assigns it to `PATH_INFO`:

```ruby
  def recall
      env["PATH_INFO"]  = attempted_path
      flash.now[:alert] = i18n_message(:invalid) if is_flashing_format?
      self.response = recall_app(warden_options[:recall]).call(env)
    end
```

For example, suppose:

1.  `relative_url_root` = `/base_app`  
2. `attempted_path` = `/base_app/resource`.
3. In this case, `PATH_INFO` is assigned `/base_app/resource`.
4. However, `SessionController::store_redirect_path` will call `request.fullpath`, which then makes the path `/base_app/base_app/resource`. Note the redundant `base_app`.

`PATH_INFO` should only contain the part of the path that excludes the relative URL root, since `SCRIPT_NAME` includes the relative URL part (see http://api.rubyonrails.org/classes/ActionDispatch/Routing/UrlFor.html) for more details.

stanhu:
Closes #3738

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -51,7 +51,18 @@ module Devise
     end
 
     def recall
-      env["PATH_INFO"]  = attempted_path
+      config = Rails.application.config
+
+      if config.try(:relative_url_root)
+        base_path = Pathname.new(config.relative_url_root)
+        full_path = Pathname.new(attempted_path)
+
+        env["SCRIPT_NAME"] = config.relative_url_root
+        env["PATH_INFO"] = '/' + full_path.relative_path_from(base_path).to_s
+      else
+        env["PATH_INFO"]  = attempted_path
+      end
+
       flash.now[:alert] = i18n_message(:invalid) if is_flashing_format?
       self.response = recall_app(warden_options[:recall]).call(env)
     end
--- a/test/failure_app_test.rb
+++ b/test/failure_app_test.rb
@@ -294,5 +294,22 @@ class FailureTest < ActiveSupport::TestCase
       assert @response.third.body.include?('<h2>Log in</h2>')
       assert @response.third.body.include?('Your account is not activated yet.')
     end
+
+    if Rails.application.config.respond_to?(:relative_url_root)
+      test 'calls the original controller with the proper environment considering the relative url root' do
+        swap Rails.application.config, relative_url_root: "/sample" do
+          env = {
+            "warden.options" => { recall: "devise/sessions#new", attempted_path: "/sample/users/sign_in"},
+            "devise.mapping" => Devise.mappings[:user],
+            "warden" => stub_everything
+          }
+          call_failure(env)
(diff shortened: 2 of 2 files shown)


MichielDeMey:
+1

stanhu:
@josevalim Can you please review this? Thank you.

chrismacias:
it seems that this fixes the issue

stanhu:
Friendly ping: @carlosantoniodasilva @josevalim, can one of you review?

test/failure_app_test.rb
```diff
@@ -294,5 +294,22 @@ def call_failure(env_params={})
       assert @response.third.body.include?('<h2>Log in</h2>')
       assert @response.third.body.include?('Your account is not activated yet.')
     end
+
+    if Rails.application.config.respond_to?(:relative_url_root)
```
josevalim: Is this required to support different rails versions?

lib/devise/failure_app.rb
```diff
@@ -51,7 +51,18 @@ def http_auth
     end
 
     def recall
-      env["PATH_INFO"]  = attempted_path
+      config = Rails.application.config
+
+      if config.try(:relative_url_root)
+        base_path = Pathname.new(config.relative_url_root)
```
josevalim: Can we not use `Pathname` and call the methods in `File` or `Path` directly? I always find `Pathname` very error prone, specially if you forget to call `to_s`.

josevalim:
Sorry for the delay! I have added some comments. :) Thank you!

lib/devise/failure_app.rb
stanhu: The method makes use of Pathname's `relative_path_from` method. Is there a better substitute? As far as I can tell, there isn't anything else.

test/failure_app_test.rb
stanhu: I was under the impression `relative_url_root` could be left blank in the application config.

lib/devise/failure_app.rb
josevalim: You are right, thank you.

test/failure_app_test.rb
josevalim: Yes, but does it matter for Devise own test suite?

stanhu: There are a number of test cases that use this pattern. Should I remove them all?

josevalim: If there are other tests relying on this pattern you have answered my question, it is for Rails compat. We can probably remove it in another PR if the build passes after doing so. :)

josevalim:

stanhu:
Thanks, @josevalim!

PR #3754, opened 2015-09-16
disable validates_length_of unless password_required?

omotenko:
Hi, guys!
My actions:
  1. I created User without password.
  2. Included :validatable module to User.
  3. Want to disable password validation.
  4. Overrode password_reqired? to FALSE.
  5. Got Error 'undefined method `password' for #<User:0x007fa193fe3148>'

This commit will fix it.


josevalim:
If you want to customize validations, it is better to drop the validatable module and roll your own. The password_required? is not about disabling password, but checking if the password field should be checked at that moment or not.

Thank you for the PR anyway!

PR #3770, opened 2015-10-01
Avoid self where not required

xymbol:
Avoid useless uses of `self`.


josevalim:
Thank you but this is mostly a code style change and we avoid code style changes as it messes up history and what not. For example, someone in a month may add them back because they prefer `self.` as it is more explicit.

xymbol:
@josevalim Sure, though this seems to be a [generally accepted](https://github.com/bbatsov/rubocop/blob/f6ad364be3f53b590eda1d46dc236b314cde0760/lib/rubocop/cop/style/redundant_self.rb) practice. Thanks!

PR #3774, opened 2015-10-02
Only clear the reset password token if the model has already been persisted

stanhu:
If a new user is created with a reset password token, the previous behavior would automatically clear the token even when it were desired for setting the password for the first time.

### DIFF
--- a/lib/devise/models/recoverable.rb
+++ b/lib/devise/models/recoverable.rb
@@ -32,7 +32,7 @@ module Devise
 
       included do
         before_save do
-          if email_changed? || encrypted_password_changed?
+          if persisted? && (email_changed? || encrypted_password_changed?)
             clear_reset_password_token
           end
         end
--- a/lib/devise/models/recoverable.rb
+++ b/lib/devise/models/recoverable.rb
@@ -31,10 +31,16 @@ module Devise
       end
 
       included do
+        def expire_reset_token?
+          # Expire the reset token only if the e-mail or password were changed
+          # since the last time the record was saved to the database. An admin
+          # may want to retain the token to give the newly-created user a chance
+          # to set the password for the first time.
+          persisted? && (email_changed? || encrypted_password_changed?)
+        end
+
         before_save do
-          if persisted? && (email_changed? || encrypted_password_changed?)
-            clear_reset_password_token
-          end
+          clear_reset_password_token if expire_reset_token?
         end
       end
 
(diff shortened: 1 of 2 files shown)


stanhu:
@josevalim, can you review this? Thank you.

stanhu:
Friendly ping: @carlosantoniodasilva @josevalim, can one of you review?

lucasmazza:
@stanhu can you please expand on your use case for this fix? When it makes sense to call `send_reset_password_instructions` with an unpersisted user?

stanhu:
@lucasmazza, this is the flow of events in GitLab:

1. Create new user.
2. Set the e-mail of new user.
3. Set the reset token so that user can change the password for the first time.
4. Save user.
5. Send an e-mail to the user with the password reset link.

The problem is that Devise insists on clearing the reset token since step 2 occurred, so the reset token is no longer valid.

stanhu:
It looks like others are running into this use case:

See: #3787, #3782, #3673

lucasmazza:
Thanks for the extended report, sounds like a reasonable fix to make :+1:.

lib/devise/models/recoverable.rb
```diff
@@ -32,7 +32,7 @@ def self.required_fields(klass)
 
       included do
         before_save do
-          if email_changed? || encrypted_password_changed?
+          if persisted? && (email_changed? || encrypted_password_changed?)
```
lucasmazza: What do you think of wrapping this up into an instance method with a piece of documentation explaining the semantics of all these checks at once? I think this will help out whenever we need to revisit this later.

josevalim:
:+1: for this.

stanhu:
Done!

stanhu:
@josevalim, can you take a look?

jjarmoc:
Why doesn't the HMAC just include the email address under the reset token HMAC when it's generated?  

That way if the email doesn't match what was present when the token was generated, the calculated HMAC won't match the stored value and reset fails. We avoid all the logic around tracking the changes, and while it's probably still wise to clear the reset token when the email address is modified, it might not be strictly necessary.

lucasmazza:
@stanhu thanks! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

lucasmazza:
@jjarmoc I guess it would make sense to use the email address on the token but I wonder if I guess we would have to deal with supporting the tokens that were generated without the email - but @josevalim might know better about this piece of the codebase.

stanhu:
Thaks, @lucasmazza! When will a new release for devise be available?

lucasmazza:
@stanhu I'll cut `3.5.3` later today.

stanhu:
@lucasmazza, thanks! Looking forward to 3.5.3.

PR #3777, opened 2015-10-07
Add gem version to README

AlexVPopov:


josevalim:
I think we had this badge but we removed it in order to keep things simple some time ago. :) Thank you though!

PR #3778, opened 2015-10-12
Drop unneeded argument

xymbol:
Don't pass attribute values as method arguments.

### DIFF
--- a/app/controllers/devise/confirmations_controller.rb
+++ b/app/controllers/devise/confirmations_controller.rb
@@ -9,7 +9,7 @@ class Devise::ConfirmationsController < DeviseController
     self.resource = resource_class.send_confirmation_instructions(resource_params)
     yield resource if block_given?
 
-    if successfully_sent?(resource)
+    if successfully_sent?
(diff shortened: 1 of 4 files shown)


josevalim:
Thank you! The biggest issue is that this is a backwards incompatible change, so although it is technically correct, it can break existing code and therefore we shouldn't merge it as there are no practical gains.  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

xymbol:
@josevalim No worries but, just curious. These are `protected` methods, not the public API. The practical gain of incremental refactoring is maintaining less code, reducing complexity and, saving others from the "why is this here?" moment. OTOH, you're concerned about users calling these without any coverage, disregarding versions, etc. Would you rather accept these as deprecations instead?

josevalim:
@xymbol sometimes folks just copy controllers whole sale. And although they should not be doing do that, they certainly do it and they would curse me.

I also don't agree that it reduces complexity. I generally prefer to pass the arguments a method is relying on explicitly instead of relying on the inner controller state. While one could argue which one is the common practice for Ruby, I wouldn't argue about complexity. :)

PR #3786, opened 2015-10-18
Remove ActionController::RackDelegation for Rails 5 support

maratgaliev:
First fixes for Rails 5 support.


lucasmazza:
Hey,

I'm going to close this in favor of working on top of #3714 (please see https://github.com/plataformatec/devise/pull/3714#issuecomment-149254525 for more info) to deal with all of the Rails 5 deprecations and changes - we already have the test suite required setup (although we haven't tested with more recent commits on rails head) and we will take the `RackDelegation` removal into account when we revisit that branch - but thanks anyway for the Pull Request :heart:.

PR #3788, opened 2015-10-19
Replace deprecated prepend_before_filter and before_filter

maratgaliev:
Only for Rails 4+


seuros:
That will break the build for Rails 3.2

lucasmazza:
Thanks for your Pull Request, but Devise master still supports Rails 3.2 which does not have `before_action`, so we will stick with `before_filter` until we make the move to Rails 4/5+ (see #3714).

PR #3789, opened 2015-10-20
Custom Error messages

tauhidul35:
Most of time developers needed to customize devise error messages. In this case they needed to override helper method. As so I customize the devise_error_messages! helper method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,7 +405,7 @@ Caution: Devise Controllers inherit from ApplicationController. If your app uses
 
 Devise includes some options to customise devise error messages. There are two options -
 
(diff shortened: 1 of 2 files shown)


rubyrider:
CircleCI has failed to build.

lucasmazza:
Thanks for the Pull Request but in cases like this [we recommend](https://github.com/plataformatec/devise/blob/master/app/helpers/devise_helper.rb#L2-L7) developers to override this helper or implement their own error rendering with application specific views instead of adding extra behaviour to the `devise_error_messages!` method.

Issue #3790, opened 2015-10-21
Automatically pluralize scoped views generator


excid3:
I ran into the same gotcha that these guys hit in #2536 where the generator accepts singular names and doesn't pluralize. Since the gem's configured to always look for the pluralized name, does it make sense to fix this generator to auto-pluralize the scoped views folders?

Happy to make a PR for this, just wanted to check and make sure that it wasn't intentional to not auto-pluralize.

josevalim:
Pluralizing in the views generator? That sounds great, thank you!

PR #3792, opened 2015-10-21
Automatically underscore and pluralize scoped views generator

excid3:
This fixes issue #3790.

Example of previous functionality. This basically just doesn't match up with what Devise expects when rendering the views:
```bash
rails g devise AdminUser
rails g devise:views AdminUser # incorrectly generates the folder name
=>      create    app/views/AdminUser/shared
```

When rendering the scoped views, Devise will always look for the underscored and pluralized folder name, so this trips up a few people.

Fixed version:

```bash
rails g devise AdminUser
rails g devise:views AdminUser # correctly underscores and pluralizes the scope
=>      create    app/views/admin_users/shared
```

### DIFF
--- a/lib/generators/devise/views_generator.rb
+++ b/lib/generators/devise/views_generator.rb
@@ -47,7 +47,7 @@ module Devise
       def view_directory(name, _target_path = nil)
         directory name.to_s, _target_path || "#{target_path}/#{name}" do |content|
           if scope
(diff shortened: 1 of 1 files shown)


rodrigoargumedo:
+1

ryanckulp:
+1

Davidzhu001:
+1

PR #3793, opened 2015-10-22
save in after_create hook breaks devise confirmation

chrismacias:
the confirmable.rb module has a before_create method that generates a confirmation token and a before_update method that calls method :postpone_email_change_until_confirmation_and_regenerate_confirmation_token

Apparently, the second method, recognizes that the email changes from empty string to the one specified and triggers true for method email_changed?

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -254,7 +254,7 @@ module Devise
         end
 
         def postpone_email_change?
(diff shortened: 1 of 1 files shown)


chrismacias:
can you take a look at this please @josevalim

lucasmazza:
@chrismacias can you please write a test case that replicates the issue that we are fixing here?

Issue #3643, opened 2015-06-21
Redirect loop when using `authenticate` in routes.rb (plus pull request that caused it)


stevenmilton58:
Rails 4.2.2, Devise 3.5.1, Warden 1.2.3

**Note**: If  I revert the changes made in this pull request, things work as expected: https://github.com/plataformatec/devise/pull/3452/files

**Description**
I have an Admin model and am trying to use it to authenticate access to Sidekiq with Devise's "authenticate" method in Rails routes.

routes.rb :

    devise_for :admins

    authenticate :admin do
      require 'sidekiq/web'
      mount Sidekiq::Web => "/admin/sidekiq"
    end

I get a "redirect loop" error.

Upon visiting `http://localhost:3000/admin/sidekiq` I get redirected to 
`http://localhost:3000/admin/sidekiq/admins/sign_in` instead of being redirected
to `http://localhost:3000/admins/sign_in`.

Not sure how to fix this.

I see that the route being generated in `def scope_url` in failure_app.rb (https://github.com/plataformatec/devise/blob/master/lib/devise/failure_app.rb#L129)
is `http://localhost:3000/admin/sidekiq/admins/sign_in` instead of the expected
`http://localhost:3000/admins/sign_in`.

abevoelker:
I'm having the same problem. According to git bisect this is the commit that broke it: af8d38e45b24ad06f89af254c4c0dc4c6c39f723

josevalim:
Can someone please send a pull request? Apparently we will need to detect the Rails version because the fix is required for Rails 4.2.

pragmaticivan:
I'm having the same problem with Rails 4.2.3. Is there any plans for accepting the last PR?

sloops77:
The fix doesnt apply to rails below 4.2 but is an issue on our app with 4.1.12 (and is fixed by the above commit simply by changing the minimum minor_version to 1.

PR #3807, opened 2015-11-04
Fix infinite redirect fix that broke engine routes

cipater:
In Rails 4.2 it appears that not setting :script_name causes routes to
be computed within mount points that don’t actually contain a devise
scope (see #3643). Commit aa675f7 unfortunately sets :script_name to
nil for Rails 4.2, preventing routes from being computed in their
correct contexts.

This commit recovers the :script_name from env should the found context
be that of an engine allowing correct route generation, while still
setting :script_name to nil in Rails 4.2 for all other cases.

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -22,7 +22,7 @@ module Devise
       @respond.call(env)
     end
 
-    # Try retrieving the URL options from the parent controller (usually 
+    # Try retrieving the URL options from the parent controller (usually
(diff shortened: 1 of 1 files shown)


cipater:
Three tests are falling because those tests mock a "context" object that does not respond to the routes method. Trying to figure the best way to update the tests...

ulissesalmeida:
Hi @cipater . Did you have any progress with tests?

lucasmazza:
Merged a similar fix from #4081, so we can close this for now.

PR #3808, opened 2015-11-04
Allow skipping the timeoutable hook

blase:
I was working on a "session timer" functionality in my current project and making requests without changing `last_request_at` was very useful. What do you think?


josevalim:

PR #3810, opened 2015-11-06
Allow passwords

djGrill:
`.present?` returns `false` for whitespace-only strings.

Since passwords with only whitespaces are also valid, we should allow them.


josevalim:
Thank you but I would rather not allow them still. An all blank password would be a mistake for 99.99% of the users.

PR #3813, opened 2015-11-10
set minimum password length on reset failure

jlerpscher:
On the reset password form, if the form is invalid, the `@minimum_password_length` variable is not set. This PR fix that by calling `set_minimum_password_length` in the update method. I can add some tests if needed :)


josevalim:

PR #3817, opened 2015-11-11
Add scoped sessions to TestHelpers module

AnomalousBit:
Hello!

I've added support for scoped sessions to the `TestHelper` module. Many applications use the `user_session` or `admin_session`, and the code in this branch should allow people to cover more of that functionality inside of their tests.

I've included several basic tests in `test_helpers_tests.rb` to demonstrate usage and guarantee scoped sessions can only be used in tests once a `sign_in` has occurred. 

Thanks for the great Gem! Hope my additions are helpful.

### DIFF
--- a/lib/devise/test_helpers.rb
+++ b/lib/devise/test_helpers.rb
@@ -13,6 +13,16 @@ module Devise
       end
     end
 
+    def self.define_helpers(mapping)
+      mapping ||= :user
+
+      class_eval <<-METHODS, __FILE__, __LINE__ + 1
+        def #{mapping}_session
+          warden.session(:#{mapping})
+        end
+      METHODS
+    end
+
     # Override process to consider warden.
     def process(*)
       # Make sure we always return @response, a la ActionController::TestCase::Behaviour#process, even if warden interrupts
@@ -22,6 +32,7 @@ module Devise
     # We need to setup the environment variables and the response in the controller.
     def setup_controller_for_warden #:nodoc:
       @request.env['action_controller.instance'] = @controller
+      TestHelpers::define_helpers(:user)
     end
 
     # Quick access to Warden::Proxy.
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you but I would push for those to be accessed rather from the controller instead of extending the test API. :) Glad you are enjoying Devise!

AnomalousBit:
Hi Jose,

What you describe is exactly my situation, maybe my explanation for adding scoped sessions to tests was insufficient.

I am currently using `user_session` in several of my controllers and would not only like to assert that functionality inside of my functional tests, I cannot get my test cases to pass (as planned) without adding data to user_session inside of my test cases. 

At present, user_session is not available from `ActiveController::TestCase` even when adding the module via `include Devise::TestHelpers`. Maybe this test case below will make more sense.


**demo_controller_test.rb**
```
class DemoControllerTest < ActionController::TestCase
    include Devise::TestHelpers

    test "will fail because user_session raises NoMethodError" do

         # without my branch, this call to user_session is not possible, NoMethodError is raised.
        user_session['required_property'] = true

        # index requires user_sessions['required_property'] or else the action will respond with failure
        get :index
        assert_response :success
    end

end
```

**demo_controller.rb**
```
class DemoController < ActionController::Base

    def index
        if user_session['required_property'].nil?
            redirect_to error_path and return
        end
    end

end
```


Is there another way to populate user_session from a test case without having to basically implement integration testing to flow through my usual login and session initialization process? That seems very cumbersome when I can just set the user_session manually to provide accurate test coverage for my controllers.

Thanks again!
Jason

AnomalousBit:
I misunderstood @josevalim's response above and was fortunate enough to have @lucasmazza point out that scoped sessions can be easily accessed via the controller instance being tested. Simply referring to `@controller.user_session` in your Rails / MiniTest `TestCase` permits the session handling required to get full test coverage. Thanks again to both of you guys.

Issue #3787, opened 2015-10-19
save in after_create hook breaks devise confirmation


krukgit:
Using `save` in `after_create` hook on devise model breaks confirmation link. 

Test repo: 
https://github.com/krukgit/devise_confirmation_issue

Steps to reproduce:
1. Run rails server: `rails s`
2. Navigate to signup page: `http://localhost:3000/users/sign_up`
3. Fill in the form, submit. Letter opener will open the email with confirmation link.
4. Click on the link

Result:
`Confirmation token is invalid` error

Expected result:
Succesful confirmation.

Tested with:
Rails 4.2.4
Devise 3.5.2

Related discussion:
https://github.com/plataformatec/devise/issues/2615

chrismacias:
the confirmable method in Devise creates the confirmation token when user is created and since you call the break_confirmation method after the creation of the user it creates a new confirmation token at self.save.
I recommend using the before_create and just leave out the self.save since its implicitly created 

user.rb
```ruby
before_create :break_confirmation

  def break_confirmation
    self.role = 'user'
  end
```

wdas03:
sad

chrismacias:
why is it sad?

wdas03:
<script type="text/javascript">
alert("hello");
</script>

bingxie:
I have the same issue.

lucasmazza:
@krukgit @bingxie we got #3774 merged with a fix that might be related to this issue - can you folks take a look and see if it works? If it doesn't, we would need more details on how to replicate this issue so we can dig deeper in the problem.

bingxie:
@lucasmazza I updated to the 3.5.3 version, the after_create got called. But I did update_attributes(slug: slug) in the after_create block, the new problem is it didn't send the confirmation email.

lucasmazza:
@bingxie I've merged another relevant fix + test that fixes the issue replicated in @krukgit sample app - the patch is on the `3-stable` - can you try it out so I can cut a `3.5.4` with the fix?

bingxie:
@lucasmazza I tried 3-stable branch, yes, it fixed the issue. Thanks!

justin808:
[...]

PR #3835, opened 2015-11-24
test for save in after_create hook breaks devise confirmation [3787]

kv109:
Hi!

I would like to start contributing to Devise as a thank you for a great project. 
I've noticed [this](https://github.com/plataformatec/devise/pull/3793) PR. It fixes https://github.com/plataformatec/devise/issues/3787, but has [no tests](https://github.com/plataformatec/devise/pull/3793#issuecomment-156698377). I thought that adding this test could be a good start. So there it is.


lucasmazza:
@kv109 thanks a lot - I tested both branches locally and your test fits the patch from #3787.

PR #3846, opened 2015-12-07
Allow resources with no email field to be recoverable

seddy:
@josevalim - raising this PR folling the discussions on [this commit](https://github.com/plataformatec/devise/commit/e641b4b7b97159054b7d92fb14df557ac18ae6f4)

The current implementation is opinionated in that the resource should have
an "email" column on it if it is to be recoverable, which isn't
necessarily the case.  For example, developers may decide to pull emails
out into their own model or have some other way of communicating
password resets to their users (e.g. text message)

I'm not sure there's an easy test to put together for this case, as
minitest doesn't make it very easy to stub the "email_changed?" to raise
an error. Happy to look into building another model in the
"test/rails_app" if you want to have this properly tested though? Or for
a nice way to get calls to "email_changed?" to raise; minitest isn't
a test framework I'm overly familiar with :).

As a side note, it would be nice if the Validatable module also took
this into account, I may raise another PR for that.

### DIFF
--- a/lib/devise/models/recoverable.rb
+++ b/lib/devise/models/recoverable.rb
@@ -28,7 +28,7 @@ module Devise
 
       included do
         before_save do
-          if email_changed? || encrypted_password_changed?
+          if (respond_to?(:email_changed?) && email_changed?) || encrypted_password_changed?
             clear_reset_password_token
           end
         end
--- a/test/models/recoverable_test.rb
+++ b/test/models/recoverable_test.rb
@@ -65,6 +65,18 @@ class RecoverableTest < ActiveSupport::TestCase
     assert_nil user.reset_password_token
   end
 
+  test 'should clear reset password successfully even if there is no email' do
+    user = create_user_without_email
+    assert_nil user.reset_password_token
+
+    user.send_reset_password_instructions
+    assert_present user.reset_password_token
+    user.password = "123456678"
+    user.password_confirmation = "123456678"
+    user.save!
+    assert_nil user.reset_password_token
+  end
+
   test 'should not clear reset password token if record is invalid' do
     user = create_user
     user.send_reset_password_instructions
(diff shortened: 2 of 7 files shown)


josevalim:
Can we have a test somehow? :) Maybe explicitly undefined email_changed? in a new model and show it won't blow up? Thank you!

josevalim:
Sorry, I just now I noticed your comment. You can maybe write a test like this:

    u = User.new
    u.singleton_class.send :remove_method, :==
    u.function_that_triggers_the_code_above

I am not sure if it will work though as Rails usually redefines those things on the fly.

seddy:
Thanks @josevalim - I'll try and get that sample code working, if not I'll fallback to a new model (had tried that initially but it felt a bit overkill so thought I'd see what you thought first).  Will update tomorrow :)

josevalim:
Ah, we may already have a model without an e-mail. Search for respond_to? in other places in the code, we may likely have a test or a model for it.

seddy:
@josevalim - I've added a test for this case, can you take a look and let me know what you think?  I ended up going creating a new model for it.  I'm not sure how clear my comment is [here](https://github.com/plataformatec/devise/pull/3846/files#diff-5c2e492757012dd9ee573bce25d95c17R13):

```
  # This test stub is a bit rubbish because it's tied very closely to the
  # implementation where we care about this one case. However, completely
  # removing the email field breaks "recoverable" tests completely, so we are
  # just taking the approach here that "email" is something that is a not an
  # ActiveRecord field.
```

More than happy to reword it if you've got any thoughts on it?  

I would rather have created a whole new model which just doesn't have an `email` field, which I tried initially but found myself shaving a yak, so thought that for the scope of this change simply handling the specific `respond_to?` and `email_changed?` methods could be enough.

As an aside, I was thinking I'd raise another PR around the `Validatable` module, because it also requires an email, but I don't see why it should (if, for example, you just allow authentication by username), so I can improve this mocked `UserWithoutEmail` more effectively when doing that perhaps?

josevalim:
Thank you! :heart:

We don't need to worry about validable because it is mostly meant to be a stub. If you want other validations, because for example you don't have the e-mail field, it is easier to  just drop validatable.

seddy:
Fair enough :+1:

Thanks @josevalim :)!

PR #3847, opened 2015-12-09
Only clear the reset token on an update

stanhu:
This solves the issue where a package might do:

```ruby
user = User.new
user.email = 'test@test.com'
token = user.generate_reset_token
user.save

send_reset_email(token)
```

Since the save clears the reset token, the user will receive a stale token that no longer works.

Closes #3774


josevalim:

PR #3849, opened 2015-12-11
fa (Iran/Persian) Translation

intuxicated:


nashby:
@intuxicated thanks for the pull request but we don't add translations to this repository. But you can add it to the https://github.com/tigrish/devise-i18n/ if it's not already there. Thanks!

intuxicated:
oops!! my bad.

PR #3853, opened 2015-12-15
Removed unnecessary module fixes #3852

### LINKED ISSUE #3852: module included for no reason
Unless I'm missing something, DateHelper is not necessary here: https://github.com/plataformatec/devise/blob/master/lib/devise/models/confirmable.rb#L43

Since the only method that seems to use it in that module is time_ago_in_words, which is called on  `Devise::TimeInflector` and not on `self`.  Can we safely remote the ` include ActionView::Helpers::DateHelper` from devise/confirmable.rb?

betesh:


betesh:
#3852

PR #3856, opened 2015-12-15
Remove `ParametersSanitizer` inheritance.

lucasmazza:
We no longer need to support the `BaseSanitizer` implementation for apps without
the Strong Parameters API, and this section is lacking a minimal set of
docs to document the expected behavior besides the `README` section.

An important change here is that `sanitize` now returns a `HashWIthIndifferentAccess` instead of an `ActionController::Parameters` as it doesn't inherits from `Hash` anymore, and our internals rely on it being a `Hash` that we can operate on (This is one of the headaches from supporting Rails 5).

I didn't documented the `for` method as I believe its API to be quite confusing at the moment. It can accept a block that is stored in the Sanitizer instance and it can mutate an existing `Array` and return it. I think we should normalize it and add a new method for customizing the sanitizer, and deprecate `for`.
My first sketch would be use `permit` instead (it is currently a private shortcut to `ActionController::Parameters#permit`).

```ruby
devise_parameter_sanitizer.permit(:sign_up, attributes: %i(firstname last name))
# or `keys` instead of `attributes`.
devise_parameter_sanitizer.permit(:sign_up, keys: %i(firstname last name))

# Block form.
devise_parameter_sanitizer.permit(:sign_up) do |user|
  user.permit(nested: [:foo, :bar])
end
```

### DIFF
--- a/lib/devise/parameter_sanitizer.rb
+++ b/lib/devise/parameter_sanitizer.rb
@@ -1,99 +1,208 @@
 module Devise
-  class BaseSanitizer
-    attr_reader :params, :resource_name, :resource_class
+  # The +ParameterSanitizer+ deals with permitting specific parameters values
+  # for each +Devise+ scope in the application.
+  #
+  # The sanitizer knows about Devise default parameters (like +password+ and
+  # +password_confirmation+ for the `RegistrationsController`), and you can
+  # extend or change the permitted parameters list on your controllers.
+  #
+  # === Permitting new parameters
+  #
+  # You can add new parameters to the permitted list using the +permit+ method
+  # in a +before_action+ method, for instance.
+  #
+  #    class ApplicationController < ActionController::Base
+  #      before_action :configure_permitted_parameters, if: :devise_controller?
+  #
+  #      protected
+  #
+  #      def configure_permitted_parameters
+  #        # Permit the `subscribe_newsletter` parameter along with the other
+  #        # sign up parameters.
+  #        devise_parameter_sanitizer.permit(:sign_up, keys: [:subscribe_newsletter])
+  #      end
+  #    end
+  #
+  # Using a block yields an +ActionController::Parameters+ object so you can
+  # permit nested parameters and have more control over how the parameters are
+  # permitted in your controller.
+  #
+  #    def configure_permitted_parameters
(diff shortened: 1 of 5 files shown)


lib/devise/parameter_sanitizer.rb
```diff
@@ -1,99 +1,153 @@
...
-    # TODO: We do need to flatten so it works with strong_parameters
-    # gem. We should drop it once we move to Rails 4 only support.
+    # Cast a sanitized +ActionController::Parameters+ to a +HashWithIndifferentAccess+
+    # that can be used elsewhere.
+    #
+    # Returns an +ActiveSupport::HashWithIndifferentAccess+.
+    def cast_to_hash(params)
+      if params
+        # Rails 4.1 doesn't have `to_unsafe_h`, only `to_hash`.
+        hash = params.respond_to?(:to_unsafe_h) ? params.to_unsafe_h : params.to_hash
```
twalpole: It looks like this is only called with parameters that have already been permitted - in which case just calling .to_h would be better

josevalim:
:shipit:

lib/devise/parameter_sanitizer.rb
lucasmazza: Yeah :+1:.

lucasmazza:
@josevalim I've pushed the changes for the new API and the deprecation messages for the old one showcasing how to upgrade: tomorrow I'll rewrite the docs to use the new methods.

lib/devise/parameter_sanitizer.rb
```diff
@@ -1,99 +1,179 @@
...
-
-    def default_sanitize(kind)
-      default_params
+    # Cast a sanitized +ActionController::Parameters+ to a +HashWithIndifferentAccess+
+    # that can be used elsewhere.
+    #
+    # Returns an +ActiveSupport::HashWithIndifferentAccess+.
+    def cast_to_hash(params)
+      if params
+        params.to_hash.with_indifferent_access
```
twalpole: to_hash returns all parameters not just the permitted ones - to_h returns a HWIA with only the allowed parameters.

lucasmazza:
Pushed the docs updates for the sanitizer class and the README examples.

@twalpole thanks for the on point comments.

PR #3857, opened 2015-12-15
Fix Rails 5 test

twalpole:
This removes a test that was added for a Rails 5 behavior which was deemed to be a bug -  fixed here - https://github.com/rails/rails/pull/22564


twalpole:
@lucasmazza This fixes the failing Rails 5 test

PR #3858, opened 2015-12-15
Explicitly permit rather than using to_unsafe_h

twalpole:
When implementing the rails 5 updates I had used to_unsafe_h in some methods.  This changes those instances to explicitly permit expected parameters instead.


twalpole:
@lucasmazza I think this is a little clearer/safer than my use of to_unsafe_h in the rails 5 updates

PR #3859, opened 2015-12-15
Explicitly permit rather than using to_unsafe_h

twalpole:
When implementing the rails 5 updates I used ActionController::Parameters#to_unsafe_h in some methods. This changes those instances to explicitly permit expected parameters instead.

### DIFF
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -253,8 +253,11 @@ module Devise
 
         # Find or initialize a record with group of attributes based on a list of required attributes.
         def find_or_initialize_with_errors(required_attributes, attributes, error=:invalid) #:nodoc:
-          attributes = attributes.to_unsafe_h.with_indifferent_access if attributes.respond_to? :to_unsafe_h
-          attributes = attributes.slice(*required_attributes).with_indifferent_access
+          attributes = if attributes.respond_to? :permit
+            attributes.slice(*required_attributes).permit!.to_h.with_indifferent_access
+          else
+            attributes.with_indifferent_access.slice(*required_attributes)
+          end
           attributes.delete_if { |key, value| value.blank? }
 
           if attributes.size == required_attributes.size
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -314,8 +314,8 @@ module Devise
 
         # Find a record for confirmation by unconfirmed email field
         def find_by_unconfirmed_email_with_errors(attributes = {})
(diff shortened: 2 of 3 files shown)


twalpole:
@lucasmazza I think this is a little clearer/safer than my use of to_unsafe_h in the rails 5 updates

lucasmazza:
> @lucasmazza I think this is a little clearer/safer than my use of to_unsafe_h in the rails 5 updates

sounds safer than the `to_unsafe_h` but I wonder if it might break any existing behaviour due the `slice` -> `permit` change as `permit` can raise an error on unpermitted values, I'll see if I can dig deeper and figure out what might go wrong here.

twalpole:
@lucasmazza Good point -- I've changed it over to call slice -> permit! on ActionParameters which should remove that issue

lib/devise/models/authenticatable.rb
```diff
@@ -253,8 +253,11 @@ def find_or_initialize_with_error_by(attribute, value, error=:invalid) #:nodoc:
 
         # Find or initialize a record with group of attributes based on a list of required attributes.
         def find_or_initialize_with_errors(required_attributes, attributes, error=:invalid) #:nodoc:
-          attributes = attributes.to_unsafe_h.with_indifferent_access if attributes.respond_to? :to_unsafe_h
-          attributes = attributes.slice(*required_attributes).with_indifferent_access
+          attributes = if attributes.respond_to? :permit
+            attributes.slice(*required_attributes).permit!.to_h.with_indifferent_access
```
sidonath: Hi @twalpole, curious why did you choose to test `:permit` before using `:permit!`? Unfortunately I have a case of monkey-patched `Hash` with the `#permit` method in an app, so I'm getting a `NoMethodError` here for `#permit!`. I guess it should be safer to test for `#permit!`?

ulissesalmeida: 👍 Make sense, if we are using `permit!`, we should check for `permit!`

sidonath: I'll make a PR

ulissesalmeida: Done in 90acc6f

ulissesalmeida: @sidonath Sorry, I didn't read you would send a PR. My bad

sidonath: Oh, well, this was far quicker 😄 thanks!

ulissesalmeida: 😆

PR #3866, opened 2015-12-18
Remove `KeyGenerator` implementation ported from Rails 4.x.

lucasmazza:
We can now use Active Support implementation directly and let it handle the caching dependency `thread_safe` was replaced by `concurrent` for Rails 5).


josevalim:
Yes! Please!

PR #3875, opened 2015-12-28
Update controller generator templates for new ParameterSanitizer syntax

james:
Hi, just playing around on a new project, and having a poke around HEAD on devise (mainly because I had read the README of HEAD and not 3.5.3, so stuck with what I'd learned about).

Not really too familiar with devise, so not sure if I'm missing something here, but think I may have noticed something you've missed, so here's a PR.

Devise::ParameterSanitizer has a new syntax for permitting additional attributes to a model. This commit updates the generated controllers to reflect that.


lucasmazza:
@james yeah, we missed those comments. Thanks a lot :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:.

PR #3882, opened 2016-01-02
Registrations#create yield before resource save

JensDebergh:
I've run into this issue where I needed to customise the `RegistrationsController` to fire some arbitrary service before saving the `resource` model. 

The problem here was that the resource is `yielded` after the resource gets saved, which in turn fired the model validations.

Example snippet:

```
# RegistrationsController
class RegistrationsController < Devise::RegistrationsController
  def create
     super do |resource|
        # Creates a workspace for a user and sets default_workspace_id
        UserRegistration.new(resource).call     
     end
  end
end

# User model
class User < ActiveRecord::Base
  # Model will not be able to save without the changes 
  # below and overriding the entire registrations#create method
  validates :default_workspace_id, presence: true
end
```

This change allows a developer to extend the registration process without overriding/(copy pasting) the entire `create` method definition.

Please correct me if I'm taking the wrong turn on this and an easier method is available. :tada:

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -14,8 +14,8 @@ class Devise::RegistrationsController < DeviseController
   def create
     build_resource(sign_up_params)
 
-    resource.save
     yield resource if block_given?
+    resource.save
     if resource.persisted?
       if resource.active_for_authentication?
         set_flash_message :notice, :signed_up if is_flashing_format?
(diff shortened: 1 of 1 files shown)


lucasmazza:
> Please correct me if I'm taking the wrong turn on this and an easier method is available.

Maybe overriding the `build_resource` method might work for you? We can move the `yield` call because  that would be quite a breaking changing for existing users of this feature.

PR #3899, opened 2016-01-15
stop using deprecated method

PashaLVWD:
Prior to rails 4.2
`60.ago` was equivalent to `60.seconds.ago`

`Numeric#ago` is gone after this commit
rails/rails@f1eddea

Now `60.ago` throws an exception
NoMethodError: undefined method `ago' for 60:Fixnum


josevalim:
The configuration in your app is likely wrong. For example, our default value does call `.minutes` accordingly: https://github.com/plataformatec/devise/blob/master/lib/devise.rb#L146

Thank you!

PR #3901, opened 2016-01-17
Update 'Lazy' strong parameters line in README

lukerollans:
Within README under the Strong Parameters section, the code example given tries to send `permit` from the context of `ApplicationController`. `permit` is a private method and `ApplicationController` can not call it.

The code block currently is as

```ruby
class ApplicationController < ActionController::Base
  before_action :configure_permitted_parameters, if: :devise_controller?

  protected

  def configure_permitted_parameters
    devise_parameter_sanitizer.permit(:sign_up, keys: [:username])
  end
end
```

It should be...

```ruby
class ApplicationController < ActionController::Base
  before_action :configure_permitted_parameters, if: :devise_controller?

  protected

  def configure_permitted_parameters
    devise_parameter_sanitizer.for(:sign_up) << :username
  end
end
```

There are further strong params customisation examples and I'm not sure if this issue also extends to them

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ class ApplicationController < ActionController::Base
   protected
 
   def configure_permitted_parameters
-    devise_parameter_sanitizer.permit(:sign_up, keys: [:username])
+    devise_parameter_sanitizer.for(:sign_up) << :username
   end
 end
 ```


lucasmazza:
@lukerollans thanks for the PR but the README is displaying the new API for the unreleased Devise 4, that is incompatible with previous versions. I've added a notice about the API change to the README on 0a4da78.

lukerollans:
Ah, great, thanks for the clarification!

bDrwx:
OMG

PR #3907, opened 2016-01-21
Make Devise check for Rails >= 4

Quintasan:
Commit `eba91e6` made Devise read `secret_key_base` if it was set but
the method checking for Rails version used `==` making it fail the check
when running Rails 5.

### DIFF
--- a/lib/generators/devise/install_generator.rb
+++ b/lib/generators/devise/install_generator.rb
@@ -22,7 +22,7 @@ module Devise
       end
 
       def rails_4?
-        Rails::VERSION::MAJOR == 4
+        Rails::VERSION::MAJOR >= 4
       end
     end
   end


lucasmazza:
If I'm not mistaken this `rails_4?` method is no longer used on master as we removed support for Rails 3 on https://github.com/plataformatec/devise/commit/2024fca4dfa3323070c3477e262b8422cadf6a42#diff-29.

Quintasan:
@lucasmazza if that's the case then there is something else entirely wrong. I'm running Rails and Devise from master and I tried starting a Rails console for my production server. I went ahead and did:

```
RAILS_ENV=production SECRET_KEY_BASE=insetsomethinghere ASRAPI_DATABASE_USER=user ASRAPI_DATABASE_PASSWORD=password rails console
```
which failed with
```
asrapi@server:~/asrapi$ /home/asrapi/.rvm/gems/ruby-2.3.0@asrapi/bundler/gems/devise-cc02d06c5823/lib/devise/rails/routes.rb:500:in `raise_no_secret_key': Devise.secret_key was not set. Please add the following to your Devise initializer:

  config.secret_key = '046da7994ccb8c29d2ccdb3c55dc4340b3af5ec32b22170a3d3300bc1a50151716100bea2778046f5fb4138eb7093aec4980c9312e1b1334131c15ed0e3ec4e9'

Please ensure you restarted your application after installing Devise or setting the key.
```

lucasmazza:
Can you provide a fresh app that replicates your issue?

Quintasan:
@lucasmazza I was unsuccessful despite two attempts. It appears I ran into something strange, sorry for the commotion.

PR #3917, opened 2016-01-26
Try parsing time in Rememberable before giving up

boblail:
I encountered the exception that 048d05a55358a53bbc6d4ea9c78c206f309fcd74 fixes; but the value of `generated_at` (now being discarded) was a serialized time: `"2016-01-26 17:13:52 UTC"`.

After following these steps:
    1. Sign in with **Remember me** checked
    2. Delete the session key (but not `remember_user_token`) from the cookies
    3. Refresh

Using devise `v3.5.5`, my application would crash.

Using 048d05a55358a53bbc6d4ea9c78c206f309fcd74, you'd be signed out.

Using this pull request, you'd be signed in.

### DIFF
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -130,6 +130,7 @@ module Devise
 
         def serialize_from_cookie_with_or_without_record(record, args)
           id, token, generated_at = args
+          generated_at = Time.parse(generated_at) rescue nil if generated_at.is_a?(String)
 
           # The token is only valid if:
           # 1. we have a date


josevalim:
I would rather try to figure out why it is becoming a string in the first place and avoid it. For example, we could try to always make it a string or always an integer to avoid handling all possible scenarios.

josevalim:
Are you using any special mechanism for cookie serialization?

lucasmazza:
@josevalim afaik the JSON cookie serialization doesn't deserialize timestamps as `Time` objects.

lucasmazza:
Similar fixes we had on `timeoutable` when rails migrated to JSON over Marshal for serialization: https://github.com/plataformatec/devise/commit/eb9db7ba3a552ed410b20f1040e167569838b068 and https://github.com/plataformatec/devise/commit/eb9db7ba3a552ed410b20f1040e167569838b068. 

The sane way back then to support for serializers was to deal with integers, and we can do the same here.

josevalim:
I guess we have no option then. Let's handle this in the function you have just extracted. We will need to backport it though. :S

lucasmazza:
@boblail can you move your path to the `remember_me?` method?

lucasmazza:
Closed by #3927.

PR #3918, opened 2016-01-27
Refactor `Rememberable#serialized_in_cookie?` to split class/instance API

lucasmazza:
An alternative API to deal with the `serialize_from_cookie_with_or_without_record` pile of responsibilities. /cc @josevalim

### DIFF
--- a/lib/devise/controllers/rememberable.rb
+++ b/lib/devise/controllers/rememberable.rb
@@ -12,8 +12,8 @@ module Devise
       def remember_me_is_active?(resource)
         return false unless resource.respond_to?(:remember_me)
         scope = Devise::Mapping.find_scope!(resource)
-        cookie = cookies.signed[remember_key(resource, scope)]
-        resource.class.serialized_in_cookie?(resource, *cookie)
+        _, token, generated_at = cookies.signed[remember_key(resource, scope)]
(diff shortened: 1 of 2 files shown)


josevalim:
I don't like that we are a using a serialized_in_cookie? in the instance. Can we find a better name? Maybe `remember_active?` or something of sorts. Also, I wouldn't pass the ID in. So I would go with something like `remember_active?(token, generated_at)`. Thoughts?

lucasmazza:
> Maybe remember_active? or something of sorts.

:+1:

> So I would go with something like remember_active?(token, generated_at). Thoughts?

We could do that, and it would help to move the cookie Time serialization handling outside of the model. The only downside here is that the "cookie splatting" logic will move away from the method that generates the cookie (`serialize_into_cookie`), but maybe that isn't so bad as one might consider.

josevalim:
@lucasmazza  the cookie splatting logic is only needed for backwards compatibility. so it is ok to be a bit ugly for now, we can leave a todo and remove it in future versions.

lucasmazza:
@josevalim pushed a new commit, how does it look?

josevalim:
Beautiful  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #3920, opened 2016-01-28
Change #password_required? implementation

graudeejs:
This change will allow updating user with update_attributes containing
blank password and password_confirmation fields (common case for profile
updates).

What do you think?

### DIFF
--- a/test/models/validatable_test.rb
+++ b/test/models/validatable_test.rb
@@ -64,7 +64,7 @@ class ValidatableTest < ActiveSupport::TestCase
     user = create_user
 
     user.password = ''
-    user.password_confirmation = ''
+    user.password_confirmation = 'x'
 
     assert user.invalid?
     assert_equal 'can\'t be blank', user.errors[:password].join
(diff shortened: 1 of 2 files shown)


test/models/validatable_test.rb
```diff
@@ -64,7 +64,7 @@ class ValidatableTest < ActiveSupport::TestCase
     user = create_user
 
     user.password = ''
-    user.password_confirmation = ''
+    user.password_confirmation = 'x'
```
josevalim: You are changing an existing tests here.

josevalim:
For the profile page, you should use update_with_password and update_without_password accordingly. Other than that, this seems to be a breaking change. Thank you anyway!

test/models/validatable_test.rb
graudeejs: Yes, that's because blank string doesn't trigger password update.
And it doesn't trigger validation any longer.
So one of fields had to have ```present? == true``` to trigger validations.

graudeejs:
That doesn't work, when you need to update user as nested object.
Also in that case you need to check weather password was supplied.

My implementation makes this whole lot easier.

test/models/validatable_test.rb
josevalim: You can always customize password_required? in your app as well as it seems you have a very specific scenario. :)

graudeejs: That's what I do. Just wanted to offer this fix to save many people trouble doing that.
http://stackoverflow.com/questions/5113248/devise-update-user-without-password/35064640#35064640

PR #3921, opened 2016-01-28
Added DeviseController#set_flash_message!

oss92:
* Introduced `DeviseController.set_flash_message!` for conditional flash messages setting to reduce method complexity.

* Reduced complexity of `DeviseHelper#devise_error_messages!` and `Devise::FailureApp#scope_url`.

* `DeviseHelper#devise_error_messages_html` not only reduces complexity but also makes error messages easier to override by developers who want to change styling of error message.

* Changed `confirm!` to `confirm` in tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Unreleased
+
+* enhancements
+  * Introduced `DeviseController#set_flash_message!` for conditional flash messages setting to reduce complexity.
+
 ### 4.0.0.rc1 - 2016-01-02
 
 * Support added to Rails 5 (by @twalpole).
--- a/app/controllers/devise/confirmations_controller.rb
+++ b/app/controllers/devise/confirmations_controller.rb
@@ -22,7 +22,7 @@ class Devise::ConfirmationsController < DeviseController
     yield resource if block_given?
 
     if resource.errors.empty?
-      set_flash_message(:notice, :confirmed) if is_flashing_format?
+      set_flash_message!(:notice, :confirmed)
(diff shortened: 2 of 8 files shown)


lucasmazza:
I dig the `set_flash_message!` and internal reorganisation of `scope_url` but I'm not too keen on the addition of devise_error_messages_html` and `generate_url`. They might push some of the complexity outside of their methods to new ones but they might new not-exactly-public APIs that we probably don't want to support and let developers to override it.

@josevalim @carlosantoniodasilva any thoughts on this?

josevalim:
I agree with the changes but we need to #:nodoc: the new methods exactly so users do not consider them to be part of our API.

oss92:
Added #:nodoc: to new methods.

carlosantoniodasilva:
My 2 cents: I kinda agree with @lucasmazza on this one, the flash message extraction and the internal refactorings inside the methods in general look good, but I'd rather not introduce extra methods for the sake of reducing complexity. It might make code climate happier, but it might also feel that reading the code gets harder.

Thank you for your contribution @oss92! :heart:

oss92:
Ok so should I remove `generate_url` and `devise_error_messages_html` to make this PR ready?

And you are welcomed :+1: I pretty much use devise in ALL of my Rails applications and I should start contributing to it :100:

lucasmazza:
> Ok so should I remove generate_url and devise_error_messages_html to make this PR ready?

Yes, please - and rebase your branch against `master` as it show that there are some conflicts to be solved. Thanks!

oss92:
Done @lucasmazza

PR #3926, opened 2016-01-29
[close 3925] Fix version for older Bundler compat

schneems:


lucasmazza:
@schneems thanks :heart:

PR #3927, opened 2016-01-29
Attempt to coerce the `generated_at` cookie to a `Time` object.

lucasmazza:
Time objects aren't properly coerced back when using the JSON cookie serialization,
so we need to do it ourselves.

To avoid any new JSON serialization issues, we now store the `generated_at` as
an String with the timestamp seconds + miliseconds in the cookie but still the
previous JSON encoded format.

Thanks to @boblail at https://github.com/plataformatec/devise/pull/3917 for the
initial patch.

### DIFF
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -97,6 +97,12 @@ module Devise
       end
 
       def remember_me?(token, generated_at)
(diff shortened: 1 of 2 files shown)


josevalim:

boblail:
:+1: Thanks, @lucasmazza!

PR #3933, opened 2016-02-03
Update passwords_controller.rb

bogdanvlviv:
fix bug: undefined method `user_url' for #<Devise::PasswordsController


bogdanvlviv:
i have this bug when user exist, successfully_sent?(resource) => false. on my own project i need to override `successfully_sent?` method.

josevalim:
Thank you. Could we please have a test case?

bogdanvlviv:
when in Devise::PasswordsController:
`respond_with(resource)`

it is `false`: 
`assert_nothing_raised(NoMethodError) { post :create, params: { "user" => { "email" => @user.email } } }`

bogdanvlviv:
when exist "Soft delete a user" https://github.com/plataformatec/devise/wiki/How-to:-Soft-delete-a-user-when-user-deletes-account , we can successfully send reset password those deleted users, i think this is bug. i think we need show message (from methos  `inactive_message`). Of course each developer can do this logic but it is not good. each developer need override three methods for "soft delete a user", and methods `Devise::PasswordsController#create`, `Devise::PasswordsController#successfully_sent?`, `Recoverable#send_reset_password_instructions`. this i not good. i think each developer should create only  "soft delete a user" and Devise everything else.Devise is need be more flexible here(in this situation). Because it is Devise :)

I can do it, but....
@josevalim, what do you think about it?

and this pull request fix bug when user is saved bun can not receive reset password (because of our logic, for example he is ''soft deleted' and we do not want send reset password to him)

bogdanvlviv:
You can give me questions. I am ready to answer.

ulissesalmeida:
@josevalim @lucasmazza When we are planning to release it?

lucasmazza:
@ulissesalmeida I think we can push this on 4.2

@bogdanvlviv can you please rebase your branch with the recent changes on `master`? Thanks.

test/controllers/passwords_controller_test.rb
```diff
@@ -29,4 +29,24 @@ def put_update_with_params
...
+  end
+
+  test 'redirect to http://test.host/users/password/new after unsuccessfully sent reset password instructions' do
+    Devise::PasswordsController.any_instance.stubs(:successfully_sent?).with(@user).returns(false)
+
+    post :create, params: { "user" => { "email" => @user.email } }
+    assert_redirected_to 'http://test.host/users/password/new'
+  end
+
+  test 'nothing raised after unsuccessfully sent reset password instructions for users that exist' do
```
lucasmazza: This test case exercises the same code as the previous test, and `assert_nothing_raised` is not a very helpful assertion to have around since we don't expect the previous test to raise anything (and I'm planning on removing the other occurrences of this assertion from our test suite).

Can you please remove this test case?

test/controllers/passwords_controller_test.rb
```diff
@@ -29,4 +29,24 @@ def put_update_with_params
...
+
+  test 'redirect to http://test.host/users/sign_in after successfully sent reset password instructions' do
+    Devise::PasswordsController.any_instance.stubs(:successfully_sent?).with(@user).returns(true)
+
+    post :create, params: { "user" => { "email" => @user.email } }
+    assert_redirected_to 'http://test.host/users/sign_in'
+  end
+
+  test 'redirect to http://test.host/users/password/new after unsuccessfully sent reset password instructions' do
+    Devise::PasswordsController.any_instance.stubs(:successfully_sent?).with(@user).returns(false)
```
lucasmazza: There is any change we can set up this test scenario without stubbing the `successfully_sent?` method? We are stubbing an internal method of our controller, and with the stubbing we are not testing it's implementation (which is a core piece of the changes from this Pull Request).

test/integration/recoverable_test.rb
```diff
@@ -209,43 +225,30 @@ def reset_password(options={}, &block)
...
+    request_forgot_password
+    assert_contain 'Your account is not activated yet.'
   end
 
-  test 'unlocks and signs in locked user automatically after changing it\'s password if unlock strategy is :email' do
-    swap Devise, unlock_strategy: :email do
-      user = create_user(locked: true)
-      request_forgot_password
-      reset_password
+  test 'does not send the reset password instructions to not exist user' do
```
lucasmazza: `not exist user` => `non existing user` ?

test/integration/recoverable_test.rb
```diff
@@ -209,43 +225,30 @@ def reset_password(options={}, &block)
...
+    request_forgot_password
+    assert_contain 'Email not found'
   end
 
-  test 'unlocks and signs in locked user automatically after changing it\'s password if unlock strategy is :both' do
-    swap Devise, unlock_strategy: :both do
-      user = create_user(locked: true)
-      request_forgot_password
-      reset_password
+  test 'does not send the reset password instructions to not confirm user' do
```
lucasmazza: `not confirm user` => `unconfirmed user` ?

app/controllers/devise/passwords_controller.rb
```diff
@@ -77,4 +82,27 @@ def unlockable?(resource)
     def translation_scope
       'devise.passwords'
     end
+
+    # Helper for use after calling send_*_instructions methods on a resource.
+    # If we are in paranoid mode, we always act as
+    # if the resource was valid and active for authentication
+    # and instructions were sent.
+    def successfully_sent?(resource)
```
ulissesalmeida: Reading the name of the method I expect just a `true` or `false` return.
But this method also set a flash message. Can we move flash manipulation from this method and to another?

bogdanvlviv: i think will be correctly simple rename this method?

test/controllers/passwords_controller_test.rb
bogdanvlviv: You are right. i should stubs here method `active_for_authentication?`

lucasmazza: :+1:

app/controllers/devise/passwords_controller.rb
ulissesalmeida: @bogdanvlviv I'm sorry. I've looked the `DeviseController` and the `successfully_sent?` have this same behavior of check and set flash messages. I think we can keep this way and do this refactor in another moment. :)

bogdanvlviv: :+1:

lucasmazza:
@bogdanvlviv can you please remove the ident style changes from this Pull Request? We don't strictly follow the rails code style and these changes are noise to the proposed fix :)

bogdanvlviv:
@lucas-nelson, done!

bogdanvlviv:
stale

PR #3946, opened 2016-02-10
Removes Hakiri, since it thinks Devise is a Rails app

georgeguimaraes:
I'd like to remove the Hakiri badge from devise's README.

Since we do check in Gemfile and Gemfile.lock for development
and testing purposes (contrary to the popular belief that gems
don't need those files), Hakiri thinks Devise is a Rails app,
instead of a library/gem.

Depending on the Rails version that is locked in Gemfile.lock,
Hakiri will show several CVEs for Rails. But our Travis pipeline
builds against several Rails' branches, all of them updated.

And it is up to the developers using Devise to update
Rails in their apps.

Those warnings (Rails CVEs) don't make sense for devise.


josevalim:

Issue #3800, opened 2015-10-27
Check for active_record / ORM support during devise:install


streetlogics:
I have a rails app where I wasn't using active_record to start.  I installed devise, ran `rails g devise:install` and then `rails g devise:views`.  However, after running the command, I got an error saying "'require': cannot load such file -- devise/orm/false".  I found this issue - https://github.com/plataformatec/devise/issues/2764, which got me started in the right direction.  I got all my requires squared away, but was still getting the error.  I ended up finding the line in the autogenerated `config/intializers/devise.rb` had been set to `require 'devise/orm/false'`, which needed to be `require 'devise/orm/active_record'`.  

I know this was ultimately sort of my fault for not having active_record required up front, however, with that said, your package also obviously has a "false" value that it's working with to generate that file, so a check should be added to the generate method to make sure active_record or another ORM is loaded, and if not, throw an error during the install script, as otherwise, a developer might (and rightly assume) that a successful install script == a successful install.  :)  Thanks!

georgeguimaraes:
if you remove the line `require 'devise/orm/false' The wiki is maintained by the community. So if there aren't any up to date instructions, we recommend you to explore the solution yourself and hopefully contribute your findings back!

georgeguimaraes:
If you remove the line `require 'devise/orm/false'`, does the generator `rails g devise:views` run successfully?

georgeguimaraes:
(sorry for the mistake on the first reply. I already deleted it)

streetlogics:
This is what the progression looks like if I comment out that line:

```
$ rails g devise:views
      invoke  Devise::Generators::SharedViewsGenerator
      create    app/views/devise/shared
      create    app/views/devise/shared/_links.html.erb
      invoke  form_for
      create    app/views/devise/confirmations
      create    app/views/devise/confirmations/new.html.erb
      create    app/views/devise/passwords
      create    app/views/devise/passwords/edit.html.erb
      create    app/views/devise/passwords/new.html.erb
      create    app/views/devise/registrations
      create    app/views/devise/registrations/edit.html.erb
      create    app/views/devise/registrations/new.html.erb
      create    app/views/devise/sessions
      create    app/views/devise/sessions/new.html.erb
      create    app/views/devise/unlocks
      create    app/views/devise/unlocks/new.html.erb
      invoke  erb
      create    app/views/devise/mailer
      create    app/views/devise/mailer/confirmation_instructions.html.erb
[...]

PR #3954, opened 2016-02-14
Raise error if no ORM is found during install generator

arjunsharma:
This is my stab at fixing #3800. I'm happy to keep working on this if you have feedback, I've used Devise on many projects and figured I would try and do my part to contribute.

Thanks!

Arjun

### DIFF
--- a/lib/generators/devise/install_generator.rb
+++ b/lib/generators/devise/install_generator.rb
@@ -10,6 +10,7 @@ module Devise
       class_option :orm
 
       def copy_initializer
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @arjunsharma! I think we should also check if `ActiveRecord` is defined or not. If it is defined, then we should not raise. :)

josevalim:
Ignore me. The patch seems fine as is. :) Thank you!

arjunsharma:
Awesome!

josevalim:

PR #3957, opened 2016-02-15
Use after_sign_up_path_for in case of resource.persisted? is false

olegantonyan:
This should be self-explanatory.
If I define `after_sign_up_path_for` or `after_update_path_for` I expect these to be used in both cases: success or failure

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -29,7 +29,7 @@ class Devise::RegistrationsController < DeviseController
     else
       clean_up_passwords resource
       set_minimum_password_length
-      respond_with resource
+      respond_with resource, location: after_sign_up_path_for(resource)
     end
   end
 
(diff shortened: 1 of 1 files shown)


lucasmazza:
@olegantonyan thanks for the patch but I don't think that's the case want here - in case of failure we want the controller to render back the `registrations/new` to display any errors and let the end user try again, rather than redirecting it to the same path as success scenarios. 

Do you have any extra rationale behind this change?

olegantonyan:
The problem is that default path might not exists. For example, when I override `build_resource` and use an object which has no polymorphic url:
```ruby
class Users::RegistrationsController < Devise::RegistrationsController
  def build_resource(hash_params = {})
    self.resource = User::MerchantOwner.new(hash_params)
  end
...
class User::MerchantOwner < User
...
end
```
Here `User::MerchantOwner` has no polymorphic path and this is fine, I don't want to pollute it with crazy routing logic. But devise will try to use this non-existent polymorphic path because of `respond_with resource`.

lucasmazza:
@olegantonyan in that case I recommend overriding the action or breaking the polymorphic controller apart, as this is an application specific scenario.

PR #3965, opened 2016-02-22
Fix request_values to correctly access the request object 

sblackstone:
I was working from the directions located here: 
https://github.com/plataformatec/devise/wiki/How-to:-Scope-login-to-subdomain

and nothing was working until I took a look at how authenticatable was accessing values from the request hash using `request.send(:key)` instead of `self.request[k]` 

At least in Rails 4.2, I have not been able to demonstrate `send` ever working to access an attribute from the `request` object.

The tests that now fail are somewhat suspect to me - but I'm hoping someone with more experience with this code base will be able to confirm - it seems like an invalid request key is expecting NoMethodError - which seems to occur regardless of whether or not the attribute exists or not. 

But the documentation from the above link points out that an error will occur if result_keys are passed by array - but not by hash assuming the values are "false" for "not required"

This seems to be implemented in `parse_authentication_key_values` but false is returned rather than throwing an error.  

It doesn't seem like its possible to use this functionality since request_values comes before parse_authentication_key_values and will always throw an error regardless of how the value_keys are passed (by hash will false values or by array)

### DIFF
--- a/lib/devise/strategies/authenticatable.rb
+++ b/lib/devise/strategies/authenticatable.rb
@@ -148,7 +148,7 @@ module Devise
 
       def request_values
         keys = request_keys.respond_to?(:keys) ? request_keys.keys : request_keys
-        values = keys.map { |k| self.request.send(k) }
+        values = keys.map { |k| self.request[k] }
         Hash[keys.zip(values)]
       end
 


sblackstone:
Nevermind!

javs-perez:
@sblackstone I'm running into this same error with devise. What did you do to resolve it?

sblackstone:
@javs-perez Use authorization_keys instead of request_values

javs-perez:
@sblackstone you lost me a bit there, request_values is been used by devise, how can I have it used the authorization keys instead?

sblackstone:
My issue was that, by the name, I was confused that request_values didn't include the values from the request - what its meant for is literally getting values out of the request object that are put there by rails.

If you do devise :database_authenticatable, authorization_keys: [:a, :b]  then it will look for :a and :b to be passed in on your user object for the form and make them available to you in your authentication method on your user class.

javs-perez:
Thanks for taking the time to explain that. I was thinking the same way you were also, I was trying to pass 'client_id' as one of the request_values but finally figured out that the request_key are put there by rails and include things like :subdomain, :domain, :ip, etc.

I'll take a look at the authorization_keys and see if I can pass 'client_id' there instead. Thanks again, you were a lot of help.

PR #3968, opened 2016-02-24
fix test helpers for relative_url_root

evanboho:
BUG:
* in Rails 4.2, Devise goes into an infinite loop if relative_url_root is not set in the Rails config
* when relative_url_root is set, Devise's test_helpers do not pass the attempted_url in the warden options
* this causes an error at `Pathname.new(attempted_path)` in the failure_app

FIX:
* add attempted_path to the warden options in the test_helpers

Versions: 
* Devise 3.5.6
* Rails 4.1.13
* Warden 1.2.6

### DIFF
--- a/lib/devise/test_helpers.rb
+++ b/lib/devise/test_helpers.rb
@@ -93,6 +93,7 @@ module Devise
 
     def _process_unauthenticated(env, options = {})
       options[:action] ||= :unauthenticated
+      options[:attempted_path] ||= ::Rack::Request.new(env).fullpath
       proxy = env['warden']
       result = options[:result] || proxy.result
 


lucasmazza:
@evanboho thanks a lot for the fix - any change we can write a test case for this?

ulissesalmeida:
@evanboho Thank you for your report. Could please give more information in how to reproduce this bug?

In your description, you have mentioned about an infinite loop on Rails 4.2. But in the end of description you've mentioned the versions you tested was Devise 3.5.6 and Rails 4.1.13.

I've downloaded your branch, but your branch is running Devise `3.0.0.rc` and Rails `4.0.0.rc1`. I've uncommented the code you added and test suite is passing.

Could you please give us more information in how to reproduce it?

ulissesalmeida:
@evanboho I'm closing this PR. Feel free to reopen when you have more information in how to reproduce this issue, like a test case or a sample application for example.

Issue #3338, opened 2014-11-23
Reset password link wrong resource link


killerham:
Using devise 3.4.1 

I have both an api namespace using the ActionController::API as a base class and a regular rails app using the ActionController::Base.

My routes are arranged as follows:

``` ruby
Rails.application.routes.draw do

  namespace :api, defaults: { format: "json" } do
    namespace :v1 do
      devise_for :users
    end
  end

  devise_for :users
  resources :users
end
```

When going to users/password/new and enter an email, I get the email. However, the link in the email has /api/v1/users/password/edit?reset_password_token as the prefix. Why is this using the api/v1 namespace?

The second issue I'm having is that I want to post /api/v1/users/password/ to send the password reset link, which works but again the link has /api/v1/users/password/edit?reset_password_token as the prefix (which makes sense). How can I remove the /api/v1 prefix in the reset link?

markedmondson:
@rubyrider I ran into this today, instead of forcing customization of the mailer partials for each namespace, the user scope could be collected from the path in the controller and pushed through to the mailer. 

The problem I was having was around this sort of configuration:

```
namespace :api
    devise_for :users, only: []
end

namespace :admin
    devise_for :users
end
```

This mapped into Devise::Mappings like so...

```
{
   :api_user => ...,
   :admin_user => ...
}
```

In **admin**/passwords_controller (or confirmations/unlocks) the matched User would be loaded as the object, this is passed all the way down to the mailer which calls `initialize_from_record` on the User object and hits the first matching result of `Devise::Mapping` with the same model name, finding :api_user. Once in the view, this is used to load the `edit_password_url(@resource)` where @resource was a :api_user, thus failing as there would be no route defined with this scope.

I've not had change to write any tests around it but this is the idea. Worth pursuing?

https://github.com/markedmondson/devise/tree/mailer-scope

PR #3973, opened 2016-02-25
Add the option to pass the current user scope down from the controller to mailers

markedmondson:
Born from https://github.com/plataformatec/devise/issues/3338


josevalim:
Thank you but this is a lot of work and breaking changes to justify the benefits that actually affect only a small fraction of Devise users. It also feels conceptually incorrect to be passing this information down to the models. It is simpler to just customize your mailer layouts. Thanks :)

avit:
> It is simpler to just customize your mailer layouts

How should the mailer layouts know which scope to link to?

PR #3985, opened 2016-03-07
Do not use the dynamic `:action` segment on Omniauth routes.

### LINKED ISSUE #3983: rails 5 with devise_omniauth_callback
im getting this error,

W, [2016-03-07T01:08:19.779204 #12853]  WARN -- : DEPRECATION WARNING: Using a dynamic :action segment in a route is deprecated and will be removed in Rails 5.1. (called from devise_omniauth_callback at /usr/local/rvm/gems/ruby-2.2.4/bundler/gems/devise-342304375946/lib/devise/rails/routes.rb:452)

i tried comment off my omniauth, it still persists with this error?
what is the new way to implement omniauth with devise?

thank you

lucasmazza:
This was deprecated on rails/rails#23980.

We now generate scope and provider specific routes, like `user_facebook_omniauth_callback`
or `user_github_omniauth_callback`.

We could deprecate the `omniauth_authorize_path` in favor of the generated routes, but
the `shared/links.html.erb` depends on it to generate all omniauth links at once.

Closes #3983.

### DIFF
--- a/lib/devise/omniauth/url_helpers.rb
+++ b/lib/devise/omniauth/url_helpers.rb
@@ -4,14 +4,14 @@ module Devise
       def self.define_helpers(mapping)
       end
 
-      def omniauth_authorize_path(resource_or_scope, *args)
+      def omniauth_authorize_path(resource_or_scope, provider, *args)
         scope = Devise::Mapping.find_scope!(resource_or_scope)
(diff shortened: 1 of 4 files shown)


josevalim:
:shipit:

PR #4004, opened 2016-03-23
removed deprecated private method Devise::Controllers::Helpers#expire_session_data_after_sign_in!

bogdanvlviv:


ulissesalmeida:
Thanks! @josevalim Can we merge it right now?

PR #4014, opened 2016-03-30
Humanize authentication keys in failure flash message

byzg:
If developer wants to use other authentication keys instead of email and writes in him model
```devise :database_authenticatable, authentication_keys: [:username]authentication_keys: [:user_name]```
and writes in him locale 
```
devise:
  failure:
    invalid: '%{authentication_keys} o contraseña no válidos.'
```
and after that user types wrong user_name or password, flash get message **"user_name  o contraseña no válidos."** instead of **"User name  o contraseña no válidos."**. Changes of this commit fix it. 
PS: sorry for my english

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -103,7 +103,7 @@ module Devise
         options[:scope] = "devise.failure"
         options[:default] = [message]
         auth_keys = scope_class.authentication_keys
-        keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map(&:to_s).map(&:humanize)
+        keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class.human_attribute_name(key) }
         options[:authentication_keys] = keys.join(I18n.translate(:"support.array.words_connector"))
         options = i18n_options(options)
 
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -103,7 +103,7 @@ module Devise
         options[:scope] = "devise.failure"
         options[:default] = [message]
         auth_keys = scope_class.authentication_keys
-        keys = auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys
+        keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map(&:to_s).map(&:humanize)
         options[:authentication_keys] = keys.join(I18n.translate(:"support.array.words_connector"))
         options = i18n_options(options)
 
--- a/test/failure_app_test.rb
+++ b/test/failure_app_test.rb
@@ -133,14 +133,14 @@ class FailureTest < ActiveSupport::TestCase
 
     test 'uses the proxy failure message as symbol' do
       call_failure('warden' => OpenStruct.new(message: :invalid))
-      assert_equal 'Invalid email or password.', @request.flash[:alert]
+      assert_equal 'Invalid Email or password.', @request.flash[:alert]
       assert_equal 'http://test.host/users/sign_in', @response.second["Location"]
     end
 
     test 'supports authentication_keys as a Hash for the flash message' do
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by ulissesalmeida** [comment]
@byzg Thank you for your contribution. I think when you writing your custom translation you can put your desired translation directly in your YAML. Using your example:

```
devise:
  failure:
    invalid: User name o contraseña no válidos.
```

**Comment by byzg** [comment]
@ulissesalmeida thank you for your note, but the one hand when you want use other authentication keys and default Devise locale [en.yml](https://github.com/plataformatec/devise/blob/master/config/locales/en.yml) you have to overwrite
```
failure:
      invalid: "Invalid %{authentication_keys} or password."
```
to
```
failure:
      invalid: "Invalid user name or password."
```


  On the other hand it appears that Devise gives you `%{authentication_keys}` in [failure_app.rb](https://github.com/byzg/devise/blob/master/lib/devise/failure_app.rb)
```
options[:authentication_keys] = keys.join(I18n.translate(:"support.array.words_connector"))
options = i18n_options(options)
I18n.t(:"#{scope}.#{message}", options)
```
for using in custom translations, but you can't use `%{authentication_keys}` if authentication_key consists of two or more words.

**Comment by ulissesalmeida** [comment]
@byzg You're right. Since Devise provides the `authentication_keys` to use in the translation files, would be nice if it has a good format for humans.

But I think maybe, we should not just humanize, but instead translate it for users.

Something like:

```ruby
keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class. human_attribute_name(key) }
```

The `scope_class` has the current active record class, for example `User`.
The method `human_attribute_name` translates the attribute based on i18n configured for the model. It will translate the User Name to "Nome de usuário" in PT-BR for example.

What do you think? Do you wanna try it?

byzg:
@ulissesalmeida exactly! This solution much better! I fixed chenges by your note.

ulissesalmeida:
@byzg Thanks! Looks good!

connorshea:
I just noticed this change after upgrading to 4.1.0. With an incorrect password the user sees "Invalid Login or password." "Login" shouldn't be capitalized in this case, as it's not a proper noun.

Would this be considered an issue in Devise itself, or something we should fix in our specific application? I don't speak Portuguese, so I don't know the grammar of the language and don't really understand the change entirely.

ulissesalmeida:
@connorshea I think you're right. With this change, the attribute is translated based on your i18n model configuration. Maybe we should call downcase on these strings if it's a common pattern store the translations capitalized.

connorshea:
I'm not actually sure where the "Login" string is coming from in this case.

The application this is occuring in is GitLab, specific string here: https://github.com/gitlabhq/gitlabhq/blob/master/config/locales/devise.en.yml#L12

Note that the master branch isn't on Devise 4, so you won't see the issue if you run it.

connorshea:
@ulissesalmeida if you downcase the keys themselves but they're at the beginning of the sentence, will it still become capitalized? If not, is there a way we can implement it that'd cause that?

ulissesalmeida:
@connorshea If we downcase the keys and change the message to move the keys to the beginning of the sentence, like this:
```
invalid: "%{authentication_keys} or password."
```
The first word will not become capitalised. We may want to add some code to check it and capitalize correctly, but I don't think it worth. Because some users may want to add a message without capitalizing the first word, if we do some "automatic" change it will end the possibility of message customization.

I think we should fix the issue to works with the `default` message.

byzg:
I just note that just downcase or capitalise acording to the beginning of the sentence is may be bad idea cause we can't talk about all languages. For example, in German, all nouns begin with a capital letter, regardless of the position in a sentence:
`Ungültige Login oder Passwort`

connorshea:
Should we move this discussion over to the issue I opened?

fny:
Just a thought. Why not change the translations to guarantee the authentication keys come first?

For example `invalid: "%{authentication_keys} or password is invalid"`

PR #4030, opened 2016-04-08
Added device type to trackable 

Jrsnow8921:
I love using devise, I think this will help a lot of people with their logic for mobile and desktop styling if it's built right into devise that way they can use current_user methods for device types making cross platform styling simple. Any question or concerns please let me know!


lucasmazza:
@Jrsnow8921 thanks for the pull request but a change like this isn't backwards compatible and it's an application specific feature rather than something worth having on Devise by default.

Jrsnow8921:
Okay no problem thanks though! @lucasmazza

PR #4031, opened 2016-04-09
Update README

tomkadwill:


README.md
```diff
@@ -109,7 +109,7 @@ The generator will install an initializer which describes ALL of Devise's config
 rails generate devise MODEL
 ```
 
-Replace MODEL with the class name used for the application’s users (it’s frequently `User` but could also be `Admin`). This will create a model (if one does not exist) and configure it with default Devise modules. The generator also configures your `config/routes.rb` file to point to the Devise controller.
+Replace MODEL with the class name used for the application’s users (it’s frequently `User` but could also be `Admin`). This will create a model (if one does not exist) and configure it with default the Devise modules. The generator also configures your `config/routes.rb` file to point to the Devise controller.
```
ulissesalmeida: Should not be `with the default Devise modules`?

tomkadwill: Thanks @ulissesalmeida, my mistake.

ulissesalmeida:
@tomkadwill Thanks!

PR #4033, opened 2016-04-09
Note change to Rails 5 that affects Devise example

codeodor:
[A change in Rails 5](https://github.com/rails/rails/commit/39794037817703575c35a75f1961b01b83791191) left me hunting for hours on why I could no longer log in to my application. To save others the trouble, I thought it would be nice to note it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ Devise will create some helpers to use inside your controllers and views. To set
 before_action :authenticate_user!
 ```
 
+For Rails 5, note that `protect_from_forgery` is no longer prepended to the `before_action` chain, so if you have set `authenticate_user` before `protect_from_forgery`, your request will result in "Can't verify CSRF token authenticity." To resolve this, either change the order in which you call them, or use `protect_from_forgery prepend: true`.
+
 If your devise model is something other than User, replace "_user" with "_yourmodel". The same logic applies to the instructions below.
 
 To verify if a user is signed in, use the following helper:


lucasmazza:
@codeodor thanks!

arthurnn:
Note to help someone else in the future:
I upgraded an app to rails 5, that was using devise and peek.
In order to enable the peek bar, I had this in my application_controller:
```ruby
def peek_enabled?
  defined?(:current_admin) && nil != current_admin
end
```
However, peek will prepend an action, that will call that method, see: https://github.com/peek/peek/blob/675c21ba11ea5d81425ba065f8768bc56f6113e2/lib/peek/controller_helpers.rb#L6

Therefore, even when prepending the `protect_from_forgery` call, it will still not work, because the peek prepend will happen first. While on Rails 4, the `protect_from_forgery` prepend happened first, as it was inside Rails.

arthurnn:
btw, Does anyone knows why calling `current_*` before the CSFR check would make the CSFR invalid? does it have something to do with `warden.authenticate` resetting the session? Can anyone point me to the code?
thanks

arthurnn:
Found out where csfr token resets the token: https://github.com/plataformatec/devise/blob/7d3d6fb3f04caea95e343956a88654f753b45af4/lib/devise/hooks/csrf_cleaner.rb#L5

It is a devise config, to reset the CSFR token, after an authenticate call to warden.

So if we call `current_admin` before the token gets validated, we will lose the token, because devise add a hook to warden to cleanup that token.

PR #4036, opened 2016-04-12
Add warning about default config change

ulissesalmeida:
This change add warnings for these configurations:

* `strip_whitespace_keys` - It is already explicit on config template, now
it will be the same of the template.
* `email_regexp` - In the new version this regexp will be more
permissive.
* `reconfirmable` - It is already explicit on config template, now
it will be the same of the template.
* `skip_session_storage` - It is already explicit on config template, now
it will be the same of the template.
* `sign_out_via` - It is already explicit on config template, now
it will be the same of the template.

These ones is important to change, since the configuration says current
explicit value are the default. It can lead to misunderstanging if users
remove the explicit configuration.

It also updates the template explicit values:

* Warns the `config.mailer_sender` is nil by default
* Update `config.password_length` to use the current default
* Make the e-mail configuration explicit

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -53,6 +53,12 @@ module Devise
   # True values used to check params
   TRUE_VALUES = [true, 1, '1', 't', 'T', 'true', 'TRUE']
 
+  # Track the configs that user explicit changed the default value. It is
+  # helpfull to not warn users about default values changing when they willing
+  # changed.
+  mattr_accessor :app_set_configs
+  @@app_set_configs = Set.new
+
   # Secret key used by the key generator
   mattr_accessor :secret_key
   @@secret_key = nil
@@ -82,8 +88,19 @@ module Devise
   @@case_insensitive_keys = [:email]
 
   # Keys that should have whitespace stripped.
-  mattr_accessor :strip_whitespace_keys
+  # TODO: 4.1 Do: @@strip_whitespace_keys = [:email]
+  mattr_reader :strip_whitespace_keys
   @@strip_whitespace_keys = []
+
+  def self.strip_whitespace_keys=(strip_whitespace_keys)
+    app_set_configs << :strip_whitespace_keys
+    @@strip_whitespace_keys = strip_whitespace_keys
+  end
+
+  def strip_whitespace_keys=(strip_whitespace_keys)
+    app_set_configs << :strip_whitespace_keys
+    @@strip_whitespace_keys = strip_whitespace_keys
+  end
 
   # If http authentication is enabled by default.
   mattr_accessor :http_authenticatable
@@ -104,8 +121,19 @@ module Devise
   # Email regex used to validate email formats. It simply asserts that
   # an one (and only one) @ exists in the given string. This is mainly
   # to give user feedback and not to assert the e-mail validity.
-  mattr_accessor :email_regexp
+  # TODO: 4.1 Do: @@email_regexp = [/\A[^@\s]+@[^@\s]+\z/]
+  mattr_reader :email_regexp
   @@email_regexp = /\A[^@\s]+@([^@\s]+\.)+[^@\W]+\z/
+
+  def self.email_regexp=(email_regexp)
+    app_set_configs << :email_regexp
+    @@email_regexp = email_regexp
+  end
+
+  def email_regexp=(email_regexp)
+    app_set_configs << :email_regexp
+    @@email_regexp = email_regexp
+  end
 
   # Range validation for password length
   mattr_accessor :password_length
@@ -139,8 +167,19 @@ module Devise
 
   # Defines if email should be reconfirmable.
   # False by default for backwards compatibility.
-  mattr_accessor :reconfirmable
+  # TODO: 4.1 Do: @@reconfirmable = true
+  mattr_reader :reconfirmable
   @@reconfirmable = false
+
+  def self.reconfirmable=(reconfirmable)
+    app_set_configs << :reconfirmable
+    @@reconfirmable = reconfirmable
+  end
+
+  def reconfirmable=(reconfirmable)
+    app_set_configs << :reconfirmable
+    @@reconfirmable = reconfirmable
+  end
 
   # Time interval to timeout the user session without activity.
   mattr_accessor :timeout_in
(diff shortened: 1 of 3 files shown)


lib/devise.rb
```diff
@@ -280,6 +301,32 @@ module Strategies
...
     yield self
+
+    config_defaul_change_warn(:email_regexp, '/\A[^@\s]+@([^@\s]+\.)+[^@\W]+\z/', '/\A[^@\s]+@[^@\s]+\z/')
+    config_defaul_change_warn(:reconfirmable, 'false', 'true')
+    config_defaul_change_warn(:sign_out_via, ':get', ':delete')
+    config_defaul_change_warn(:skip_session_storage, '[]', '[:http_auth]')
+    config_defaul_change_warn(:strip_whitespace_keys, '[]', '[:email]')
+  end
+
+  def self.config_defaul_change_warn(config, current_default, new_default)
```
lucasmazza: `warn_default_config_changed` ?

ulissesalmeida:
It also permits change the e-mails regex on #4001 in the Devise `4.1`

lib/devise.rb
```diff
@@ -82,8 +88,11 @@ module Strategies
   @@case_insensitive_keys = [:email]
 
   # Keys that should have whitespace stripped.
-  mattr_accessor :strip_whitespace_keys
+  # TODO: 4.1 Do: @@strip_whitespace_keys = [:email]
+  mattr_reader :strip_whitespace_keys
   @@strip_whitespace_keys = []
+  def self.strip_whitespace_keys=(strip_whitespace_keys); explicit_set_configs << :strip_whitespace_keys; @@strip_whitespace_keys = strip_whitespace_keys; end
+  def strip_whitespace_keys=(strip_whitespace_keys); explicit_set_configs << :strip_whitespace_keys; @@strip_whitespace_keys = strip_whitespace_keys; end
```
rafaelfranca: Why not multiline methods?

ulissesalmeida: 💣

ulissesalmeida: Done!

lib/devise.rb
ulissesalmeida: Done

lib/devise.rb
```diff
@@ -280,6 +341,32 @@ module Strategies
...
+    warn_default_config_changed(:reconfirmable, 'false', 'true')
+    warn_default_config_changed(:sign_out_via, ':get', ':delete')
+    warn_default_config_changed(:skip_session_storage, '[]', '[:http_auth]')
+    warn_default_config_changed(:strip_whitespace_keys, '[]', '[:email]')
+  end
+
+  def self.warn_default_config_changed(config, current_default, new_default)
+    unless config.in?(explicit_set_configs)
+      warn = <<-MESSAGE.strip_heredoc
+        [Devise] config.#{config} will have new default in the Devise 4.1
```
lucasmazza: `will have a new default on Devise 4.1`

lib/devise.rb
```diff
@@ -280,6 +341,32 @@ module Strategies
...
+        To keep the current behavior please set in your devise.rb initializer file:
```
lucasmazza: "...set in your `config/initializers/devise.rb` the following:" :question:

lib/devise.rb
```diff
@@ -280,6 +341,32 @@ module Strategies
...
+            config.#{config} = #{current_default}
+          end
+
+        If you want to use the new default:
+
+          Devise.setup do |#{config}|
+            config.#{config} = #{new_default}
+          end
+      MESSAGE
+      ActiveSupport::Deprecation.warn(warn) unless :email_regexp.in?(explicit_set_configs)
```
lucasmazza: why we have the `email_regexp` check here? Can we push it to a constant and use `include?(THINGY)` instead?

lib/devise.rb
```diff
@@ -53,6 +53,12 @@ module Strategies
   # True values used to check params
   TRUE_VALUES = [true, 1, '1', 't', 'T', 'true', 'TRUE']
 
+  # Track the configs that user explicit changed the default value. It is
+  # helpfull to not warn users about default values changing when they willing
+  # changed.
+  mattr_accessor :explicit_set_configs
```
lucasmazza: `app_set_configs` instead of  `explicit_set_configs`?

lucasmazza:
@ulissesalmeida we will be able to remove this migration code on `4.2` ?

lib/devise.rb
ulissesalmeida: :bomb: need :scissors:

ulissesalmeida: We don't move the configs to a constant because the main idea was only warn people that are using the default configurations.

lucasmazza: I meant pushing the `email_regexp` to a constant, but looks like we don't need this guard, right?

ulissesalmeida:
@lucasmazza The idea is launch this code on `4.0.0`. And remove on `4.1`. We don't need to wait `4.2`.

lib/devise.rb
ulissesalmeida: @lucasmazza Yep, we don't need this guard 😆

ulissesalmeida:
@lucasmazza @rafaelfranca Thank you for your revisions. I've done the suggestions.

lib/devise.rb
```diff
@@ -280,6 +341,32 @@ module Strategies
...
+    warn_default_config_changed(:reconfirmable, 'false', 'true')
+    warn_default_config_changed(:sign_out_via, ':get', ':delete')
+    warn_default_config_changed(:skip_session_storage, '[]', '[:http_auth]')
+    warn_default_config_changed(:strip_whitespace_keys, '[]', '[:email]')
+  end
+
+  def self.warn_default_config_changed(config, current_default, new_default)
+    unless app_set_configs.include?(config)
+      warn = <<-MESSAGE.strip_heredoc
+        [Devise] config.#{config} will have new default on Devise 4.1
```
lucasmazza: **a** new default

lib/devise.rb
```diff
@@ -280,6 +341,32 @@ module Strategies
...
+        To keep the current behavior please set in your config/initializers/devise.rb the following:
+
+          Devise.setup do |#{config}|
+            config.#{config} = #{current_default}
+          end
+
+        If you want to use the new default:
+
+          Devise.setup do |#{config}|
```
lucasmazza: shouldn't this just be `Devise.setup do |config|` instead of the interpolated `#{config}` ?

ulissesalmeida: oh lord, you're right

lleger:
Hey guys! I wanted to chime in here. I was setting up a new app with Devise today and noticed these changes. The commit changes the password length range max back to 128. We changed that in #3319 due to the limitations of BCrypt. I didn't see any discussion on this PR regarding that, so I thought I'd bring it up. Perhaps it was considered, but I thought it'd be worth pinging y'all over just in case.

reedloden:
@lleger file a new issue? Could also work around the issue by hashing the password before running it through BCrypt.

ulissesalmeida:
@lleger Thanks for point it. I think I'm out of context, then please correct if I'm wrong.

BCrypts still have the limitation of 72 characters?
A better password length default would be `6..72` characters?

lleger:
@reedloden @ulissesalmeida AFAIK it's still an issue (you can see it's [still in Rails](https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/activemodel/lib/active_model/secure_password.rb#L5-L8) as 72). I would suggest reapplying the changes from #3319, so yes, `6..72` should be fine.

PR #4038, opened 2016-04-13
Fixes issue with devise setting  as an alert message on session expir…

edgarsilva:
…ed with timeoutable.

If a session expires two alerts are shown one with message `true` and another one with `"Your session expired. Please sign in again to continue."`

There is no need to check if it is json format at this time and setup a new flash with `true`. Since that was already taking cared of at: https://github.com/plataformatec/devise/blob/master/lib/devise/failure_app.rb#L78-L89

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -78,8 +78,7 @@ module Devise
     def redirect
       store_location!
       if is_flashing_format?
-        if flash[:timedout] && flash[:alert]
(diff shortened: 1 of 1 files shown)


ulissesalmeida:
@edgarsilva Thank you for the contribution. Could you create a test to demonstrate this unexpected behaviour?

tegon:
I'm closing this in favor of #4464

PR #4039, opened 2016-04-14
Fix extented remember me

### LINKED ISSUE #3994: Extending remember me expiration enabled by default.
This commit c92996646aba2d25b2c3e235fe0c4f1a84b70d24 unwillingly enabled extending the remember me period for our application. (Devise 3.5.6)

I added this to our models to disable the behaviour:
```
private :extend_remember_period=
```

and removed (which was the default value for this config):
```
config.extend_remember_period = false
```

What would be the best way to do this in a way that will be the supported in future versions?

ulissesalmeida:
Now the config `extend_remember_period` is used to:

`true` - Every time the user cookie is authenticated, the
cookie expiration is updated.
`false` - Does not update the cookie expiration when the cookie is authenticated.

Closes #3994

### DIFF
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -39,7 +39,7 @@ module Devise
     module Rememberable
       extend ActiveSupport::Concern
 
-      attr_accessor :remember_me, :extend_remember_period
+      attr_accessor :remember_me
 
       def self.required_fields(klass)
         [:remember_created_at]
@@ -64,6 +64,10 @@ module Devise
 
       def remember_expires_at
         self.class.remember_for.from_now
+      end
+
+      def extend_remember_period
+        self.class.extend_remember_period
       end
 
       def rememberable_value
@@ -147,9 +151,6 @@ module Devise
           end
         end
 
-        private
-
-        # TODO: extend_remember_period is no longer used
         Devise::Models.config(self, :remember_for, :extend_remember_period, :rememberable_options, :expire_all_remember_me_on_sign_out)
       end
     end
--- a/test/integration/rememberable_test.rb
+++ b/test/integration/rememberable_test.rb
@@ -94,7 +94,6 @@ class RememberMeTest < Devise::IntegrationTest
     assert_response :success
     assert warden.authenticated?(:user)
     assert warden.user(:user) == user
-    assert_match /remember_user_token[^\n]*HttpOnly/, response.headers["Set-Cookie"], "Expected Set-Cookie header in response to set HttpOnly flag on remember_user_token cookie."
   end
 
   test 'remember the user before sign up and redirect them to their home' do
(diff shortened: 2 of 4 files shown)


josevalim:
:shipit:

test/integration/rememberable_test.rb
```diff
@@ -120,6 +120,40 @@ def cookie_expires(key)
...
+
+      travel_to 1.day.ago do
+        get root_path
+        old_remember_token = request.cookies['remember_user_token']
+      end
+
+      get root_path
+      current_remember_token = request.cookies['remember_user_token']
+
+      refute_equal old_remember_token, current_remember_token
```
ulissesalmeida: @josevalim My assertion was the cookie contents changing.

I could be more assertive checking the expiration keys, but the cookie is encrypted.

Worth decrypt and check the key? How I can do it?

lib/devise/models/rememberable.rb
```diff
@@ -66,6 +66,10 @@ def remember_expires_at
         self.class.remember_for.from_now
       end
 
+      def extend_remember_period
```
lucasmazza: If we remove this method, does the metaprogramming chain that backs these configs will work like the others or we need this delegation for this case?

test/integration/rememberable_test.rb
josevalim: changing is fine i would say

ulissesalmeida:
@josevalim Before I ship it, this [test](https://github.com/plataformatec/devise/blob/master/test/integration/rememberable_test.rb#L91) started to fail :(

Since we are not updating the cookie, what is the expected behavior of this feature?

lib/devise/models/rememberable.rb
ulissesalmeida: @lucasmazza Need a delegation for this case. Because that configurations methods generations are included on the "ClassMethods" of the module. The the instance objects does not have access.

lucasmazza:
:shipit:

ulissesalmeida:
@josevalim @lucasmazza Giving you more context:

That [test](https://github.com/plataformatec/devise/blob/master/test/integration/rememberable_test.rb#L91) is [failing](https://travis-ci.org/plataformatec/devise/jobs/123186522#L343) because before the cookie was always being updated. Now the default behavior is not updated anymore.

That test was checking if user authentication with remembering token.

I think this assertion does make sense anymore. Because we are not setting cookie. My suggestion is remove this assertion:
```ruby
    assert_match /remember_user_token[^\n]*HttpOnly/, response.headers["Set-Cookie"], "Expected Set-Cookie header in response to set HttpOnly flag on remember_user_token cookie."

```

josevalim:
@ulissesalmeida I believe you are right, we can change the test. We should probably double check in an actual application to make sure we are not breaking remember me.

PR #4043, opened 2016-04-18
Send confirmation instructions when a user updates the email address from nil

lmduc:
Send confirmation instructions when a user updates his `email` from `nil`.

Skip reconfirmation in the case that a record is created with #save called in callbacks in a proper way.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -254,13 +254,13 @@ module Devise
         end
 
         def postpone_email_change?
-          postpone = self.class.reconfirmable && email_changed? && email_was.present? && !@bypass_confirmation_postpone && self.email.present?
+          postpone = self.class.reconfirmable && email_changed? && !@bypass_confirmation_postpone && self.email.present?
           @bypass_confirmation_postpone = false
           postpone
         end
 
         def reconfirmation_required?
-          self.class.reconfirmable && @reconfirmation_required && self.email.present?
+          self.class.reconfirmable && @reconfirmation_required && (self.email || self.unconfirmed_email)
         end
 
         def send_confirmation_notification?
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -169,6 +169,7 @@ module Devise
         # in models to map to a nice sign up e-mail.
         def send_on_create_confirmation_instructions
           send_confirmation_instructions
+          skip_reconfirmation!
         end
 
         # Callback to overwrite if confirmation is required or not.
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -260,7 +260,7 @@ module Devise
         end
 
         def reconfirmation_required?
-          self.class.reconfirmable && @reconfirmation_required && (self.email || self.unconfirmed_email)
+          self.class.reconfirmable && @reconfirmation_required && (self.email.present? || self.unconfirmed_email.present?)
         end
 
         def send_confirmation_notification?
--- a/test/models/confirmable_test.rb
+++ b/test/models/confirmable_test.rb
@@ -488,8 +488,8 @@ class ReconfirmableTest < ActiveSupport::TestCase
   end
 
   test 'should not require reconfirmation after creating a record' do
-    user = create_admin
-    assert !user.pending_reconfirmation?
+    admin = create_admin
+    assert !admin.pending_reconfirmation?
   end
 
   test 'should not require reconfirmation after creating a record with #save called in callback' do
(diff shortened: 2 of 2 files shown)


ulissesalmeida:
@lmduc Thank you for your contribution. Nice catch! 

Which version are you using?

lmduc:
@ulissesalmeida Thank you. I am using Devise version 3.5.6

lib/devise/models/confirmable.rb
```diff
@@ -169,6 +169,7 @@ def skip_reconfirmation!
         # in models to map to a nice sign up e-mail.
         def send_on_create_confirmation_instructions
           send_confirmation_instructions
+          skip_reconfirmation!
```
printercu: @lmduc What does it solve?
It has broken updating email right after create:
```ruby
user = create(:user)
new_email = 'new@email.com'
user.update!(email: new_email)
user.unconfirmed_email == new_email # was before PR
user.email == new_email # now
```

lmduc: @printercu Hi, it is used in the case that a recored is created with #save called in callbacks. It prevents the email to be sent twice in that case.
However, this method is only called in `after_create` so it doesn't affect your update.

printercu: It does. Just try the example. It sets `@bypass_confirmation_postpone` to true and doesn't reset it after save. Why save must be called in callbacks, maybe it's just application specific demand.

You don't check what field contains new email in https://github.com/plataformatec/devise/pull/4043/files#diff-b1fd8c2bf91721612bc850fce022cb14R404

New value should be in `email` but `assert_match "new_test@example.com", admin.email` fails, 'cause it's saved to `unconfirmed_email`. This is breaking change, but it has been released as a patch.

ping @ulissesalmeida 

I've run this test in 4.0.0 (made from this pr) and it passed:
```ruby
  test 'should send confirmation instructions by email after changing email from nil' do
    admin = create_admin(email: nil)
    # assert_email_sent "new_test@example.com" do
      assert admin.update_attributes(email: 'new_test@example.com')
    # end
    # assert_match "new_test@example.com", ActionMailer::Base.deliveries.last.body.encoded

    # added this line, it fails after pr:
    assert_match "new_test@example.com", admin.email
  end
```

printercu: ping @lucasmazza 
In short there was breaking change released in 4.0.1: when updating email from `nil` it is saved to `unconfirmed email`, before this PR it was saved to `email`. Should this be fixed?

ulissesalmeida: @printercu You're right.

When we are creating an account, the e-mail doesn't exist. Then when we create the account with the "confirmable" feature, the `email` attribute is set.

When we are updating an account with an e-mail, but the previous e-mail is `nil`, it means the e-mail does not exists. For me, this situation is very similar when we are creating an account. Then it should have the same behavior, set `e-mail` not the `unconfirmed_email`.

Then it's a bug that needs to be fixed. Don't you wanna to open a PR with the test case and the patch?

Thank you for reporting! ✨

printercu: @ulissesalmeida  I'll try. What about this PR at all? Does anybody else faced same problem? After this PR updating record right after creation works different:
```ruby
user = User.create(email: some_mail)
user.update(email: new_email) # this will NOT send confirmation instructions
user.reload.update(email: new_email) # this too
User.find(user.id).update(email: new_email) # but this will
```
IMO This is not consistent behavior, PR solves some app-specific case and should not be part of Devise. WDYT?

lmduc: @printercu @ulissesalmeida  I don't think it's an app-specific case since there had been a test for it https://github.com/lmduc/devise/blob/7d10019d2a0feaba26af875ed7a7b3e3f4a1a05d/test/models/confirmable_test.rb#L503 . I tried to improve the code of it and mistakenly created a bad behavior, which caused your tests failed. In reality, there aren't many chances for users to face the same problem because there aren't many situations that the system will update emails of users right after creating their accounts. If a user updates his email, the system will find the user from the database, which will set `@bypass_confirmation_postpone` to be `false` and it will works properly. In your test, you update the email of the user right after creating him, then `@bypass_confirmation_postpone` is `true`, which cause the fail. But if you load the user from the database (which happens in real life), `@bypass_confirmation_postpone` is `false` and it works properly. 

Regarding the behavior of updating email from `nil` to a new email, I agree that the new email should be updated to `email` instead of `unconfirmed_email` since it's familiar with creating a new user.

printercu: @lmduc if there already had been this testcase you mentioned, and it hadn't been failed before your PR, so is not an example, that people faces same problem like you do.

Why do you think it should not send mail when updating user right after create?

printercu: Why don't you want add https://github.com/plataformatec/devise/blob/ac702843ddaa5652fde7a0050ff2378e7368149e/lib/devise/models/confirmable.rb#L46 to your model? It's hard to remove a callback in Rails, but still easy to add.

printercu: @lmduc i've added #4156. I don't remove or change any tests there, so your code should work fine.

lmduc: @printercu `Why do you think it should not send mail when updating user right after create?` It's an unexpected behavior and in my opinion, it's a bug and we should fix it.

printercu: Sorry, misunderstood.

PR #4050, opened 2016-04-20
Make sure attribute_will_change! method exists before calling it

cadejscroggins:
I'm using the [devise-nobrainer](https://github.com/nviennot/devise-nobrainer) ORM adapter with rails 5. I'm getting this error on the sign-in page: `undefined method 'attribute_will_change!'`. I assume this is because I'm not using ActiveRecord (this seems to be a method related to ActiveRecord ??). When I get rid of this line everything works as expected. So I added a conditional to make sure it exists before running it.

### DIFF
--- a/lib/devise/models/database_authenticatable.rb
+++ b/lib/devise/models/database_authenticatable.rb
@@ -41,7 +41,6 @@ module Devise
       # For legacy reasons, we use `encrypted_password` to store
       # the hashed password.
       def password=(new_password)
-        attribute_will_change! 'password'
         @password = new_password
         self.encrypted_password = password_digest(@password) if @password.present?
       end


ulissesalmeida:
@cadejscroggins The `attribute_will_change!` is from http://api.rubyonrails.org/classes/ActiveModel/Dirty.html

It useful to track if the attributes is changed or not, like for example `email_was_change?` from active support. 

Which version of Devise are you using?

lib/devise/models/database_authenticatable.rb
```diff
@@ -41,7 +41,7 @@ def self.required_fields(klass)
       # For legacy reasons, we use `encrypted_password` to store
       # the hashed password.
       def password=(new_password)
-        attribute_will_change! 'password'
+        attribute_will_change! 'password' if defined? attribute_will_change!
```
ulissesalmeida: @twalpole Do we really need this `attribute_will_change! `? I've removed it and all tests is passing. I think we don't this track, what do you think?

cadejscroggins:
@ulissesalmeida the latest. 22468f375684177bb9ffbc62595cf9ed10db9bbc

lib/devise/models/database_authenticatable.rb
twalpole: @ulissesalmeida when it was added it was needed for the rails 5 beta, it may not be needed anymore with the fixes/changes in later rails 5 betas

ulissesalmeida: @twalpole Thank you! :)

@cadejscroggins I think we can remove this line then, we don't need the `if`. ;)

cadejscroggins: @ulissesalmeida updated!

ulissesalmeida:
@cadejscroggins Thanks!

PR #4053, opened 2016-04-21
Make reloading of routes optional

sidonath:
In #3153 we've seen that some applications require routes to be loaded before the code is eagerly loaded so we can't just remove it. However, this implies that **all** applications using Devise need to have routes reloaded twice (as was mentioned in https://github.com/plataformatec/devise/pull/3241#issuecomment-114038540). Double route reload can incur a slowdown for larger apps that have a lot of routes or a lot of controllers.

This PR strives for compromise, making the route reloading optional.

In our app, enabling this setting shaves off 15-25% of boot time in staging environment (we have more than 1500 routes).

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -304,6 +304,10 @@ module Devise
   mattr_accessor :clean_up_csrf_token_on_authentication
   @@clean_up_csrf_token_on_authentication = true
 
+  # When false, Devise will not attempt to reload routes on eager load
+  mattr_accessor :reload_routes
+  @@reload_routes = true
+
   # PRIVATE CONFIGURATION
 
   # Store scopes mappings.
--- a/lib/devise/rails.rb
+++ b/lib/devise/rails.rb
@@ -11,7 +11,9 @@ module Devise
     end
 
     # Force routes to be loaded if we are doing any eager load.
-    config.before_eager_load { |app| app.reload_routes! }
+    config.before_eager_load do |app|
+      app.reload_routes! if Devise.reload_routes
+    end
 
     initializer "devise.url_helpers" do
       Devise.include_helpers(Devise::Controllers)


ulissesalmeida:
Looks good. Any consideration? cc @lucasmazza @josevalim

lib/devise.rb
```diff
@@ -304,6 +304,10 @@ def sign_out_via=(sign_out_via)
   mattr_accessor :clean_up_csrf_token_on_authentication
   @@clean_up_csrf_token_on_authentication = true
 
+  # When false, Devise will not attempt to reload routes on eager load
+  mattr_accessor :reload_routes
+  @@reload_routes = true
```
ulissesalmeida: In our template configuration file, we should explicit set it? cc @lucasmazza

lucasmazza: We don't need to set it on the generated config file, but we can add a commented example explaining the purpose of the setting.

ulissesalmeida: 👍 I can do it

lucasmazza: @ulissesalmeida :shipit:

sidonath:
@lucasmazza @ulissesalmeida what do you think about making this the default value in the release after the next? There are probably much more applications that don't need routes to be reloaded than those which do.

lucasmazza:
@sidonath I guess we can go down this road but I'm not sure if we can push this into a `4.2.0` release - might be worth to gather some more feedback from other developers and maybe emit a warning for the upcoming change to help out the upgrade path, considering how unrelated the errors that might surface looks at first glance.

sidonath:
Makes sense 😄 Let me know if I should add something here

ulissesalmeida:
Hey @sidonath . You can add configuration line and a explanation about this config on the `https://github.com/plataformatec/devise/blob/master/lib/generators/templates/devise.rb` template file. Explaining what happens when you set `true` and `false`, when it is safe to set to `false` and gain the benefits of not reloading twice. What do you think?

sidonath:
Sure, will do

lucasmazza:
@sidonath thanks for the patch ❤️  

I pushed the template change on db8e247aa59f88a9fb0882af224739287332a19b.

sidonath:
@lucasmazza thanks for wrapping it up, and sorry about dropping the ball

PR #4057, opened 2016-04-24
Internationalise all of the views.

galkon:
Hello, this PR is to figure out if there is much of an appetite for internationalising everything. I'm going through the process of doing this at the moment in a rails application so thought it might be worth while contributing it back so that others don't have to.

TODO:
- [x] Standard views
- [x] Simple Form views 
- [x] Markerb mailer views
- [x] Shared links
- [x] Make sure en.yml is in alphabetical order
- [x] Squash commits

At the moment the approach I'm taking is to have everything as a seperate key even if the content is the same as you can see in this example here. This is to increase the flexibility incase users of the gem want to customise the button text for example.

Thanks :)

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -1,4 +1,4 @@
-<h2>Resend confirmation instructions</h2>
+<h2><%= t('devise.confirmations.new.title') %></h2>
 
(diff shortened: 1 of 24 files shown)


josevalim:
Hi @galkon,

Thank you but we don't plan to accept this PR because the views were always meant to be a starting point and not a final version. Also, adding translation calls to them make the views much harder to understand to everyone not using i18n.

Please see  #2058 #1642 #1983 #1989 #3684 and many others similar issues/pull requests for the reasons.

PR #4064, opened 2016-04-27
Send on create confirmation email after commit

### LINKED ISSUE #4062: Asynchronous email job cannot find user
I get an exception of 
`ActiveJob::DeserializationError:  Error while trying to deserialize arguments: Couldn't find User with 'id'=xxx` while trying to follow the README to let the devise send email asynchronously. 
https://github.com/plataformatec/devise#activejob-integration

I think this is because `send_on_create_confirmation_instructions` is called `after_create`, and emails are sent before user is persisted.

allenwq:
Fix #4062 

Call `send_on_create_confirmation_instructions` in `after_commit` instead of `after_create`, I think this is no harm in general and it makes things like async job work.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ group :test do
   gem "omniauth-openid", "~> 1.0.1"
   gem "webrat", "0.7.3", require: false
   gem "mocha", "~> 1.1", require: false
+  gem 'test_after_commit', require: false
 end
 
 platforms :jruby do
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -43,9 +43,15 @@ module Devise
 
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
-        after_create  :send_on_create_confirmation_instructions, if: :send_confirmation_notification?
+        after_create :skip_reconfirmation!, if: :send_confirmation_notification?
+        if respond_to?(:after_commit) # ActiveRecord
+          after_commit :send_on_create_confirmation_instructions, on: :create, if: :send_confirmation_notification?
+          after_commit :send_reconfirmation_instructions, on: :update, if: :reconfirmation_required?
+        else # Mongoid
+          after_create :send_on_create_confirmation_instructions, if: :send_confirmation_notification?
+          after_update :send_reconfirmation_instructions, if: :reconfirmation_required?
+        end
         before_update :postpone_email_change_until_confirmation_and_regenerate_confirmation_token, if: :postpone_email_change?
-        after_update  :send_reconfirmation_instructions,  if: :reconfirmation_required?
       end
 
       def initialize(*args, &block)
@@ -169,7 +175,6 @@ module Devise
         # in models to map to a nice sign up e-mail.
         def send_on_create_confirmation_instructions
           send_confirmation_instructions
-          skip_reconfirmation!
         end
 
         # Callback to overwrite if confirmation is required or not.
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -145,6 +145,8 @@ GEM
       activesupport (>= 4.0)
       sprockets (>= 3.0.0)
     sqlite3 (1.3.11)
+    test_after_commit (1.0.0)
+      activerecord (>= 3.2)
     thor (0.19.1)
     thread_safe (0.3.5)
     tzinfo (1.2.2)
@@ -173,6 +175,7 @@ DEPENDENCIES
   rails (~> 4.2.6)
   rdoc
   sqlite3
+  test_after_commit
   webrat (= 0.7.3)
 
 BUNDLED WITH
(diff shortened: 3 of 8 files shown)


allenwq:
Got one spec failure and it passed on my  local,  have no idea > <
```
1) Failure:
ReconfirmableTest#test_should_not_require_reconfirmation_after_creating_a_record_with_#save_called_in_callback [/home/travis/build/plataformatec/devise/test/models/confirmable_test.rb:509]:
Failed assertion, no message given.
```

allenwq:
All works now, @josevalim @lucasmazza  any comments ? Thanks !

Gemfile
```diff
@@ -12,6 +12,7 @@ group :test do
   gem "omniauth-openid", "~> 1.0.1"
   gem "webrat", "0.7.3", require: false
   gem "mocha", "~> 1.1", require: false
+  gem 'test_after_commit', require: false
```
ulissesalmeida: Do we really need this gem? Can we avoid to add this dependency?

lib/devise/models/confirmable.rb
```diff
@@ -169,7 +175,6 @@ def skip_reconfirmation!
         # in models to map to a nice sign up e-mail.
         def send_on_create_confirmation_instructions
           send_confirmation_instructions
-          skip_reconfirmation!
```
ulissesalmeida: I'm not sure if we can remove this line. It was introduced in this PR https://github.com/plataformatec/devise/pull/4043/files

ulissesalmeida:
@allenwq Thank you for your PR. I think it would be good to add some spec that mimics that problem you have mentioned, this way avoids regressions in the future. If I get some idea on how to do it, I let you know.

Gemfile
allenwq: There are two ways of avoiding using this gem I can think of now:

1. Manually patch AR to support transnational fixtures and callbacks, which is pretty much the gem does, but we can remove the dependency.
2. Disable transactional fixtures, this is the cleanest way I think, but need a lot of work because the records are not deleted after each spec and they will affect each other.

@ulissesalmeida What do you think, which one do you prefer ?

lib/devise/models/confirmable.rb
allenwq: Thanks for the reference ! I didn't remove that line because one spec will fail. Instead I moved it above.

allenwq:
@ulissesalmeida Yes I will add a spec to this.

lib/devise/models/confirmable.rb
ulissesalmeida: Sorry, now I see, you've moved to a callback.

Gemfile
ulissesalmeida: @allenwq I think we could use a third option, that is:

3. On Rails 5 the after commits callbacks will run. Today the Rails solve the fixtures tests using the `uses_transaction`method. We can see how they use in the PR that supports the after commits: https://github.com/rails/rails/pull/18458/files#diff-f51660d918ba5921aaae392122026533L1310

I never used this before, but I think it worth the try. What do you think?

allenwq: Seems I have tried this before, didn't manage to succeed. I will look into it again and get back to you. Thanks !

test/integration/registerable_test.rb
```diff
@@ -1,6 +1,8 @@
 require 'test_helper'
 
 class RegistrationTest < Devise::IntegrationTest
+  uses_transaction :test_a_guest_user_should_receive_the_confirmation_instructions_from_the_default_mailer
```
allenwq: @ulissesalmeida It does not work very well, declare `uses_transaction` on one spec will break other specs, see travis build results.  I think AR just disables the transactional fixtures when you say `uses_transaction`...

ulissesalmeida: @allenwq You're right. I agree. Maybe the option of using the gem is better.

I think we can at least, can not use the gem configuration if we are on Rails 5 and see what happens.

You can chan the Rails 5 using a code like this: https://github.com/plataformatec/devise/blob/69bee06ceee6280b54304928bb6e55c5064abad8/test/rails_app/app/controllers/home_controller.rb#L23

Then, we can let a "TODO" to remove the gem and code when we stop to support versions below than Rails 5.

allenwq: Yes, my initial commit doesn't uses the gem for rails 5 :-)  and everything passed.

allenwq:
@ulissesalmeida I reverted those changes, now using the gem and it's only applied to rails 4. Working on a spec for this.

ulissesalmeida:
@allenwq Thanks! It's looking good! ✨

ulissesalmeida:
@allenwq I know test a race condition is not a easy task. I think ti worth a release and we can add regression test later.

allenwq:
Thanks ! Sorry I am on leave these days, will update the tests in a
separate PR.

allenwq:
@ulissesalmeida Sorry I cannot get the race condition test work > < but in our production app after this commit all tests passed, there are lot of random failures before.

PR #4066, opened 2016-04-27
Friendly token

sbc100:
Use friendly_token over token_generator when only raw value is needed
    
The first value returned by token_generator.generate is
simply the return value of friendly_token so this code should
be equivalent.
    
The use of token_generator here dates back to when the
confirmation_token was stored as a digest, but that is no
longer true.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -235,8 +235,7 @@ module Devise
           if self.confirmation_token && !confirmation_period_expired?
             @raw_confirmation_token = self.confirmation_token
           else
-            raw, _ = Devise.token_generator.generate(self.class, :confirmation_token)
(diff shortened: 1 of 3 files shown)


ulissesalmeida:
Looks good. Make sense since we are only using the `raw` and its the same thing of `Devise.friendly_token`. The extra computation is not necessary.

Any consideration? @josevalim @lucasmazza

lucasmazza:
I don't have enough background of the token handling here, so I guess it's up to @josevalim for the final word 🙈

josevalim:

fschwahn:
I know this is kind of old, but `Devise::TokenGenerator` checks that the generated token is not already in the database. This check does not happen for confirmable anymore. (I never had an issue with this, but it seems strange to check it for the other tokens, but not for confirmable).

This was not changed in this PR, but rather in #3661

Issue #3967, opened 2016-02-24
minimum_password_length not accessible from registrations/edit view


Yanchek99:
This may be a feature request but I was wondering what the reason behind removing the @minimum_password_length variable from the edit registrations view? 

 I think it is still beneficial to prompt the end user of the minimum password length when changing their password.  Would be nice if I could display the same message used in the registrations/new view.

ulissesalmeida:
Hi @Yanchek99 ! I don't know when it was removed. Do you know or remember which version it was removed?

I'm totally ok to bring these variables to edit/update actions and show it in some way on the edit form view. Do you want to open pull request with feature proposal?

lucasmazza:
Chased by #4068

PR #4068, opened 2016-04-29
Set minimum password length for edit and new views

Yanchek99:
Address issue #3967

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -1,11 +1,11 @@
 class Devise::RegistrationsController < DeviseController
   prepend_before_action :require_no_authentication, only: [:new, :create, :cancel]
   prepend_before_action :authenticate_scope!, only: [:edit, :update, :destroy]
+  prepend_before_action :set_minimum_password_length, only: [:new, :edit]
 
   # GET /resource/sign_up
   def new
     build_resource({})
-    set_minimum_password_length
     yield resource if block_given?
     respond_with self.resource
   end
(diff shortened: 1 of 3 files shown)


ulissesalmeida:
@Yanchek99 I think would be cool put a similar message on form like the [`new` form](https://github.com/plataformatec/devise/blob/master/app/views/devise/registrations/new.html.erb#L13). What do you think?

Yanchek99:
@ulissesalmeida I agree, however I am not sure a good place to put it because there is also the "(leave blank if you don't want to change it)" message. Perhaps under the text field?

ulissesalmeida:
@Yanchek99 👍

Yanchek99:
@ulissesalmeida updated, with suggestions.

ulissesalmeida:
@Yanchek99 Looking good! Thanks!

lucasmazza:
@Yanchek99 thanks!

mizurnix:
Hi, I think I found a small problem regarding this feature. minimum_password_length is accessible in registrations/edit view, but only if the form is submitted without errors.

User goes to registrations/edit page and it's there:

image 7

User submits incomplete form and after the redirect, it's no longer there:

image 8

This is a vanilla devise installation (I only copied views, did not change any of them): http://testdevisepasswordlength.herokuapp.com/

Do you guys agree this should be fixed? If yes, should I open a new issue?

ethirajsrinivasan:
@mizurnix can you provide the source code for your application

mizurnix:
@ethirajsrinivasan https://github.com/mizurnix/test_devise_passwrod_length

ethirajsrinivasan:
@mizurnix  #4343 Created PR to solve the same. @lucasmazza Please have a look

Issue #4072, opened 2016-04-29
Devise::Models::Lockable::ClassMethods#unlock_strategy_enabled? always returns true for :both strategy


f3ndot:
### Tested on:
- Ruby 2.2.2
- Rails 4.2.4
- Devise 3.5.8
- Warden 1.2.6

### Problem:
[`Devise::Models::Lockable::ClassMethods#unlock_strategy_enabled?`](http://www.rubydoc.info/github/plataformatec/devise/master/Devise/Models/Lockable/ClassMethods#unlock_strategy_enabled%3F-instance_method) always returns `true` for `:both` strategy

Note `:time`, `:email`, and `:none` do not exhibit this bug.

This can be problem for those who choose to extend the functionality of Lockable or rely on the method to determine code paths based on the `:none` strategy.

## Steps to Reproduce

1. Create a minimal Rails 4 app
2. Add Devise gem, run the devise install generator, generate a devise user
3. Add the `:lockable` module for user, uncomment the lockable migration lines and run migrations
4. Go into devise initializer and ensure `config.unlock_strategy = :both` is set
5. Boot up Rails console

## Expected

```
2.2.2 :001 > User.unlock_strategy_enabled? :none
 => false
2.2.2 :001 > User.unlock_strategy_enabled? :sadsa
 => false
2.2.2 :002 > User.unlock_strategy_enabled? :both
 => true
2.2.2 :003 > User.unlock_strategy_enabled? :time
 => true
2.2.2 :004 > User.unlock_strategy_enabled? :email
 => true
```

## Actual

```
2.2.2 :001 > User.unlock_strategy_enabled? :none
 => true
2.2.2 :001 > User.unlock_strategy_enabled? :sadsa
 => true
2.2.2 :002 > User.unlock_strategy_enabled? :both
 => true
2.2.2 :003 > User.unlock_strategy_enabled? :time
 => true
2.2.2 :004 > User.unlock_strategy_enabled? :email
 => true
```

As you can see `:sadsa` and `:none` were considered `true`.

rishi-freshbooks:
👍

ulissesalmeida:
Fixed on `4.1.0`, `4.0.2` and `3.5.9`.

PR #4073, opened 2016-04-29
Fix strategy checking in #unlock_strategy_enabled?

f3ndot:
The argument was used in the wrong spot and caused a bug where all values would return `true` if `self.unlock_strategy` was `:both`

See https://github.com/plataformatec/devise/issues/4072

### DIFF
--- a/lib/devise/models/lockable.rb
+++ b/lib/devise/models/lockable.rb
@@ -181,7 +181,7 @@ module Devise
 
         # Is the unlock enabled for the given unlock strategy?
         def unlock_strategy_enabled?(strategy)
-          [:both, strategy].include?(self.unlock_strategy)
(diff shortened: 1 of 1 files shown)


ulissesalmeida:
@f3ndot Thank you for the awesome issue report and for the change proposal. The build is failing with your change, could you check if the tests are updated to the correct behaviour?

f3ndot:
So two things:

1. I targeted the wrong branch (this also affects master) so I'll have to close the PR
2. Being in a sicky fog, I also wrote code that doesn't handle the cases.

Stand by for a new PR that targets master (although this bug applies to both Devise 3 and 4) and new tests that pass

f3ndot:
:bee

PR #4076, opened 2016-05-01
Fix strategy checking in #unlock_strategy_enabled? for :none and undefined strategies

f3ndot:
A bug that if the unlock strategy was set to `:both`, it would return true for all & any inputs

See #4072
PR is an improvement on #4073

### DIFF
--- a/lib/devise/models/lockable.rb
+++ b/lib/devise/models/lockable.rb
@@ -181,7 +181,9 @@ module Devise
 
         # Is the unlock enabled for the given unlock strategy?
         def unlock_strategy_enabled?(strategy)
-          [:both, strategy].include?(self.unlock_strategy)
+          self.unlock_strategy == strategy ||
+            # only :time and :email are subsets of the :both strategy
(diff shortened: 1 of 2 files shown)


lib/devise/models/lockable.rb
```diff
@@ -181,7 +181,9 @@ def unlock_access_by_token(unlock_token)
 
         # Is the unlock enabled for the given unlock strategy?
         def unlock_strategy_enabled?(strategy)
-          [:both, strategy].include?(self.unlock_strategy)
+          self.unlock_strategy == strategy ||
+            # only :time and :email are subsets of the :both strategy
+            (self.unlock_strategy == :both && [:time, :email].include?(strategy))
```
lucasmazza: can you please move the array to a `SUPPORTED_UNLOCK_STRATEGIES` (or something like that) constant?

ulissesalmeida: or maybe `SUPPORTED_BOTH_STRATEGIES`

ulissesalmeida:
Very good! Thanks!

PR #4078, opened 2016-05-02
Deprecate the bypass option of sign_in

### LINKED ISSUE #3981: sign_in(:user, store: false, bypass: true) stores the serialized user object in the session
I ran into this bug while using the following gem:

https://github.com/lynndylanhurley/devise_token_auth/blob/master/app/controllers/devise_token_auth/concerns/set_user_by_token.rb#L59

Calling both `store: false, bypass: true` causes the user object to be serialized into the session even though store is `false`.

ulissesalmeida:
The sign_in method permits the bypass option
that ignore the others options used. This behavior
has lead some users to a misconfusion what the
method really does.

This change deprecate the bypass option in favor
of a method that only does the sign in with bypass.

Closes #3981

### DIFF
--- a/lib/devise/controllers/sign_in_out.rb
+++ b/lib/devise/controllers/sign_in_out.rb
@@ -12,20 +12,15 @@ module Devise
       end
 
       # Sign in a user that already was authenticated. This helper is useful for logging
-      # users in after sign up.
-      #
-      # All options given to sign_in is passed forward to the set_user method in warden.
-      # The only exception is the :bypass option, which bypass warden callbacks and stores
-      # the user straight in session. This option is useful in cases the user is already
-      # signed in, but we want to refresh the credentials in session.
+      # users in after sign up. All options given to sign_in is passed forward
+      # to the set_user method in warden.
       #
       # Examples:
       #
       #   sign_in :user, @user                      # sign_in(scope, resource)
       #   sign_in @user                             # sign_in(resource)
-      #   sign_in @user, event: :authentication  # sign_in(resource, options)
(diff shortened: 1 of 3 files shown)


lucasmazza:
@ulissesalmeida do you want me to finish this one for ya?

ulissesalmeida:
@lucasmazza Yep! I'll finish it today 😗

lib/devise/controllers/sign_in_out.rb
```diff
@@ -44,6 +46,20 @@ def sign_in(resource_or_scope, *args)
...
+      # Sign in a user bypassing the warden callbacks and stores the user
+      # straight in session. This option is useful in cases the user is already
+      # signed in, but we want to refresh the credentials in session.
+      #
+      # Examples:
+      #
+      #   sign_in @user, scope: :user
+      #   sign_in @user
+      def bypass_sign_in(resource, scope: nil)
+        scope = Devise::Mapping.find_scope!(resource) unless scope
```
lucasmazza: `||=` instead of `unless` ?

lib/devise/controllers/sign_in_out.rb
```diff
@@ -44,6 +46,20 @@ def sign_in(resource_or_scope, *args)
         end
       end
 
+      # Sign in a user bypassing the warden callbacks and stores the user
+      # straight in session. This option is useful in cases the user is already
+      # signed in, but we want to refresh the credentials in session.
+      #
+      # Examples:
+      #
+      #   sign_in @user, scope: :user
```
lucasmazza: We should say `bypass_sign_in` instead of `sign_in` on these examples.

Issue #3705, opened 2015-08-12
`FailureApp`s `script_name: nil` breaks route generation within mounted engines


jvdp:
Since upgrading to Devise 3.5.2, the `Devise::FailureApp` from within a Rails engine no longer generates routes with the path it is mounted under.

The breaking change was made in https://github.com/plataformatec/devise/commit/aa675f7f6611de5d40114aa98153f3f0c0b3e8d5, which always passes a `script_name` (possibly nil) to the url generator. This is problematic because the mount point path is only computed when `script_name` is not passed to `options`: https://github.com/rails/rails/blob/v4.2.3/actionpack/lib/action_dispatch/routing/mapper.rb#L631

I'm not sure how to fix this (this is all crazy complicated code), but perhaps there's a better fix for #3643 than forcing `path_name` to be passed to `url_for`.

josevalim:
Which Rails version are you using?

jvdp:
Rails 4.2.3.

josevalim:
@jvdp so apparently if we don't set it, there is a infinite loop. If we set it, we break engines. :(

jvdp:
Yep, it sucks. Do you think this should be fixed in Rails instead?

mmitchellg5:
We're running into this as well, with Rails 4.2.4

variousred:
FWIW as a test I deleted
```
      # Rails 4.2 goes into an infinite loop if opts[:script_name] is unset
      if (Rails::VERSION::MAJOR >= 4) && (Rails::VERSION::MINOR >= 2)
        opts[:script_name] = (config.relative_url_root if config.respond_to?(:relative_url_root))
      else
        if config.respond_to?(:relative_url_root) && config.relative_url_root.present?
          opts[:script_name] = config.relative_url_root
        end
      end
``` and I did NOT get an infinite loop, everything worked.

mmitchellg5:
@jvdp @josevalim

jvdp:
@mmitchellg5 hey if that fixes it, great, but doesn't that break a configuration option?

For completeness I think #3738 is a duplicate of this (and #3739 a proposed fix.)

variousred:
@jvdp sure does, but you can just delete the conditional since apparently rails 4.2.4 does NOT go into an infinite loop without the setting.

stanhu:
[...]

Issue #3613, opened 2015-06-02
Using authenticated routes with mounted engines causes a redirect loop


ajhodges:
Using Rails 4.2.1, Devise 3.5.1, Warden 1.2.3

This is a regression in 3.5.x. I have `rails_admin` and `sidekiq` mounted under an authenticated admin route in my app like so:

    devise_for :admin

    authenticate :admin do
      scope '/admin' do
        #reqsue admin panel
        mount Sidekiq::Web => "/sidekiq"

        #rails_admin admin panel
        mount RailsAdmin::Engine => '/', as: 'rails_admin'
      end
    end

The behavior with devise 3.4.1 is that if I navigate to `/admin` or `/admin/sidekiq` and I am not authenticated as an `Admin`, I get redirected to `/admin/sign_in`. After upgrading to 3.5.1, I am redirected to `/admin/admin/sign_in` and `/admin/sidekiq/admin/sign_in` respectfully. This causes a redirect loop.

benjreinhart:
I have a very similar issue, also with Rails 4.2.1, Devise 3.5.1 and Warden 1.2.3.

I have the following setup:

```ruby
devise_for :users, skip: [:sessions, :registrations]

as :user do
  get 'login' => 'devise/sessions#new', :as => :new_user_session
end

namespace :admin do
  authenticate :user, -> (u) { u.is_admin? } do
    mount Sidekiq::Web => '/sidekiq'
  end
end
```

The desired functionality is that when a user is not logged in, they should be redirected to `/login`. In Devise 3.4.1 this is the case. However, in 3.5.1 they are redirected to `/admin/sidekiq/login`.

If I move the `authenticate` call out of the `namespace :admin` block but mount `Sidekiq` to `/admin/sidekiq`, I also get a redirect to `/admin/sidekiq/login` when a user is not logged in:

```ruby
authenticate :user, -> (u) { u.is_admin? } do
  mount Sidekiq::Web => '/admin/sidekiq'
end
```

This redirects to `/admin/sidekiq/login`.

Lastly, if I change the above to be mounted at just `/sidekiq`, I get a redirect to `/sidekiq/login` when user is not logged in:

```ruby
authenticate :user, -> (u) { u.is_admin? } do
  mount Sidekiq::Web => '/sidekiq'
end
```

This redirects to `/sidekiq/login`.

siruguri:
[...]

PR #4081, opened 2016-05-03
Fix authenticated engine routes

basiszwo:
_Using Rails 5.0.0.beta4, Devise Edge (latest master), Warden 1.2.6_

This pull request solves the issue of an infinite redirects for mounted engines when not being logged in.

This happens in devises' failure app. I was using rails 5 beta 3 when addressing the issue. I could reproduce the broken behaviour in rails 4.2.6 also.

First I addressed rails 5 by relaxing the rails 4.2 constraint. Then fixed the infinite loop by initializing `opts[:script_name]` with `nil`.

When investigating the behaviour I found the following related issues:

* https://github.com/plataformatec/devise/issues/3613
* https://github.com/plataformatec/devise/issues/3643
* https://github.com/plataformatec/devise/issues/3705

There's also https://github.com/plataformatec/devise/pull/3807 which also addresses the main issue but imo with unnecessary conditionals.

The test suite is green as can be seen at https://travis-ci.org/basiszwo/devise/builds/127469964
I am sure the devise travis will also report green state.

Anyway I am not sure how to cover this issue in the test suite. You may point me to the spot?

I love to see feedback from you. Thank you very much!


lucasmazza:
> Anyway I am not sure how to cover this issue in the test suite. You may point me to the spot?

That's the hard question 😩  maybe we should have a mountable engine on on our test app and have a test that hits one of its URLs?

ulissesalmeida:
This error still happening on Rails 5 rc?

basiszwo:
@ulissesalmeida just tried. latest devise from master with rails rc 1 still has the problem. I try to provide a test within the next few days.

basiszwo:
@ulissesalmeida please see the test in my last commit. Runs on my machine. Waiting for travis to pass. In case you like it that way I squash the commits.

lucasmazza:
@basiszwo awesome test, thanks ❤️

rubydev:
In my application it doesn't work:

rails 5.0.0.1
devise 4.2.0
sidekiq 4.2.6

<pre>
  authenticate :user, lambda { |u| u.admin? } do
    mount Sidekiq::Web => '/sidekiq'
  end
</pre>

Can you help me?

basiszwo:
I didn't have time to look into this with a blank slate application but I am using this in six different applications and it works with the versions from above. 

I have to add that we use this as follows

```ruby
authenticate :user do
  mount Sidekiq::Web => '/sidekiq'
end
```

Meaning without the additional lambda you are passing in. 

Can you verify that it works without the lambda block?

rubydev:
Without lambda also doesn't work.

basiszwo:
We have a spec proofing that exactly this is working. 

maybe some other gem or configuration is interfering? you may verify this with a vanilla rails app including only devise and sidekiq.

sorry that I can't help you more. As said there is a spec in devise ensuring this scenario plus I run six applications with the versions from above and they work fine and as expected.

jeffc-sn:
I'm seeing this fail with Sidekiq as well.

Rails 4.2.8
Devise 4.2.1
Sidekiq 4.2.9

I get too many redirects when not logged in for both:
```ruby
authenticate :user do
  mount Sidekiq::Web => '/sidekiq'
end
```
and
```ruby
authenticate :user, -> (u) { u.super_admin? } do
  mount Sidekiq::Web => '/sidekiq'
end
```

artfullyContrived:
Also fails for me

```
Using sidekiq 4.2.10
Using rails 4.0.13
Using devise 3.5.10
```

PR #4086, opened 2016-05-05
Fix Problem with case insentivity comparison(on mysql) when do confirmation

wendy0402:
Hi guys,
I found out that when I do confirmation, the token comparison is case insensitive. This downgrades the entropy on the token. After do investigation I found out that by default mysql  comparison is case insensitive. 
screen shot 2016-05-05 at 12 34 22 pm
screen shot 2016-05-05 at 12 34 32 pm

For case sensitive comparison in mysql there are several choices such as do [binary comparison](http://dev.mysql.com/doc/refman/5.0/en/charset-binary-op.html), or change the collation to `utf8_bin`(CMIIW).

However when developer generate user table migration using devise, the collation is not `utf8_bin`
Therefore, I make it like this

source for case insensitivity comparison in mysql: http://stackoverflow.com/questions/3936967/mysql-case-insensitive-select

I've tried to make test but it's useless since in test environment devise use sqlite(which is case sensitive, I've tried it)

thank you

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -301,7 +301,9 @@ module Devise
         # Options must have the confirmation_token
         def confirm_by_token(confirmation_token)
           confirmable = find_first_by_auth_conditions(confirmation_token: confirmation_token)
-          unless confirmable
+          # need `confirmable.confirmation_token != confirmation_token`
+          # because mysql by default is case insensitive
+          if !confirmable || confirmable.confirmation_token != confirmation_token
             confirmation_digest = Devise.token_generator.digest(self, :confirmation_token, confirmation_token)
             confirmable = find_or_initialize_with_error_by(:confirmation_token, confirmation_digest)
           end


lucasmazza:
Adding `confirmation_token` to the `case_insensitive_keys` setting could be an alternative solution to this?

/summon @josevalim

wendy0402:
what I see `case_insensitive_keys` make the comparison to be case_insensitive(based on https://github.com/plataformatec/devise/wiki/How-To:-Use-case-insensitive-emails). What I tried to achieve here is the opposite. Make the token to be case sensitive

wendy0402:
how is it going with this one @lucasmazza ?

lucasmazza:
@wendy0402 sorry - haven't had the change to look into this one over the last week - gonna see if we can dig into this on the following days /cc @ulissesalmeida @josevalim

lucasmazza:
I feel like this one is more of a MySQL issue rather than something we should deal in Devise.

@ulissesalmeida @josevalim what do you both think?

tegon:
I'm closing this since it looks more like a MySQL issue.

PR #4100, opened 2016-05-10
Update README.md

shinyaK14:
I've added $ mark into console instruction.
$ marks help us to understand which is console and which is Ruby's source code.
It's the same as [Rails README.md](https://github.com/rails/rails/blob/master/README.md)

**Before:**
```ruby
gem 'devise'
```
```console
rails generate devise:install
```

**After:**
```ruby
gem 'devise'
```
```console
$ rails generate devise:install
```

Thank you.


nashby:
Thanks!

PR #4104, opened 2016-05-12
Fix suggested omniauth path

LucasAU:
The suggested omniauth path noted in the CHANGELOG.md under 4.0.0.rc2 
did not work. It was missing an `_omniauth` in the middle of the 
method name.

user_github_authorize_path => user_github_omniauth_authorize_path


lucasmazza:
Thanks 😬

PR #4105, opened 2016-05-12
Update example code in authenticatable model

giedriusr:
Hello,

Just found out that the example of the original method is not 100% the same, so proposing to keep in sync. Not sure if the comment should be there either. Please let me know. Thanks.


ulissesalmeida:
It's looks good. cc @lucasmazza Any consideration?

PR #4125, opened 2016-05-20
Change behavior to skip reconfirmation after creating a record with #save called in callback

ArneZsng:
The following commit was breaking stuff on our system for a new email confirmation when the email is changed while the object is still in memory:
https://github.com/plataformatec/devise/commit/1d57169c7bd12977a6697a5a06cda265442bb5c8

This is due to `skip_reconfirmation!` in `send_on_create_confirmation_instructions` because it sets `@bypass_confirmation_postpone` to true which lets `postpone_email_change?` return false.

The case was added to the confirmable test case. Basically, the skipping of the reconfirmation as part of an after_create hook needs to be treated differently from a general bypass. To ensure that this is skipped when save is called in an after_create hook but NOT if the email is changed from an existing email, additional checks were added.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -43,7 +43,7 @@ module Devise
 
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
-        after_create :skip_create_confirmation!, if: :send_confirmation_notification?
+        after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
         if respond_to?(:after_commit) # ActiveRecord
           after_commit :send_on_create_confirmation_instructions, on: :create, if: :send_confirmation_notification?
           after_commit :send_reconfirmation_instructions, on: :update, if: :reconfirmation_required?
@@ -56,7 +56,7 @@ module Devise
 
       def initialize(*args, &block)
         @bypass_confirmation_postpone = false
-        @skip_create_confirmation = false
+        @skip_reconfirmation_in_callback = false
         @reconfirmation_required = false
         @skip_confirmation_notification = false
         @raw_confirmation_token = nil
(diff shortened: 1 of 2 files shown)


lib/devise/models/confirmable.rb
```diff
@@ -165,6 +166,12 @@ def skip_reconfirmation!
 
       protected
 
+        # To not require reoncfirmation after creating with #save called in a
```
ulissesalmeida: A little typo here `reconfirmation`

ulissesalmeida:
Looks good to me. Any consideration @josevalim @lucasmazza 

I think it may need a backport to `3.5.x`.

lib/devise/models/confirmable.rb
```diff
@@ -165,6 +166,12 @@ def skip_reconfirmation!
 
       protected
 
+        # To not require reconfirmation after creating with #save called in a
+        # callback call skip_create_confirmation!
+        def skip_create_confirmation!
```
lucasmazza: @ArneZsng @ulissesalmeida I feel like we should have better names to express the checks for dealing with the save + after_create scenarios - `skip_create_confirmation` sounds like we are skipping the confirmation on a regular create scenario, regardless of the callback issue.

ArneZsng:
@lucasmazza I updated the naming to `skip_reconfirmation_in_callback` (method and instance var). Would that fit better?

ArneZsng:
@lucasmazza @ulissesalmeida Looks good to you now?

ArneZsng:
@lucasmazza @ulissesalmeida Anything outstanding from your side?

ulissesalmeida:
:shipit:

lucasmazza:
@ArneZsng thanks for the patch, and sorry for the delay 😬

PR #4146, opened 2016-06-09
Add missing support of  Rails.application.config.action_controller.relative_url_root

kosdiamantis:
According to Ruby on Rails documentation there are 2 ways deploy your application to a subdirectory.

[doc: ](http://edgeguides.rubyonrails.org/configuring.html#deploy-to-a-subdirectory-relative-url-root)
```
class Application < Rails::Application
  ...
  config.relative_url_root = '/my_route'
  ...
end
```

or

[doc: ](http://edgeguides.rubyonrails.org/configuring.html#configuring-action-controller)
```
class Application < Rails::Application
  ...
  config.action_controller.relative_url_root = '/my_route'
  ...
end
```

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -252,5 +252,17 @@ module Devise
     def request_format
       @request_format ||= request.format.try(:ref)
     end
+
+    def relative_url_root
+      @relative_url_root ||= begin
+        config = Rails.application.config
+
+        config.try(:relative_url_root) || config.action_controller.try(:relative_url_root)
+      end
+    end
+
+    def relative_url_root?
+      relative_url_root.present?
+    end
   end
 end
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -50,13 +50,11 @@ module Devise
     end
 
     def recall
-      config = Rails.application.config
-
-      header_info = if config.try(:relative_url_root)
-        base_path = Pathname.new(config.relative_url_root)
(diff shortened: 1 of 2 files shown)


lib/devise/failure_app.rb
```diff
@@ -252,5 +246,17 @@ def is_flashing_format?
     def request_format
       @request_format ||= request.format.try(:ref)
     end
+
+    def relative_url_root
+      @relative_url_root ||= begin
+        config = Rails.application.config
+
+        config.try(:relative_url_root) || config.try(:action_controller).try(:relative_url_root)
```
lucasmazza: There is any scenario where the `config.action_controller` call would fail for it to require the `try` approach on it?

kosdiamantis: You are right, there is no need to use `try`, i will change it.
[guide#4.2.x](http://guides.rubyonrails.org/v4.2/configuring.html#configuring-action-controller)
[guide#4.1.x](http://guides.rubyonrails.org/v4.1/configuring.html#configuring-action-controller)
[guide#4.0.x](http://guides.rubyonrails.org/v4.0.8/configuring.html#configuring-action-controller)
[guide#3.2.x](http://guides.rubyonrails.org/v3.2.18/configuring.html#configuring-action-controller)
[guide#2.3.x](http://guides.rubyonrails.org/v2.3.11/configuring.html#configuring-action-controller)

kosdiamantis: changed 👌

PR #4153, opened 2016-06-18
Fix Rails 5 env deprecation warning

krishandley:
Using Rails 5.0.0.rc1 and the remember_me controller method from Devise::Controllers::Rememberable I was getting this warning a lot in my tests ```DEPRECATION WARNING: env is deprecated and will be removed from Rails 5.1```


lucasmazza:
Thanks!

Issue #14, opened 2009-10-29
Improve update email process


gavinhughes:
## The current update email process needs some love.  I suggest:

Given I am signed in
When I update my email address
Then a confirmation email should be sent
And I should be on my account page
And I should see a message "A confirmation email has been sent to your new email address. Your old email address, {resource.email}, is still active until you confirm the new one."
When I follow the link in the confirmation email
Then I should be on my account page
## And I should see a message "Your email address has been confirmed."

This is a bit more complicated, but I think it's best practice.  The User model will need a :pending_email attribute.  You could go a bit further and add something like this in a Devise-generated resource show view:

-if resource.pending_email
   You have requested to change your email address to {resource.pending_email}. An email has been sent there to make sure it is a valid address.
   = link_to 'Resend confirmation email', ...
   = link_to 'Cancel request', ...

Having all this by right out the box would be beautiful!

PR #4156, opened 2016-06-19
Fixed: changing email from nil updates email but not unconfirmed_email

printercu:
https://github.com/plataformatec/devise/pull/4043#discussion_r64995026

@ulissesalmeida here is failing test: https://github.com/plataformatec/devise/blob/ac702843ddaa5652fde7a0050ff2378e7368149e/test/models/confirmable_test.rb#L161
I can change PR to make it pass, but why it doesn't send confirmation mail to new email?

I've found https://github.com/plataformatec/devise/commit/8c1bab4951eaabc11d942a334b9ac620e54a53a8, its related to #14, but I can't understand the decision from conversation.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -43,20 +43,30 @@ module Devise
 
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
-        after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
+        before_update :postpone_email_change_until_confirmation_and_regenerate_confirmation_token, if: :postpone_email_change?
         if respond_to?(:after_commit) # ActiveRecord
           after_commit :send_on_create_confirmation_instructions, on: :create, if: :send_confirmation_notification?
-          after_commit :send_reconfirmation_instructions, on: :update, if: :reconfirmation_required?
+          after_commit on: :update do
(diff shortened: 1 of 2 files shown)


ulissesalmeida:
@printercu 

Thank you for point out all the the discussion about send confirmation mail to a new e-mail. Looking the thread, the confirmation of the new e-mail was avoided because it can't simply use the sign up confirmation process. The process of update will have different result, thinking in a scenario that we already have a confirmed e-mail and the new e-mail wasn't confirmed, should we block the account? I think this answer will be different depending of the application needs. But of course, it's my supposition. I think @josevalim may remember the entire context...

Anyway, if we want to apply this fix like a patch, we can't change this behaviour.

Thanks for the PR \o/

printercu:
@ulissesalmeida updating email doesn't block an account: https://github.com/plataformatec/devise/blob/master/lib/devise/models/confirmable.rb#L140-142
It uses `!confirmation_required?` which is `!confirmed_at?` (through a chain of methods). Updating email doesn't change `confirmed_at`, it just sets `unconfirmed_email`.

Contrariwise, when user updates email he doesn't see that email is updated, and get message that he must confirm it. However devise doesn't send confirmation message at this point and user must ask for resending this message manually.

tegon:

Thank you!

PR #4159, opened 2016-06-21
fixed : ActiveRecord check

dfabreguette-ap:
Seems like "send_reconfirmation_instructions" (line 49) was triggered because Mongoid 5 now implements :after_commit callbacks.

Just changed the active record check based on the ActiveRecord::Base definition.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -44,7 +44,7 @@ module Devise
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
         after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
-        if respond_to?(:after_commit) # ActiveRecord
+        if defined?(ActiveRecord::Base) # ActiveRecord
           after_commit :send_on_create_confirmation_instructions, on: :create, if: :send_confirmation_notification?
           after_commit :send_reconfirmation_instructions, on: :update, if: :reconfirmation_required?
         else # Mongoid


lucasmazza:
@dfabreguette-ap if Mongoid implements `after_commit` callbacks hopefully Devise should work the same as with Active Record - can you elaborate which problems can appear with the current code that justifies this change? Thanks.

lucasmazza:
Going through the open issues I found #4160 - sorry for the earlier commit. Can you please push a sample app that replicates the error or write a standalone test case for it (or on devise test suite itself)?

dfabreguette-ap:
All right, I'm gonna work on that and get back soon. 
I came across this issue upgrading an app from rails 4.1 to 4.2 which lead to mongoid 4 to 5 upgrade...

fjg:
Actually it's not defined in mongoid gem as said [here](https://jira.mongodb.org/browse/MONGOID-3520)

but in mongoid-paperclip [here](https://github.com/mrrooijen/mongoid-paperclip/blob/master/lib/mongoid_paperclip.rb)

@lucasmazza can you change the way devise detect Mongoid/Activerecord, like merging the PR?

lib/devise/models/confirmable.rb
```diff
@@ -44,7 +44,7 @@ module Confirmable
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
         after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
-        if respond_to?(:after_commit) # ActiveRecord
+        if defined?(ActiveRecord::Base) # ActiveRecord
```
texpert: Neither respond_to?(:after_commit), nor defined?(ActiveRecord::Base) are not working. What I found is:
```
if self.is_a?(ActiveRecord::Base)
```

tegon: Makes sense. Using the `defined?` could also bring some problems for projects that use both `mongoid` and `active_record` at the same time.

tegon:
@dfabreguette-ap Are you going to add a test case for this?

tegon:
Closing in favor of #4302

PR #4163, opened 2016-06-24
Fix line in README.md

khalidharun:
Quick little piece that seems out-of-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,8 +432,8 @@ If you're using RSpec, you can put the following inside a file named
 
 ```ruby
 RSpec.configure do |config|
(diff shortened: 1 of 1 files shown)


lucasmazza:
@khalidharun it isn't - the README states the code for the current `master` branch, which currently has unreleased changes. `Devise::Test::ControllerHelpers` is not in Devise 4.1.1, it's only in the current master branch. If you [browse the code for 4.1.1](https://github.com/plataformatec/devise/tree/v4.1.1) you'll find that it still uses `Devise::TestHelpers`, and the documentation in the README is correct.

khalidharun:
@lucasmazza Thanks for pointing that out. 👍

PR #4168, opened 2016-06-25
Update README.md

bgreg:
Explain the `MODEL` replacement before the command is shown to prevent confusion.


bgreg:
Cool, thanks!

PR #4169, opened 2016-06-26
Test cases (not signed-in 302 redirection) failed on Rails5 RC2

hamadata:
As reported here ( https://github.com/rails/rails/issues/25269 ),  ActionDispatch::IllegalStateError is raised on Rails5 RC2.

This is a workaround to solve this error.


ronyv89:
👍

lucasmazza:
Awesome, thanks! I'll review some of the pending tickets over this week before cutting a new release, so anyone that is running Rails rc2 and using controller tests should use Devise master for a while before upgrading to `4.2.0`.

Issue #4145, opened 2016-06-08
Send dedicated reconfirmation email


basiszwo:
Devise offers the option to reconfirm an email address that has been changed. This is great and very useful. 

A tiny flaw in this process is that devise sends out the normal (account) confirmation email which is kind of misleading for a user that just changed his email address. 

I'd be happy to provide a PR for a separate email when confirming an updated / changed email address.

Waiting for feedback now. Thanks.

lucasmazza:
@basiszwo if you want to tackle this, please do - I asked myself a similar question when working with a custom reconfirmable email recently, and would be awesome to have it as a default on Devise - let's just have a good upgrade path so folks upgrading Devise won't get the new email template without opting in through a global configuration for now, but we can make this behavior a default in a next major release.

SaiVardhan:
#4182 addresses this

leewaa:
Nothing happening here ? Over a year old now

rodrigobdz:
Any plans for this to be merged soon?

PR #4182, opened 2016-07-04
Issue#4145-Send dedicated reconfirmation email

SaiVardhan:


lucasmazza:
@SaiVardhan thanks a lot for this Pull Request! :)
Can you please add a setting under the `Devise` namespace to disable this feature for now? When upgrading Devise to a new version, is not completely desired for it to start sending new emails on existing apps, as developers might want to tweak the email layout first before rolling it out to their users - we can enable this by default on Devise 5 but we should have a feature toggle for it for now.

SaiVardhan:
@lucasmazza ,

Can you give me a road map of adding a setting under the *devise* namespace

lucasmazza:
@SaiVardhan here is the gist of what we need to do:

1 - Add a new`mattr_accessor` attribute on [`lib/devise.rb`](https://github.com/plataformatec/devise/blob/master/lib/devise.rb#L156-L158) similar to the ones we already have, with a default of `false` and a `# TODO: ` mentioning that we should turn it `true` on a future major release.
2 - On [`lib/generators/template/devise.rb`](https://github.com/plataformatec/devise/blob/master/lib/generators/templates/devise.rb#L116), we can add a commented section explaining this new setting and how developers can turn it on.
3 - Change `can_send_reconfirmation?` to account for this new flag so developers need to turn it on when upgrading their apps.

Hope this helps!

lib/devise.rb
```diff
@@ -145,6 +145,10 @@ module Test
   mattr_accessor :reconfirmable
   @@reconfirmable = true
 
+  #TODO- we should turn it true on a future major release
+  mattr_accessor :new_email_confirmable
```
lucasmazza: What do you think of using `send_reconfirmation_instructions` for this accessor instead?

SaiVardhan:
@lucasmazza ready :)

app/controllers/devise/registrations_controller.rb
```diff
@@ -83,9 +83,10 @@ def cancel
   protected
 
   def update_needs_confirmation?(resource, previous)
-    resource.respond_to?(:pending_reconfirmation?) &&
-      resource.pending_reconfirmation? &&
-      previous != resource.unconfirmed_email
+    resource.respond_to?(:can_send_reconfirmation?) &&
```
lucasmazza: I don't think it's OK to use `can_send_reconfirmation?` over `pending_reconfirmation?` inside the controller/view, as we can let the the controller do its work without the model sending the new email. Can you please revert these changes?

lib/generators/templates/devise.rb
```diff
@@ -135,6 +135,11 @@
   # unconfirmed_email column, and copied to email column on successful confirmation.
   config.reconfirmable = true
 
+  #If true, requires any email changes to be confirmed (exactly the same way as
+  # initial account confirmation) to be applied.
+
+  config.new_email_confirmable = false
```
lucasmazza: This template should use the `send_reconfirmation_instructions` from `lib/devise.rb` :)

app/controllers/devise/registrations_controller.rb
SaiVardhan: as we are getting a flash notice if we use `pending_reconfirmation?` and toggle is off so I made these changes to make it unique

SaiVardhan: @lucasmazza , any thoughts

lucasmazza: which flash notice message? I don't think I follow.

I'm trying to go down the path that this patch should only change the mailer template we use for the reconfirmation cases, and shouldn't affect anything on the user flow or UI feedback we already have - that's the whole scope of what was discussed on plataformatec/devise#4145.

SaiVardhan: I can see this flash notice 
`You updated your account successfully, but we need to verify your new email address. Please check your email and follow the confirm link to confirm your new email address.`

So I did these changes as this should be displayed only when  toggle for reconfirmation is ON

lucasmazza: @SaiVardhan our new toggle shouldn't affect this logic, the reconfirmation works even when using the old email template. Let's go back to use `pending_reconfirmation?` instead of `can_send_reconfirmation?` here.

app/views/devise/registrations/edit.html.erb
```diff
@@ -8,7 +8,7 @@
     <%= f.email_field :email, autofocus: true %>
   </div>
 
-  <% if devise_mapping.confirmable? && resource.pending_reconfirmation? %>
+  <% if devise_mapping.confirmable? && resource.can_send_reconfirmation? %>
```
lucasmazza: Same thing here - the `can_send_reconfirmation?` shouldn't replace `pending_reconfirmation?` on controllers and views.

lib/devise/models/confirmable.rb
```diff
@@ -121,8 +125,18 @@ def send_reconfirmation_instructions
         @reconfirmation_required = false
 
         unless @skip_confirmation_notification
-          send_confirmation_instructions
```
lucasmazza: we should still can `send_confirmation_instructions` if `Devise.send_reconfirmation_instructions` is false, otherwise we will break existing apps.

lib/devise/models/confirmable.rb
```diff
@@ -125,6 +129,16 @@ def send_reconfirmation_instructions
         end
       end
 
+      def send_new_email_confirmation_instructions
```
lucasmazza: Do we call this method anywhere in confirmation workflow? We should be triggering the new notification inside `send_confirmation_instructions` (instead of the existing `confirmation_instructions`).

tegon:
Hi @SaiVardhan, thanks for your pull request.
Would you mind doing a rebase with master and take a look at the last comment?
Thanks.

tegon:

Thank you!

Issue #2341, opened 2013-03-22
Send multipart emails


tmaier:
Since #1326 devise has markerb support, which is great.

Unfortunately, it just sends the HTML email, as the call to `mail` at https://github.com/plataformatec/devise/blob/master/lib/devise/mailers/helpers.rb#L16 does not have a block, like 

``` ruby
    mail(:to => @recipient, :from => "john.doe@example.com") do |format|
      format.text
      format.html
    end
```

https://github.com/plataformatec/markerb

arturoherrero:
How about [create a custom mailer for devise][1] like this:

```
class MyMailer < Devise::Mailer
  default parts_order: [ "text/plain", "text/html" ]
end
```

[1]: https://github.com/plataformatec/devise/wiki/How-To:-Use-custom-mailer

PR #4193, opened 2016-07-13
add block passthrough to devise_mail

3wille:
ActionMailer's ``mail`` method may receive a block for customizing the mails
format
``devise_mail`` now has the same functionality by just passing the block to ``mail`` call.

fixes plataformatec/devise#2341

### DIFF
--- a/test/mailers/mailer_test.rb
+++ b/test/mailers/mailer_test.rb
@@ -1,0 +1,20 @@
+require "test_helper"
+
+class MailerTest < ActionMailer::TestCase
+  test "pass given block to #mail call" do
+    class TestMailer < Devise::Mailer
+      def confirmation_instructions(record, token, opts = {})
+        @token = token
+        devise_mail(record, :confirmation_instructions, opts) do |format|
+          format.html(content_transfer_encoding: "7bit")
+        end
+      end
+    end
+
+    Devise.mailer = TestMailer
+    create_user
+    mail = ActionMailer::Base.deliveries.first
+
+    assert mail.content_transfer_encoding, "7bit"
+  end
+end
--- a/test/mailers/mailer_test.rb
+++ b/test/mailers/mailer_test.rb
@@ -11,8 +11,7 @@ class MailerTest < ActionMailer::TestCase
       end
     end
 
-    Devise.mailer = TestMailer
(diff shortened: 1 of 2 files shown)


3wille:
Hey there,
I am not sure where to test this feature, if someone could point me into the right direction, I would be very pleased :)

lucasmazza:
@3wille I think you can create a new test case on a new file `test/mailers/mailer_test.rb` and write a test that sends a dummy multipart email :)

3wille:
@lucasmazza it's a somewhat weird way of testing I used, I tried to explain it a bit in the commit message. I think it's the best way to test this without changing the actual code just for testing.

test/mailers/mailer_test.rb
```diff
@@ -0,0 +1,20 @@
...
+    class TestMailer < Devise::Mailer
+      def confirmation_instructions(record, token, opts = {})
+        @token = token
+        devise_mail(record, :confirmation_instructions, opts) do |format|
+          format.html(content_transfer_encoding: "7bit")
+        end
+      end
+    end
+
+    Devise.mailer = TestMailer
```
lucasmazza: we can replace this setup with something like this:

```ruby
TestMailer.confirmation_instructions(create_user, 'confirmation-token').deliver_now
```

This way we don't change the global `Devise.mailer` or depend on `create_user` sending the email for us.

lucasmazza:
@3wille the subclassing is OK - is usually a good way to test inherited APIs or behavior that needs some tweaking to make it testable - I just left a comment regarding the test intent and side effects.

test/mailers/mailer_test.rb
3wille: @lucasmazza we don't need ``deliver_now`` in this case (it also fails on Rails 4.1, because it's ``deliver`` there).
``confirmation_instructions`` returns the mail object we want to test.

lucasmazza: @3wille makes sense - so we only need to call the class method to create the mail object.

3wille:
@lucasmazza you think this is done now?

lucasmazza:
@3wille looks perfect - Thanks! :heart:

Issue #4189, opened 2016-07-12
Rails 5/RSpec/Device - method_missing: undefined method setup


Blazy7:
Hello, after upgrade from Rails 5.0.0rc1 to Rails 5.0.0 and try to `bundle exec rspec spec/` i get error `method_missing': undefined method `setup'

```
/Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/rspec-core-3.5.1/lib/rspec/core/example_group.rb:732:in `method_missing': undefined method `setup' for RSpec::ExampleGroups::AdminOrdersController:Class (NoMethodError)
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/devise-4.2.0/lib/devise/test/controller_helpers.rb:28:in `block in <module:ControllerHelpers>'
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/activesupport-5.0.0/lib/active_support/concern.rb:120:in `class_eval'
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/activesupport-5.0.0/lib/active_support/concern.rb:120:in `append_features'
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/devise-4.2.0/lib/devise/test_helpers.rb:9:in `include'
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/devise-4.2.0/lib/devise/test_helpers.rb:9:in `block in included'
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/devise-4.2.0/lib/devise/test_helpers.rb:4:in `class_eval'
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/devise-4.2.0/lib/devise/test_helpers.rb:4:in `included'
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/rspec-core-3.5.1/lib/rspec/core/configuration.rb:2063:in `include'
	from /Users/Hristo/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/rspec-core-3.5.1/lib/rspec/core/configuration.rb:2063:in `safe_include'
[...]

PR #4211, opened 2016-07-22
Add respond_to? to controller_helpers

zephiransas:
As @viamin says, In some case setup method isn't available.
So, I add respond_to? when call setup method.

But I don't know how to write test case. If you could point right direction, i would very pleased.

fixes plataformatec/devise#4189

### DIFF
--- a/lib/devise/test/controller_helpers.rb
+++ b/lib/devise/test/controller_helpers.rb
@@ -25,7 +25,7 @@ module Devise
       extend ActiveSupport::Concern
 
       included do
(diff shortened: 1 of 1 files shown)


lucasmazza:
The `setup` method is available when including the test helpers on `ActiveSupport::TestCase` subclasses or on `rspec-rails` example groups. As I mentioned on https://github.com/plataformatec/devise/issues/4189#issuecomment-234617970, if you are including the helper modules on a intermediate module you should ensure that the Devise modules are included on the test class directly and not on the intermediate module.

PR #4221, opened 2016-07-26
Feature/error message

thib44:
Add the error message in case of registration fail (for example with custom validation)


lucasmazza:
@thib44 thanks for the Pull Request, but this sounds like something to be done on app application basis and not by default, as Devise already uses the responders format for flash messages.

Issue #4209, opened 2016-07-21
Add deliver_later option in send_devise_notification


HexterCH:
Hi, all - 

Sometimes we use background jobs to deal with the mailer, but in `send_devise_notification` method did not provide this option. I think is a good idea to add `deliver_later` option in this method can make more convenience.

I'd be happy to provide a PR for this issue. 

Waiting for feedback now. Thanks

PR #4224, opened 2016-07-28
[WIP] Issue #4209 - Add deliver_later option in send_devise_notification

HexterCH:
#4209

Add a class variable `deliver_later_option` and modify the method `send_devise_notification` 

Now you can just add `config.deliver_later_option = true` in your `config/initailizers/devise.rb`, and devise mailer will use `deliver_later` to send every letter.

Hope to get some feedback. 😄

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -287,8 +287,8 @@ module Devise
   @@token_generator = nil
 
   # Store devise notification delivery method, default is deliver now
-  mattr_accessor :delivery_method
-  @@delivery_method = :deliver_now
+  mattr_accessor :deliver_later_option
+  @@deliver_later_option = false
 
   # Default way to set up Devise. Run rails generate devise_install to create
   # a fresh initializer with all configuration values.
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -286,6 +286,10 @@ module Devise
   mattr_accessor :token_generator
   @@token_generator = nil
 
+  # Store devise notification delivery method, default is deliver now
+  mattr_accessor :delivery_method
+  @@delivery_method = :deliver_now
+
   # Default way to set up Devise. Run rails generate devise_install to create
   # a fresh initializer with all configuration values.
   def self.setup
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -129,6 +129,10 @@ module Devise
         Devise.mailer
       end
 
+      def delivery_method
+        Devise.delivery_method
+      end
+
       # This is an internal method called every time Devise needs
       # to send a notification/mail. This can be overridden if you
       # need to customize the e-mail delivery logic. For instance,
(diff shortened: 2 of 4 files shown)


lib/devise/models/authenticatable.rb
```diff
@@ -187,8 +191,9 @@ def devise_mailer
       def send_devise_notification(notification, *args)
         message = devise_mailer.send(notification, self, *args)
         # Remove once we move to Rails 4.2+ only.
+
         if message.respond_to?(:deliver_now)
```
lucasmazza: I think we can remove this `if` now, right? 😄

lucasmazza:
@HexterCH thanks for tackling this, can you please add a small test case that asserts that the email is enqueued when the config is on?

lib/devise.rb
```diff
@@ -286,6 +286,10 @@ module Test
   mattr_accessor :token_generator
   @@token_generator = nil
 
+  # Store devise notification delivery method, default is deliver now
+  mattr_accessor :deliver_later_option
```
lucasmazza: We should add a new commenting section to the initializer template to document this for new apps.

HexterCH:
Sure, I am working on it.

lib/devise/models/authenticatable.rb
lucasmazza: @HexterCH what do you think of this? I guess that's the only pending piece we have here.

HexterCH: @lucasmazza , but if remove this `if`, that meanings if user's environment don't support `deliver_now` or `deliver_later` will failed.

And really sorry about that I respond so lately. 😭

HexterCH:
I am stuck on this test, any relative test that I can reference ? 🆘

brunowego:
@HexterCH exactly what I want! Waiting anxiously.

tegon:
@HexterCH Thanks for your contribution, we appreciate it!
Sorry for the late response, but I think I can help you with that failing spec.

Your test failed in builds for rails 4.2 and `mongoid`, with the error `ActiveJob::SerializationError: Unsupported argument type: User`.
I did some debugging here, and the reason that it failed is: on rails 4.2, `ActiveJob` requires that any model passed to it should include the module `GlobalID::Identification`. Your build doesn't fail with `active_record` because the module is included by default.
Some references: 
- http://edgeguides.rubyonrails.org/active_job_basics.html#globalid
- https://github.com/rails/rails/b lob/6375c9a3ae34cadd0d98e31ae1474fee26cf92ea/guides/source/4_2_release_notes.md#active-job 
- https://stackoverflow.com/questions/35554457/how-to-override-activejob-to-prevent-serialization-error

We can solve this issue by including the module in our `User` test class for `mongoid`:
```diff
# test/rails_app/app/mongoid/user.rb
require 'shared_user'

class User
  include Mongoid::Document
  include Shim
  include SharedUser
+ # Required for ActiveJob serialization. Reference: http://edgeguides.rubyonrails.org/active_job_basics.html#globalid
+ include GlobalID::Identification

tegon:
I'm closing this since it has been here for a long time and probably will be too hard to rebase it.
If you find that this can still be relevant, feel free to open a new pull request with a similar implementation.

Thank you!

seanpdoyle:
@HexterCH @tegon I've re-opened this unit of work as part of https://github.com/heartcombo/devise/pull/5610.

PR #4234, opened 2016-08-03
ActiveSupport deprecation warning fix on calling #ago on a Fixnum

ncancelliere:
In `Timeoutable` there is a call to #ago on a Fixnum class.  `timeout_in` returns a Fixnum from (3.days, 15.minutes, etc. as suggested values from Devise documentation / wiki) -- and so when #ago is called you get a deprecation warning from ActiveSupport that calling #ago on a number is not going to be supported.

```
DEPRECATION WARNING: Calling #ago or #until on a number (e.g. 5.ago) is deprecated and will be removed in the future, use 5.seconds.ago instead.`
```

I updated the code per ActiveSupport's recommendation for the future.  It also works with existing ActiveSupport libraries.  Specs still run green.

### DIFF
--- a/lib/devise/models/timeoutable.rb
+++ b/lib/devise/models/timeoutable.rb
@@ -26,7 +26,7 @@ module Devise
 
       # Checks whether the user session has expired based on configured time.
       def timedout?(last_access)
-        !timeout_in.nil? && last_access && last_access <= timeout_in.ago
+        !timeout_in.nil? && last_access && last_access <= timeout_in.seconds.ago
       end
 
       def timeout_in


lucasmazza:
Duplicate from #3899 - we don't use numeric values on the timeout configurations on Devise, and app developers shouldn't use it too - this is something that should be upgraded on each app that uses the deprecated API.

ncancelliere:
That isn't what I see happening.  In my User model I have timeout_in defined so I can dynamically change it.

```
  # app/model/users.rb

  ##
  # Devise timeout (dynamically generated so it can have different lengths)
  #
  def timeout_in
    is_admin? ? 1.days : super
  end
```

It's when you do this that it complains with the deprecation warning unless you either monkey patch Numeric with #ago or do what I did in my PR to fix timeoutable.

lucasmazza:
@ncancelliere does your app sets the `timeout_in` option on your Devise initializer? Which version of Devise are you using with your app? Devise 4.2 has a default of `30.minutes`, so pristine apps shouldn't hit this deprecation by default.

ncancelliere:
Using devise 4.2.0 with the devise-token_authenticatable gem 0.5.2 on Rails 4.1.

I actually figured out what was going on -- the `1.days` value is actually a SettingsLogic gem value (I replaced it in my example so not to have to explain all that) -- but it turns out that when you save the ActiveSupport duration (1.days) it returns a Fixnum (I thought it'd work just like the devise yml)... so I updated my method to do `.seconds` to that to make it happy.  I started to investigate why the Devise config.yml worked fine but the SettingsLogic didn't -- it might be because of how they parse it or cast the values ... not sure exactly but I at least understand better what's going on.  

Sorry to bug you. :-)

PR #4250, opened 2016-08-11
Fix a minor typo in specs

marcgg:
Hello !

There is nothing much to this one, it's in the test's description so it shouldn't impact a thing and will make this line more greppable.


nashby:

PR #4274, opened 2016-08-26
Fix omniauth provider generated URL in links partial

caristomenopoulos:
The omniauth_authorize_path(resource_name, provider) does not work any more, and a correct URL is eg. for facebook:  user_facebook_omniauth_authorize_path


lucasmazza:
@caristomenopoulos that helper method still exists at [`Devise::OmniAuth:: UrlHelpers`](https://github.com/plataformatec/devise/blob/d293e00ef5f431129108c1cbebe942b32e6ba616/lib/devise/omniauth/url_helpers.rb#L4-L7) due to this partial, and it seems to be working as expected.

 In which scenario the helper isn't working for you?

PR #4280, opened 2016-09-01
[4279] Don't allow changing of password to same password

msumit:


msumit:
Not sure why 2 specs are failing..

msumit:
Dear Maintainers, 

Can we merge this PR?

Thanks,
Sumit

23ranjan:
+1

rubyrider:
Opinion: 
I believe this sounds kind of unsecured when we say "Don't  allow to update with the same password", Do we need to enable this service directly? 

In case we need this feature can we make it more configurable  and make it deactivated by default?  I believe not everyone's need is same!

georgeguimaraes:
I'm sorry but we are not accepting new behaviour into devise.

msumit:
Not sure how its unsecured, when user already knows the correct password.. Idea behind this PR that whoever uses password archiving/expiry thing on top of devise, they have to add the same logic to enforce not same password, cause ppl just love to use same password..

adaam2:
+1 for @msumit 

This seems like a common sense feature @georgeguimaraes

PR #4284, opened 2016-09-04
Use `strip_heredoc` for deprecation warnings

znz:
I think these deprecation warning messages should use `strip_heredoc` like others.


lucasmazza:
Thanks! :)

PR #4302, opened 2016-09-20
Change the way to detect ActiveRecord vs Mongoid

fjg:
Cause of **mongoid-paperclip** declaring _after_commit_ callback while **mongoid** (and MongoDB) does not support it.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -44,7 +44,7 @@ module Devise
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
         after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
-        if respond_to?(:after_commit) # ActiveRecord
+        if defined?(ActiveRecord) # ActiveRecord
           after_commit :send_on_create_confirmation_instructions, on: :create, if: :send_confirmation_notification?
           after_commit :send_reconfirmation_instructions, on: :update, if: :reconfirmation_required?
         else # Mongoid
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -44,7 +44,7 @@ module Devise
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
         after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
-        if defined?(ActiveRecord) # ActiveRecord
+        if defined?(ActiveRecord) && self.is_a?(ActiveRecord::Base) # ActiveRecord
           after_commit :send_on_create_confirmation_instructions, on: :create, if: :send_confirmation_notification?
           after_commit :send_reconfirmation_instructions, on: :update, if: :reconfirmation_required?
         else # Mongoid


zachweed:
I'm not a humongous fan of the way that we are doing the check. It's a weirder way of doing it but you can do a conditional check on whether or not the model is using `ActiveRecord` by running:

```
self.class < ActiveRecord::Base
=> true
```
I think it's important to do something more along these lines because this tells us, directly, whether or not the class we are included in inherits, at any point in the hierarchy, from `ActiveRecord::Base`.

tegon:
I agree with @zachweed. By checking if `ActiveRecord` is defined, the issue would still happen in projects that use both Mongoid and Active Record.
@fjg Would you mind changing your code to something like the following?
```rb
if self.is_a?(ActiveRecord::Base)

**Comment by fjg** [comment]
Just did, let me know if it's ok for you

**Comment by tegon** [comment]
@fjg It looks good!
But a lot of commits that aren't from your branched popped up here now. 🤔 
Can you make a cherry pick of your commits (https://github.com/plataformatec/devise/pull/4302/commits/1eed4c4457b326650aa215b1df49d02b9c2e25bf and https://github.com/plataformatec/devise/pull/4302/commits/acdd35698b91d2daa0da2b5fb1d53962870192b4)?
This is important in order to keep our commit history clean.

**Comment by fjg** [comment]
Sorry about that. Should be fine now.

**Comment by tegon** [comment]
No problem, thanks for fixing it so quickly 😄

**Comment by tegon** [comment]
Fixes #4160

**Comment by dark-panda** [comment]
This change doesn't appear to be working as expected, because the change for 

```ruby
self.is_a?(ActiveRecord::Base)
```

isn't actually checking to see if `self` is a subclass `ActiveRecord::Base`, it's checking to see if `self` is an instance of `ActiveRecord::Base`, which it isn't, but rather it's an instance of `Class`. The original suggestion of `self.class < ActiveRecord::Base` isn't quite correct either, as that would be checking to see if `Class` is a subclass of `ActiveRecord::Base`. The correct code would be

```ruby
self < ActiveRecord::Base
```

Alternatively, you can do `self.ancestors.include?(ActiveRecord::Base)` if you want it to be a bit more readable, although it is quite a bit slower to execute (on a quick benchmark on my system, `ancestors.include?` is about 5 times slower than using `Class#<`) but since this is presumably done during app initialization for the most part, it might be worth the extra readability. 

This appears to have slipped by the test suite, as the code path itself for the `ActiveRecord::Base` branch of the `if` statement is never hit by the tests, since `Class` can never be an `ActiveRecord::Base`, but it came up during an application upgrade I was performing to Ruby 2.5.0 that required an upgrade to Devise 4.4.0 for the Ruby syntax changes. I don't think it's possible to hit this code in an application, unless I'm really missing something here.

tegon:
@dark-panda you're right, thanks for noticing this.
I think the testes didn't catch this because either way we're defining callbacks to send the emails, so the features didn't stop working, but it's indeed a regression. I'm going to open a pull request to fix this, and also see if I can write new tests to guarantee this behavior.

Issue #3305, opened 2014-11-10
Add config option for length of token


bikezilla:
Hi!
In the [```friendly_token```](https://github.com/plataformatec/devise/blob/8e5c098e3aa53d6a72faadd8b929abaf2bb6d496/lib/devise.rb#L478) method definition I see a hard coded 15 byte token length option which results in 20 chars raw tokens.

My client's use case is that many of his users use e-mail on a mobile device only, so they might need to enter their tokens in the desktop/laptop browser (e.g. confirmation tokens) by hand, so he wants shorter tokens. 

I achieved that by overriding the ```friendly_token``` method, but thought it might make sense to contribute to the project by adding a new config option for the token length.

1. Is this a good idea, would it be accepted by the project maintainers? If yes I'm happy to sign myself up for this.
2. Is ```token_size``` a good name for the option (with a proper comment that it is actually bites, not chars)? Suggestions for better?

Thanks for your feedback!

josevalim:
You can simply monkey patch friendly_token. :)

PR #4331, opened 2016-10-13
[WIP] Improve friendly token

sot528:
Hi.
Can I add new config for changing token length?
I need it for strict rule of my project. 😂 

I know, this question is already exists.
Like this: https://github.com/plataformatec/devise/issues/3305

But normally, I don't want add monkey patch.
I think, it's not good for maintenance and operation of the future.

And also, parameter of `Devise#friendly_token` is looks characters count.
So I fixed it incidentally.

If you can accept, I will improve this PR.
Thanks. 😃


test/devise_test.rb
```diff
@@ -104,4 +104,22 @@ class DeviseTest < ActiveSupport::TestCase
       assert_no_match Devise.email_regexp, email
     end
   end
+
+  test '#friendly_token should return token of specific length' do
```
zachweed: nitpick, but this should be `.friendly_token..` or since it is a class method.

tegon:
Hi @sot528, thanks for your pull request.
I think this is a complexity we don't need in Devise, so you should probably continue with the monkey patch - even though I agree it's not ideal. 
Devise provides a default that works for most people, so I think it's better to keep as it is for now.

Thank you!

sot528:
@tegon 
OK, thank you for your opinion. ;)

PR #4343, opened 2016-10-27
set minimum password length after user registration update

ethirajsrinivasan:


lucasmazza:
Thanks @ethirajsrinivasan 😄

PR #4349, opened 2016-11-07
Refactor added record to warden

GuyAdar:


carlosantoniodasilva:
@GuyAdar thanks for your contribution. It looks like this got extra old commits which shouldn't be part of the change. I'm also not sure what's the reasoning behind the change, so for now I'm giving it a close.

Please feel free to update the commits by rebasing against master, and update the PR description with more information about the change, the reasoning, examples, etc. We can always reopen. Thanks!

PR #4351, opened 2016-11-08
Move failed_attempts increment into its own function

mobilutz:
With this change one can overwrite when the incrementation is handled without duplicating the other `valid_for_authentication` logic.

While working on an application of mine, I needed to increment the failed_attempts only on specific cases. Without this function I would need to duplicate all the logic inside function to just change the incrementation logic.

As this is just a refactor there are no new specs.

### DIFF
--- a/lib/devise/models/lockable.rb
+++ b/lib/devise/models/lockable.rb
@@ -99,8 +99,7 @@ module Devise
         if super && !access_locked?
           true
         else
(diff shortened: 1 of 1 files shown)


mobilutz:
Any info on when this could be part of devise? Thanks in advance.

feliperenan:
Should we rebase it with the master branch since it's been a long while?

PR #4355, opened 2016-11-14
Prevent reset password token leakage

akayvanfar:
The current implementation of the password reset page using a reset token leaves the reset token in the url's query parameters, which run the risk of exposing it via the Referer. This solution, grabs the reset token, stores it in the session, and redirects the user to the reset password page (/edit) without the reset token in the query parameters.


ethirajsrinivasan:
@akayvanfar specs are appreciated

akayvanfar:
@ethirajsrinivasan thanks for the quick feedback. I'll add some and re-submit.

PR #4375, opened 2016-12-10
Replace log_process_action to append_info_to_payload

fbbergamo:
This implementation replace `log_process_action` to `append_info_to_payload`  that way the status should be present in ActiveSupport::Notifications's payload and at the log

It's based on what AR is doing:
https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/activerecord/lib/active_record/railties/controller_runtime.rb#L35

No test change was needed but one test was include.

### DIFF
--- a/test/integration/authenticatable_test.rb
+++ b/test/integration/authenticatable_test.rb
@@ -543,6 +543,18 @@ class AuthenticationOthersTest < Devise::IntegrationTest
       refute warden.authenticated?(:user)
     end
   end
+
+  test 'not signed in should returns notification payload with 401 status' do
+    begin
+      subscriber = ActiveSupport::Notifications.subscribe /process_action.action_controller/ do |_name, _start, _finish, _id, payload|
+        assert_equal 401, payload[:status]
+      end
+
+      get admins_path
+    ensure
+      ActiveSupport::Notifications.unsubscribe(subscriber)
+    end
+  end
 end
 
 class AuthenticationKeysTest < Devise::IntegrationTest
--- a/test/integration/authenticatable_test.rb
+++ b/test/integration/authenticatable_test.rb
@@ -544,7 +544,7 @@ class AuthenticationOthersTest < Devise::IntegrationTest
     end
   end
 
-  test 'not signed in should returns notification payload with 401 status' do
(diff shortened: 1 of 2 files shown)


benlovell:
Is there any appetite for landing this change? Happy to help get it across the line if anything outstanding.

test/integration/authenticatable_test.rb
```diff
@@ -543,6 +543,18 @@ class AuthenticationOthersTest < Devise::IntegrationTest
       refute warden.authenticated?(:user)
     end
   end
+
+  test 'not signed in should returns notification payload with 401 status' do
```
pjg: Typo, `returns` => `return`

jonpaul:
Would love to see this re-reviewed.

myriky:
I hope this PR will be merge.

fbbergamo:
@rafaelfranca @lucasmazza 
Is there anything that I may help to merge this PR?

rafaelsales:
Seems like this actually broke my logging.

I noticed that logs in my app started reporting `Completed 401 Unauthorized` when the action failed with an unhandled exception. The client still receives 500 Internal Error, but the log reports 401.

I'm checking how to fix it

tegon:
@rafaelsales I not sure it's related, we released this on v4.4.0

rafaelsales:
@tegon sorry, I had a typo in the version numbers of my comment. Just fixed.

If I revert this particular commit, the previous behavior is established. I'm building a test case to reproduce.

rafaelsales:
@tegon Here's the test case to reproduce - this commit was made on top of the latest master
https://github.com/rafaelsales/devise/commit/302c53e9a8a2b6aad5e369605891740b5c72dc7e

I tried to dig in a little bit and I found out that in the following snippet of this PR, the `payload[:exception]` always returns `nil` even if a public action raises an error, so the `401` status is added to the payload.

```ruby
        def append_info_to_payload(payload)
          super
          payload[:status] ||= 401 unless payload[:exception]
        end
```

stanhu:
I did some investigation on this. Cross-linking my thoughts in the Lograge issue tracker: https://github.com/roidrage/lograge/issues/67#issuecomment-601604344

PR #4376, opened 2016-12-12
display flash messages in new session view

tpalmer:
This seems like good default behavior to have.


tegon:
Hi @tpalmer, thanks for your pull request.
Where to display flash messages is specific to each application. By adding it here, we can cause some confusion, for example, the messages can appear duplicated in my application's view after installing the gem. In order to change this behavior, I'd have to override the sessions view.
That's why I'm going to close this, I hope you understand my point.

Thank you!

PR #4381, opened 2016-12-19
remove warnings about default value with new thor

slbug:
Removes warning "Expected boolean default value for '--markerb'; got :erb (string)" for thor >= 0.19.2

### DIFF
--- a/lib/generators/devise/views_generator.rb
+++ b/lib/generators/devise/views_generator.rb
@@ -139,7 +139,11 @@ module Devise
                               default: defined?(SimpleForm) ? "simple_form_for" : "form_for"
 
       hook_for :markerb,  desc: "Generate markerb instead of erb mail views",
-                          default: defined?(Markerb) ? :markerb : :erb,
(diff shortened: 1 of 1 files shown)


faucct:
Can this be reviewed, please?

faucct:
Great, can this get a version bump, please?

faucct:
Ping.

rafaelfranca:
@faucct hey! I don't think we need to rush a release because of a warning. If you don't want  the warning you can use the master branch.

faucct:
These warnings are very annoying in Rubymine when working with our project which has 20+ gems: each time you change the branch the Rubymine starts recalculating rake tasks and switching focus to those warnings.
I could use the master branch, but again our project has 20 gemfiles in it and it would be much easier to just bump the version instead of adding the github reference to each Gemfile. I think that you could release a pre version.

faucct:
ping

faucct:
ping

PR #4382, opened 2016-12-21
update devise.rb to encourage use of the Rails 4.1 secrets convention

BenMorganIO:
Quite often, when I setup a new Rails application, the second and third keys that enter the secrets.yml (just after the secret_key_base) is the `devise_secret_key` and the `devise_pepper_key`. Whenever I come onto older projects, I notice that some people haven't yet switched over to using the secrets convention that has been around since Rails 4.1. Secrets will also be getting a boost in Rails 5.1 as encryption is introduced.

I've also updated the docs. Added a missing "for" and an Oxford comma. Switch "can" to "may".

If a user doesn't put in a secret or a pepper, we could default it in the app to load up the `secrets.devise_*_key` instead. This would save some people some configuration time.

### DIFF
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -2,11 +2,11 @@
 # Many of these configuration options can be set straight in your model.
 Devise.setup do |config|
   # The secret key used by Devise. Devise uses this key to generate
-  # random tokens. Changing this key will render invalid all existing
-  # confirmation, reset password and unlock tokens in the database.
-  # Devise will use the `secret_key_base` as its `secret_key`
-  # by default. You can change it below and use your own secret key.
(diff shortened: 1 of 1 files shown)


zachweed:
IMO there should definitely be a fallback to a default, given `devise_pepper_key` is not present.

tegon:
Hello @BenMorganIO, thanks for your pull request.

I guess it makes sense to change it to [`credentials`](https://www.engineyard.com/blog/rails-encrypted-credentials-on-rails-5.2) now, what do you think?
Also, I think it would be nice to add in the comments in which file they should be included and a example of how to do it.

tegon:

Thank you!

BenMorganIO:
Yep, I think it would make sense to move to credentials.

PR #4386, opened 2016-12-24
Fix warnings: `Use assert_nil if expecting nil`

koic:
Fix a following deprecated warnings.

"`Use assert_nil if expecting nil`"

Thanks.


carlosantoniodasilva:
Hey @koic, I'm sorry I missed your PR before! I merged #4445 which contained most of the changes for `assert_nil`, and added c76f0e59b7c3e36cefd313a35bbf9812201d64e5 which fixed the latest case. (all were covered here on your PR, again sorry I missed it!)

We're now green on master, and with no minitest warnings. Thanks for your contribution!

koic:
All right. Thanks to your kind explanation, I understood that well.

Thanks for your maintenance!

PR #4391, opened 2017-01-03
Clarify how store_location_for modifies URIs

olivierlacan:
This fairly straightforward refactor extracts the most obtuse portions of `store_location_for` into the following discrete private methods:
- `extract_path_from_location`
- `remove_domain_from_uri`
- `add_fragment_back_to_path`

This may seem like indirection but it was very unclear what operation was being performed on the location sent to store_location_for prior to this change.

I considered documenting the behavior but the code seemed like a prime candidate for method extraction.

### DIFF
--- a/lib/devise/controllers/store_location.rb
+++ b/lib/devise/controllers/store_location.rb
@@ -33,12 +33,9 @@ module Devise
       #
       def store_location_for(resource_or_scope, location)
         session_key = stored_location_key_for(resource_or_scope)
-        uri = parse_uri(location)
-        if uri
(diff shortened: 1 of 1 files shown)


olivierlacan:
@lucasmazza Let me know if this makes sense. The only failures in the build seem unrelated. Also this is sort of staging area for a bugfix I'd like to introduce if it makes sense to you.

tegon:
@olivierlacan Thanks for your contribution! Sorry for taking so long to review it.
I just triggered a rebuild and everything is green again 💚

PR #4403, opened 2017-01-19
fix name of the module in the docs

bakongo:
Docs say `Devise::DatabaseAuthenticatable`. 
The correct name is `Devise::Models::DatabaseAuthenticatable`.


nashby:
thanks!

PR #4405, opened 2017-01-23
Validate password_confirmation when password is set, even if password_required? is false

dgmora:
Expected:
Even if i overwrite `password_required?` to return `false`, IF I add a password, the password should be validated against its confirmation
Current:
The validation of the confirmation is skipped.

Background: In some cases, you might want to allow to create new users without password, but still have the ability to set a password (say: allow to login regularly with a password, and via sms/magic link). Even if a password is not required, if you add one, it should be valid.

The tests I added check that if password and confirmation are `blank?` the validatable is `valid`. From my point of view that should be `invalid`, and be `valid` only if `password` and `confirmation_password` are `nil`, but that might change the current behaviour.

### DIFF
--- a/lib/devise/models/validatable.rb
+++ b/lib/devise/models/validatable.rb
@@ -31,7 +31,7 @@ module Devise
           validates_format_of     :email, with: email_regexp, allow_blank: true, if: :email_changed?
 
           validates_presence_of     :password, if: :password_required?
-          validates_confirmation_of :password, if: :password_required?
+          validates_confirmation_of :password
           validates_length_of       :password, within: password_length, allow_blank: true
         end
       end
--- a/test/models/validatable_test.rb
+++ b/test/models/validatable_test.rb
@@ -107,6 +107,18 @@ class ValidatableTest < ActiveSupport::TestCase
     assert_equal 'is too long (maximum is 72 characters)', user.errors[:password].join
   end
 
+  test 'should complain about confirmation even if password is not required' do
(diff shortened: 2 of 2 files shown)


tegon:
Hello @dgmora, thanks for your pull request.

Can you explain your use case for this? Are you overriding the default `#password_required?` method?

tegon:
I'm closing this PR because it has not had recent activity.
If you find that this can still be relevant, feel free to open a new issue with the reasons that motivated this change.

Thank you!

dgmora:
Hi!

The use case is explained in the description.

```
In some cases, you might want to allow to create new users without password, but still have the ability to set a password (say: allow to login regularly with a password, and via sms/magic link). Even if a password is not required, if you add one, it should be valid.
```

After a year I don't need this anymore. I think it'd makes sense that devise would work like that and I'd rather reopen and merge this..

Cheers

tegon:
I see, well, since we don't support those kinds of logins (SMS/magic link) out of the box, I don't think we should change only this method to allow a part of it.
If in the future we add support for them, we can make the required changes here too.

dgmora:
The SMS/magic link is just an example. What the PR fixes is (to me) how authentication should work in general, despite what devise supports or not.

```
Even if i overwrite password_required? to return false, IF I add a password, the password should be validated against its confirmation
```

But well, I guess it's more comfortable to keep it closed. ¯\_(ツ)_/¯

Issue #4413, opened 2017-01-30
in rails 5, g devise MODEL ignores primary_key_type: :uuid


jpwynn:
If Rails is configured to use :uuid as the primary key for all tables, eg if config/application.rb contains:

    config.generators do |g|
        # use uuid for all tables
      g.orm :active_record, primary_key_type: :uuid
    end 

`rails g devise MODEL` still creates a migration for the model with integer primary key, but probably ought to add the uuid column type for the id instead.

georgeguimaraes:
This should be a good addition to the devise's generator. I'd accept a PR.

filip373:
I've opened a pull request with the required changes. It seems to work fine, take a look at #4426

saroar:
hey its still does not work i try

PR #4422, opened 2017-02-04
Add primary key type to migration template

travelerspb:
#4413


georgeguimaraes:
You cannot pass a primary key like that to `create_table`.

travelerspb:
Any details? Because thats how it works in ActiveRecord https://github.com/rails/rails/blob/1e11df8ae65eb4455236af6c4f5765c631f09186/activerecord/lib/rails/generators/active_record/migration/templates/create_table_migration.rb

PR #4426, opened 2017-02-10
Primary key type in migration template

### LINKED ISSUE #4413: in rails 5, g devise MODEL ignores primary_key_type: :uuid

filip373:
Adds primary key type to migration creating template in Rails 5. It uses the configuration set in Rails generator as below:
```
config.generators do |g|
  g.orm :active_record, primary_key_type: :uuid
end
```
Resolves #4413

### DIFF
--- a/lib/generators/active_record/devise_generator.rb
+++ b/lib/generators/active_record/devise_generator.rb
@@ -6,6 +6,8 @@ module ActiveRecord
     class DeviseGenerator < ActiveRecord::Generators::Base
       argument :attributes, type: :array, default: [], banner: "field:type field:type"
 
+      class_option :primary_key_type, type: :string, desc: "The type for primary key"
+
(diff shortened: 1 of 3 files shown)


tegon:
Hello, @filip373, thanks for your contribution.
Your code looks good to me, could you please do a rebase with `master` to see if the build passes?

filip373:
Hi @tegon, thanks for your response.
I've rebased the code and applied a small patch so the build passes.

tegon:
Thanks, @filip373!

PR #4434, opened 2017-02-18
Added nested attributes example for README

ahmad19:
Hi, 

This PR includes an example for `README` section related to `nested attributes`. 

Hope this adds more clarity to the README section.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,21 @@ class ApplicationController < ActionController::Base
 end
 ```
 
-The above works for any additional fields where the parameters are simple scalar types. If you have nested attributes (say you're using `accepts_nested_attributes_for`), then you will need to tell devise about those nestings and types. Devise allows you to completely change Devise defaults or invoke custom behaviour by passing a block:
+The above works for any additional fields where the parameters are simple scalar types. If you have nested attributes (say you're using `accepts_nested_attributes_for`), then you will need to tell devise about those nestings and types:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -216,7 +216,21 @@ class ApplicationController < ActionController::Base
...
+The above works for any additional fields where the parameters are simple scalar types. If you have nested attributes (say you're using `accepts_nested_attributes_for`), then you will need to tell devise about those nestings and types:
+
+```ruby
+class ApplicationController < ActionController::Base
+  before_action :configure_permitted_parameters, if: :devise_controller?
+
+  protected
+
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:sign_up, keys: [:first_name, :last_name, address_attributes: [:country, :state, :city, :area, :postal_code]])
```
olleolleolle: Would the `keys:` value be clearer with some added `{` and `}` in places? (Perhaps you can rearrange the list, so that you "had to" add curly braces.)

ahmad19: I tried to put the curly braces but curly braces are causing syntax error. I think `Permit` method doesn't accept a hash.

tegon: You're right, for nested attributes we have to use an array of keys.

tegon:
Thanks, @ahmad19! Hope this helps others too.

PR #4444, opened 2017-02-26
Allow Rails 5.1

matthewrudy:
Open up the railties dependency and add 5.1.0.beta1 to travis


matthewrudy:
There are a lot of deprecation warnings like:

```
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead. (called from block in clear_reset_password_token? at /home/travis/build/plataformatec/devise/lib/devise/models/recoverable.rb:99)
```

Not sure how best to resolve this,
apart from adding checks everywhere. 

``` ruby
if respond_to(:saved_change_to_attribute?) 
   saved_change_to_attribute?(:encrypted_password)
else
  respond_to?(:encrypted_password_changed?) && encrypted_password_changed?
end
```

Doesn't feel so nice.

matthewrudy:
I've tried to fix the tests on master,
which should make these pass too.

https://github.com/plataformatec/devise/pull/4445

renchap:
Rails 5.1 will hit RC1 soon, it might be a good idea to merge this before.

betesh:
I put this branch in a Gemfile with Rails 5.2.0.alpha (i.e. master on github), and bundler downgraded responders to 0.6.5, which causes all sorts of errors because it was designed for Rails 3.x

So responders will also need to release a new gem, and devise should specify a minimum version of responders in its gemspec.

betesh:
responders is now compatible with rails 5.1 and master branch, but you'll need to pull it from github for now

jcoyne:
Rails 5.1 final is now out, can you update to point at 5.1.0 rather than beta?

jcoyne:
Updated this PR to use the final release in #4512

kevinelliott:
@rafaelfranca Would you please update us on the reason this was closed? Was this addressed in another PR? I don't see a PR merge here, so that would leave me to believe this was simply closed. Thank you!

PR #4451, opened 2017-03-02
Add Albanian translations to locale

fatosmorina:
Locale of Devise has been previously translated into Albanian language by 
https://github.com/ttomor

I have only made a few minor changes.

### DIFF
--- a/config/locales/sq.yml
+++ b/config/locales/sq.yml
@@ -1,0 +1,60 @@
+sq:
+  devise:
+    confirmations:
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
Thank you, but we do not include translations directly on the devise repo. There are, however, a couple of external solutions/links that you can check on the wiki: https://github.com/plataformatec/devise/wiki/I18n.

Hope that helps.

PR #3874, opened 2015-12-27
 Send notification to user's old email address on email change

ethirajsrinivasan:
email notification is sent to user's old  email address when user changes his/her email in the application  
- Issue #3862

PR #4455, opened 2017-03-06
Notify original email when it's changed

### LINKED ISSUE #3862: Send e-mail notification to user on email change
The latest version of devise (3.5.3) adds support for [notifying users of a password change](https://github.com/plataformatec/devise/pull/3642), which is a great addition, I've been doing this manually up until now.

Another thing that I am still doing manually, and is also considered best security practice, is notifying a users old email address when an accounts email has been changed. This has been proposed before (https://github.com/plataformatec/devise/issues/3691) and was marked as out of scope for Devise. Maybe with the latest change to (optionally) send a notification on password change, you might reconsider this feature request.

I'd be happy to send a PR if this is wanted.

carlosantoniodasilva:
Closes #3874, Closes #3862.

Introduces a new setting `send_email_change_notification`, which will automatically send a new `email_change` notification to the original email when a change happens.

When using the `reconfirmable` option of the `confirmable` module, the email notification is sent right away, upon email change request, instead of waiting for the unconfirmed email to be confirmed. This gives people a heads up of the change that's about to happen in cases like this.

Thanks @ethirajsrinivasan for the original implementation.

/cc @lucasmazza @rafaelfranca


lib/devise.rb
```diff
@@ -153,7 +153,11 @@ module Test
   mattr_accessor :pepper
   @@pepper = nil
 
-  # Used to enable sending notification to user when their password is changed
+  # Used to send notification to the original user email when their email is changed.
+  mattr_accessor :send_email_change_notification
```
lucasmazza: I wonder if using `email_changed` is more fitting than `email_change` to display that the even has already happened (even considering that we already use similar names around the codebase)

carlosantoniodasilva: Yeah it probably makes sense. I considered that but kept `email_change` for consistency with `password_change`. 

There's also the scenario with reconfirmable where the change hasn't actually happened yet, but I guess `email_changed` would work too.

If we go with `email_changed`, I'd vote for deprecating `password_change` in favor of `password_changed` - just not sure the work is worth it :smile:

lucasmazza: I'd vote for deprecating `password_change` too, but dunno if we need to do it right away or save for a future breaking release.

carlosantoniodasilva: It seems fine to deprecate now but leave it around for a while, only remove on a bigger bump. I might do that as a separate PR though. I'll rename this one to `email_changed`. Thanks!

StephenClem72:
Do I need to be on the latest version (4.2.1) to be able to use this feature? Is there a particular configuration that I need to have to do this?

salmagomaa:
The configuration needed is to send email notification on changing email is: 
`config.send_email_changed_notification = true` in  "devise.rb"

Issue #4397, opened 2017-01-17
Devise doesn't error when setting password by token without providing valid arguments


dinomite:
On an otherwise mostly-stock Devise setup we had a password recovery form that…didn't send up the password that the user input (though it did send up the reset_password_token).  Calling `PaswordsController.update` with invalid password & password confirmation (or lacking them) still logs in the user in and says that the password was successfully changed.

Insufficient data is passed by `PaswordsController.update` into `Recoverable.reset_password_by_token`, which calls `recoverable.reset_password(<attributes>)`.  `Recoverable.reset_password()` returns the result of calling `save` on the resource class: whether or not the resource was successfully saved.  Unfortunately, `Recoverable.reset_password_by_token` [ignores the result of this method call](https://github.com/plataformatec/devise/blob/v4.2.0/lib/devise/models/recoverable.rb#L136), continuing on to update the `resource`'s `reset_password_token` and itself return to `PasswordsController.update` which, seeing an empty `resource.errors`, signs in the user and sets a success flash message.

evertonts:
Can you please provide a sample application that reproduces the error?

dinomite:
Here's a repro repo: https://github.com/dinomite/devise-password-token-repro

Run it, [try to reset the password](http://localhost:3000/users/password/edit?reset_password_token=---fq9wEMk3-nHYg9R98) and see the result.

I removed the fields from the reset password form, but when you submit it, you are logged in, get a success flash from Devise, but the password hasn't been changed.

PR #4456, opened 2017-03-06
Fix absent password params from Password#update

victor-am:
Related to issue #4397 

This hotfix adds a string coercion to new_password paramenters when
trying to reset an user's password.

Before that, when a user submitted a password recovery form with the
new_password and new_password_confirmation params as nil, Devise would
sign in the user with a success notice but without actually changing the
password.

### DIFF
--- a/lib/devise/models/recoverable.rb
+++ b/lib/devise/models/recoverable.rb
@@ -33,10 +33,14 @@ module Devise
       # Update password saving the record and clearing token. Returns true if
       # the passwords are valid and the record was saved, false otherwise.
       def reset_password(new_password, new_password_confirmation)
-        self.password = new_password
-        self.password_confirmation = new_password_confirmation
-
-        save
+        if new_password.present?
+          self.password = new_password
+          self.password_confirmation = new_password_confirmation
+          save
+        else
+          errors.add(:password, :blank)
+          false
+        end
       end
 
       # Resets reset password token and send reset password instructions by email.
--- a/test/models/recoverable_test.rb
+++ b/test/models/recoverable_test.rb
@@ -184,6 +184,16 @@ class RecoverableTest < ActiveSupport::TestCase
     assert_equal raw, reset_password_user.reset_password_token
   end
 
+  test 'should return a new record with errors if password is not provided' do
+    user = create_user
+    raw  = user.send_reset_password_instructions
+
+    reset_password_user = User.reset_password_by_token(reset_password_token: raw)
+    refute reset_password_user.errors.empty?
+    assert_match "can't be blank", reset_password_user.errors[:password].join
+    assert_equal raw, reset_password_user.reset_password_token
+  end
+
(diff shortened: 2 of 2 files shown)


lib/devise/models/recoverable.rb
```diff
@@ -33,8 +33,8 @@ def self.required_fields(klass)
       # Update password saving the record and clearing token. Returns true if
       # the passwords are valid and the record was saved, false otherwise.
       def reset_password(new_password, new_password_confirmation)
-        self.password = new_password
-        self.password_confirmation = new_password_confirmation
+        self.password = new_password.to_s
```
lucasmazza: Instead of assigning an empty `, I think we should explicitly check for `nil` values on `new_password` and add an error to the object instead.

victor-am: Changed to add a blank error (maybe empty error would be better?) to the record if the new_password is nil.

I didn't define a validation for the new_password_confirmation, do you think we should?

lib/devise/models/recoverable.rb
```diff
@@ -33,6 +33,8 @@ def self.required_fields(klass)
       # Update password saving the record and clearing token. Returns true if
       # the passwords are valid and the record was saved, false otherwise.
       def reset_password(new_password, new_password_confirmation)
+        self.errors.add(:password, :blank) && return unless new_password
```
lucasmazza: Please replace this inline check with a full fledged `if`/ `else` around the existing code - there's a lot going on on a single line and it will be more explicit to have the both scenarios handled through the `if`/`else`.

lucasmazza: And please call `new_password.blank?` to handle empty Strings as well.

victor-am: About the `.blank?` the validations take care of it automatically (that's why the original `.to_s` worked). Should I still add it to make it more explicit?

lib/devise/models/recoverable.rb
```diff
@@ -33,10 +33,13 @@ def self.required_fields(klass)
...
-        self.password = new_password
-        self.password_confirmation = new_password_confirmation
-
-        save
+        if new_password.present?
+          self.password = new_password
+          self.password_confirmation = new_password_confirmation
+          save
+        else
+          errors.add(:password, :blank)
```
carlosantoniodasilva: @victor-am it'd be good to return `false` here, to keep the return value of `reset_password` the same (people might be relying on the previous truthy/falsy return value as to whether it worked or not)

victor-am: Good point.
Just did the adjustment 👍

lucasmazza:
@victor-am great job! Now that we are done, please squash together the tree commits into one 😄

victor-am:
Done 👍

carlosantoniodasilva:
Merged, thanks @victor-am!

PR #4458, opened 2017-03-06
4454 reset inactive user passwords

### LINKED ISSUE #4454: Recoverable allows resetting passwords for inactive users
Because [`Recoverable.send_reset_password_instructions` ignores `resource#active_for_authentication?`](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/models/recoverable.rb#L118-L119), it's possible for users to reset their passwords even though they're marked as inactive.

While this in itself won't grant access, it may be combined with other exploits to compromise authentication.

ivanoblomov:
Closes #4454.

Note this also fixes a bug in `PasswordsController` that signed-in users after a password reset even when the unlock strategy was `:none` or `:time`.


tegon:
I'm curious about which response the user is going to get when he/she tries to reset a password while inactive. What's going to happen?

tegon:
I've tested here locally and it seems like the user receives a message saying that they're going to receive a mail with the password reset instructions - although the mail never arrives because their account is inactive.
inactive-passwords

I think we should send a message explaining that they cannot reset their password until they email is confirmed. We should also handle the paranoid mode, with a more generic message.

tegon:
@ivanoblomov are you still going to work on this?

ivanoblomov:
@tegon sure, I'll take it on but, with the holidays, I probably won't have anything for you until January. That work?

tegon:
@ivanoblomov sure thing, no need to rush. I just wanted to know if you're still willing to work on it. Thanks.

ivanoblomov:
@tegon, this is ready for your review. Let me know if I missed anything.

ivanoblomov:
Not sure what the unrelated [bundle failures](https://travis-ci.org/plataformatec/devise/jobs/478038709) are about.

tegon:
@ivanoblomov 

> inactive—and not a locked—account. "Inactive" and "locked" meaning, for example, User#is_active is false and #locked_at is not nil.

I don't get it. What do you mean by `User#is_active`? There's no such method in Devise.
What I meant is that for applications that include the lockable module, the `active_for_authentication?` does [check the `locked_at` column](https://github.com/plataformatec/devise/blob/master/lib/devise/models/lockable.rb#L81).

So an inactive account can be either one with unconfirmed email or one with access locked.

ivanoblomov:
My bad, `#is_active` was our code's implementation detail! What I was trying to say was that the concepts of *active* and *locked* are, and should be, separate.

As you point out, Devise checks `locked_at`, but it also leaves how to define "active" entirely up to you. And the [docs say as much](https://github.com/plataformatec/devise/blob/354df3bc65fb211f0704208f52bf2213758585cf/lib/devise/models/authenticatable.rb#L43-L47).

I think part of our misunderstanding stems from the fact that Devise defines `active_for_authentication?` to be true, by default. So Devise ignoring `active_for_authentication?` when deciding whether to reset a password produces no unexpected results.

It's only when the client code defines the active concept (by [defining the `User#active_for_authentication?` callback](https://stackoverflow.com/questions/6004216/devise-how-do-i-forbid-certain-users-from-signing-in/6004353#6004353), as recommended) that something unexpected happens: namely, an inactive account gets its password reset. And I don't think this should be the case.

tegon:
> What I was trying to say was that the concepts of active and locked are, and should be, separate.

That's my point: they are not today. If the proposal is to separe them, that's another discussion. 

> As you point out, Devise checks locked_at, but it also leaves how to define "active" entirely up to you. And the docs say as much.

That's true, but the doc also says: don't forget to call `super` - otherwise `confirmable` and `lockable` won't work. 
So as long as `super` is being called at some point the lockable check is going to be called. That's what the tests that fail today are showing us: a confirmed user (e.g. "active") with a locked account cannot reset their password anymore to recover the account.

ivanoblomov:
Let me clarify: the concepts of active and locked *become* separate if the client code adds this callback to the user resource:
```ruby
  # http://stackoverflow.com/a/6004353/1935918
  # Devise callback to assure authorization
  def active_for_authentication?
    super && is_active?
  end
```
Because of the `super` calls you point out, this `#active_for_authentication?` gets called by the relevant modules and factored into the decision of whether the user resource is active for authentication or not.

This PR doesn't change any of this. It only makes sure that [Recoverable checks the same callback before emailing password-reset instructions](https://github.com/plataformatec/devise/pull/4458/files#diff-fa817bf486c518644016136465d914b2).

I think you may be right about my fixes to the regressions though. If you agree that the above fix is desired, I'd appreciate your help in working through them.

But if you think we don't need the fix, could you explain why it makes sense to send password-reset instructions to a user the client code defines as being inactive?

tegon:
> Because of the super calls you point out, this #active_for_authentication? gets called by the relevant modules and factored into the decision of whether the user resource is active for authentication or not.

Yep, that's what I meant: the lockable check **is going to be called**. So even if you define a method `active_for_authentication?` that returns `true`, if the account is locked it will return `false`. You can test this by putting some logs in the code:
```diff
diff --git a/lib/devise/models/authenticatable.rb b/lib/devise/models/authenticatable.rb
index d4d72e90..2751151c 100644
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -88,6 +88,7 @@ module Devise
       end
 
       def active_for_authentication?
+        p "Authenticatable#active_for_authentication?============================"
         true
       end
 
diff --git a/lib/devise/models/confirmable.rb b/lib/devise/models/confirmable.rb
index dbf6d0ff..298cf5c7 100644
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -142,6 +142,7 @@ module Devise
       # is already confirmed, it should never be blocked. Otherwise we need to
       # calculate if the confirm time has not expired for this user.
       def active_for_authentication?
+        p "Confirmable#active_for_authentication?============================"
         super && (!confirmation_required? || confirmed? || confirmation_period_valid?)
       end
 
diff --git a/lib/devise/models/lockable.rb b/lib/devise/models/lockable.rb
index b8ec4dca..dce30a13 100644
--- a/lib/devise/models/lockable.rb
+++ b/lib/devise/models/lockable.rb
@@ -79,6 +79,7 @@ module Devise
       # Overwrites active_for_authentication? from Devise::Models::Activatable for locking purposes
       # by verifying whether a user is active to sign in or not based on locked?
       def active_for_authentication?
+        p "Lockable#active_for_authentication?============================"
         super && !access_locked?
       end
 
--- a/test/rails_app/app/active_record/user.rb
+++ b/test/rails_app/app/active_record/user.rb
@@ -17,4 +17,14 @@ class User < ActiveRecord::Base
     # used to check in our test if the validations were called
     @@validations_performed = true
   end
+
+  def active_for_authentication?
+    p "User#active_for_authentication?============================"
+    super && is_active?
+  end
+
+  def is_active?
+    p "User#is_active?========================="
+    true
+  end
 end
```
Now if you remove the stub from the test I mentioned before
```diff
diff --git a/test/integration/recoverable_test.rb b/test/integration/recoverable_test.rb
index 38f60039..2b23f5f1 100644
--- a/test/integration/recoverable_test.rb
+++ b/test/integration/recoverable_test.rb
@@ -213,7 +213,6 @@ class PasswordTest < Devise::IntegrationTest
 
   test 'does not sign in user automatically after changing its password if it\'s locked and unlock strategy is :none or :time' do
     [:none, :time].each do |strategy|
-      User.any_instance.stubs(:active_for_authentication?).returns(true)
       swap Devise, unlock_strategy: strategy do
         create_user(locked: true)
         request_forgot_password
diff --git a/test/rails_app/app/active_record/user.rb b/test/rails_app/app/active_record/user.rb
index e065e261..539d5ee7 100644
```
You'll see the `active_for_authentication?` method call chain, which in my machine looks like this:
```ruby
"User#active_for_authentication?============================"
"Lockable#active_for_authentication?============================"
"Confirmable#active_for_authentication?============================"
"Authenticatable#active_for_authentication?============================"
"User#active_for_authentication?============================"
"Lockable#active_for_authentication?============================"
"Confirmable#active_for_authentication?============================"
"Authenticatable#active_for_authentication?============================"
```
So the test fails because the user is not active - in this case locked. 

> But if you think we don't need the fix, could you explain why it makes sense to send password-reset instructions to a user the client code defines as being inactive?

It makes sense in the context of the locked account - which it's broken now. 
I'm not saying it does makes sense in the other contexts - e.g. the account is confirmed, unlocked but the custom application definition, like `#is_active?` returns false - but in order to make this last case work we would have to actually separate the concepts or check all the possibilities, something like `resource.confirmed? && !resource.access_locked? && resource.active_for_authentication?`.
Of course the following doesn't work because we'd have to check whether those modules are included or not, since they're optional.

My point is that in the beginning we decided to do this because it could possibly cause exploits, but we don't now any of them right now. We would be doing a big change without any clear benefits, that's why I don't think it's worth the trouble.

PR #4461, opened 2017-03-08
Confirmation link validity is wrong in different time zone

jjuliano:
Due to time zone difference, the confirmation link in email is invalid, this can be solved by using UTC for Time.now for confirmation_period_expired?


carlosantoniodasilva:
Merged, thanks @jjuliano!

Issue #1777, opened 2012-04-04
Timeoutable timeouts causing Rails to flash[:timedout] as "true"


nbibler:
I know this has been mentioned [in](https://github.com/plataformatec/devise/issues/193) [other](https://github.com/plataformatec/devise/issues/219) tickets, but it appears to still be an issue, as of 2.0.4.  From the current, master's [failure_app.rb](https://github.com/plataformatec/devise/blob/3cbced4e4e644d18c8a5a7043d8fefde9829087a/lib/devise/failure_app.rb#L80):

``` ruby
def redirect_url
  if warden_message == :timeout
    flash[:timedout] = true
    attempted_path || scope_path
  else
    scope_path
  end
end
```

The third line there is causing the Rails `Flash::FlashHash` object instance to carry a flash[:timedout] whose value is of TrueClass.  For generalized flash implementations, such as the [one generated](https://github.com/plataformatec/devise/blob/3cbced4e4e644d18c8a5a7043d8fefde9829087a/test/rails_app/app/views/layouts/application.html.erb) and recommended by this gem:

``` erb
<%- flash.each do |name, msg| -%>
  <%= content_tag :div, msg, :id => "flash_#{name}" %>
<%- end -%>
```

The following HTML is being generated and displayed to visitors:

``` html
<div id="flash_timedout">true</div>
```

Obviously, this is non-ideal. :)  Is this `:timedout` flash message actually used for anything?  I see that the failure_app is [explicitly keeping](https://github.com/plataformatec/devise/blob/3cbced4e4e644d18c8a5a7043d8fefde9829087a/lib/devise/failure_app.rb#L57) the :timedout flash message, even though it's simply a `true` statement:

``` ruby
def redirect
  store_location!
  if flash[:timedout] && flash[:alert]
    flash.keep(:timedout)
    flash.keep(:alert)
  else
    flash[:alert] = i18n_message
  end
  redirect_to redirect_url
end
```

Based on a cursory reading of this code, it seems to me that a simple instance variable, which is set to true/false and nilified with each Rack `call` request, would provide the same functionality and work significantly better.

And, yes, I did see @josevalim's [comment on the use of Flash](https://github.com/plataformatec/devise/commit/275c480f89f814af7135093a507ab57397729faf#commitcomment-925730)... I just wonder if either the README and example application layout should be updated to reflect the check for String-iness or the surprising usage of Flash for non-user messages should be removed.

deivid-rodriguez:
[...]

PR #4464, opened 2017-03-10
Fix Timeoutable timeouts flash hash keys

janraasch:
See https://github.com/plataformatec/devise/issues/1777 for an in-depth discussion of the unexpected behaviour.

See https://github.com/plataformatec/devise/pull/4038 for the fix, but awaiting tests.

The behaviour still works as expected for redirects, see

```
test 'error message with i18n'
test 'error message with i18n with double redirect'
```

in https://github.com/plataformatec/devise/blob/master/test/integration/timeoutable_test.rb#L141-L166.

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -76,8 +76,7 @@ module Devise
     def redirect
       store_location!
       if is_flashing_format?
-        if flash[:timedout] && flash[:alert]
-          flash.keep(:timedout)
+        if flash[:alert]
           flash.keep(:alert)
         else
           flash[:alert] = i18n_message
@@ -113,7 +112,6 @@ module Devise
 
     def redirect_url
       if warden_message == :timeout
-        flash[:timedout] = true if is_flashing_format?
 
         path = if request.get?
           attempted_path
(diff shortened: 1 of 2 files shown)


tegon:
@janraasch Thanks for tackling this.

It seems really odd to me that the tests still pass after the code was deleted. It makes me wonder whether the tests were actually covering the intended scenarios in the first place.
I'll try to simulate the issue this code was trying to solve in the past, but I'm also worried about what @josevalim [said](https://github.com/plataformatec/devise/issues/1777#issuecomment-4960726) about the multi-thread/process environments. 

Did you have a chance to manually test this?

janraasch:
> Did you have a chance to manually test this?

Long time no see 😉. If I remember correctly we monkey-patched this on our app, but to be honest I no longer am affiliated with that application so I cannot look into the source code. So `\_o_/` I guess 🤓.

> It seems really odd to me that the tests still pass after the code was deleted. It makes me wonder whether the tests were actually covering the intended scenarios in the first place.

That does seem strange, I absolutely agree.

tegon:
Ok, get it. Thanks for replying.

I'm going to close this then, as I don't have the context of why this change was needed or what side effects this pull request might cause. I hope you understand my reasoning.

Thanks again!

PR #4491, opened 2017-04-06
Handle application/vnd.api+json in FailureApp

lcpriest:
I know it's an edge case - so am happy to use an override in my own app if this doesn't make sense in the gem.

While using https://github.com/cerebris/jsonapi-resources, the content-type is 'application/vnd.api+json' and this method was returning :api_json in place of :json.

This is only an issue when returning an error, as the [#http_auth_body](https://github.com/lcpriest/devise/blob/master/lib/devise/failure_app.rb#L190) method does not recognize this as needing a JSON response.

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -244,7 +244,11 @@ module Devise
     end
 
     def request_format
-      @request_format ||= request.format.try(:ref)
+      @request_format ||= begin
+        format = request.format.try(:ref)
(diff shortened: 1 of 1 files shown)


tegon:
Hello @lcpriest, thanks for your pull request.

I agree with you. This seems like an edge case, so it's better to just override it in your app. If this gets more popular, we can think about supporting this in Devise.

Thanks

PR #4506, opened 2017-04-26
added note about stopping spring

calvinclaus:
For me it was necessary to stop spring for the route helpers to be defined.


tegon:
Thanks @calvinclaus! ❤️ 
This can be useful to someone else.

Issue #4513, opened 2017-04-28
undefined method `rails51?' for Devise:Module


spaquet:
Given recent merge I gave a try to devise on a Rails 5.1 project by adding `gem 'devise', github: 'plataformatec/devise'` to the Gemfile.

But now, the following error pops up when running `rails s` command:

> ~/.gem/ruby/2.4.0/bundler/gems/devise-0fe308e7a77f/lib/devise/models/database_authenticatable.rb:140:in `<module:DatabaseAuthenticatable>': undefined method `rails51?' for Devise:Module (NoMethodError)

=== full error

> ~/.gem/ruby/2.4.0/bundler/gems/devise-0fe308e7a77f/lib/devise/models/database_authenticatable.rb:140:in `<module:DatabaseAuthenticatable>': undefined method `rails51?' for Devise:Module (NoMethodError)
> 	from ~/.gem/ruby/2.4.0/bundler/gems/devise-0fe308e7a77f/lib/devise/models/database_authenticatable.rb:25:in `<module:Models>'
> 	from ~/.gem/ruby/2.4.0/bundler/gems/devise-0fe308e7a77f/lib/devise/models/database_authenticatable.rb:4:in `<module:Devise>'
> 	from ~/.gem/ruby/2.4.0/bundler/gems/devise-0fe308e7a77f/lib/devise/models/database_authenticatable.rb:3:in `<top (required)>'
> 	from ~/.gem/ruby/2.4.0/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:292:in `require'
> 	from ~/.gem/ruby/2.4.0/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:292:in `block in require'
> 	from ~/.gem/ruby/2.4.0/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:256:in `block in load_dependency'
> 	from ~/.gem/ruby/2.4.0/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:661:in `new_constants_in'
> 	from ~/.gem/ruby/2.4.0/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:256:in `load_dependency'
> 	from ~/.gem/ruby/2.4.0/gems/activesupport-5.1.0/lib/active_support/dependencies.rb:292:in `require'
> 	from ~/.gem/ruby/2.4.0/bundler/gems/devise-0fe308e7a77f/lib/devise/models.rb:88:in `const_get'
[...]

PR #4514, opened 2017-04-29
Proposed fix for #4513

spaquet:
Proposed fix for issue #4513 by adding the following lines to 'lib/devise.rb'
```
def self.rails51?
      Rails.version.start_with? '5.1'
end
```


richpeck:
Looking forward to getting this merged

PR #4516, opened 2017-04-29
Separate Devise test methods from Devise

danschultzer:
~This PR does two things:~

1. I didn't like the way test specific methods are polluting the `Devise` module, so I've moved those to `Devise::Test`. This will prevent issues like #4514 from happening.

~2. Refactored the `rails51?` changes so we keep the codebase DRY.~

### DIFF
--- a/test/integration/rememberable_test.rb
+++ b/test/integration/rememberable_test.rb
@@ -10,9 +10,9 @@ class RememberMeTest < Devise::IntegrationTest
   end
 
   def generate_signed_cookie(raw_cookie)
-    request = if Devise.rails51?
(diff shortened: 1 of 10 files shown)


danschultzer:
@rafaelfranca wdyt?

rafaelfranca:
Thank you for the pull request, but I particularly prefer the previous code. The depilation is better in this case to make clear that we have code to support two different versions of the framework. It is also clear what to do when removing support for a version of the framework. It is more performant too since the conditionals happen on class loading time and not the runtime.

I like the first commit though.

danschultzer:
Ah alright. I've removed the second commit.

PR #4519, opened 2017-05-02
Move rails-controller-testing gem to test block

JamesWatling:


rafaelfranca:
Are you using the test group to either include or exclude? For what I know the test group is not used at all in gems, so I think we should just remove it.

rafaelfranca:
I mean the keyworkd `group :test do` (and move everything to the top level) and not what is inside it. Everything in that gemfile is being used for tests.

PR #4527, opened 2017-05-05
Consistent status code on confirmation error

vlado:
Before this fix if trying to confirm already confirmed account you will
get 200 status code for some formats (html and js) and 422 status code
for other formats (json, xml, ...).

At first I wanted to fix this in `responders` gem but after looking at source code and documentation https://github.com/plataformatec/responders/blob/master/lib/action_controller/responder.rb#L25 I saw that `respond_with` doesn't use same status code for `html` and other formats by design. So I just did it here for devise cause I think it should always return 422 status code if you try to confirm same account multiple times.


tegon:
Hello, @vlado thanks for your pull request.

Can you explain your use case for this?
Usually, we don't need to return different HTTP codes when we're rendering HTML.

vlado:
Hi @tegon, I think you are right. It was actually request from our security audit to change this status code so I just implemented it without thinking too much.

Closing this PR now.

PR #4529, opened 2017-05-08
Fix corner case when confirmation_sent_at is equal to 0.days.ago

stanhu:
If `Confirmable#confirmation_sent_at` is equal to `0.days.ago`, then `confirmation_period_valid?` will be deemed valid even if the setting is configured to disable this outright. To prevent this
error, we explicitly check the configuration setting to be `0.days.ago`.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -211,7 +211,10 @@ module Devise
         #   confirmation_period_valid?   # will always return true
         #
         def confirmation_period_valid?
-          self.class.allow_unconfirmed_access_for.nil? || (confirmation_sent_at && confirmation_sent_at.utc >= self.class.allow_unconfirmed_access_for.ago)
+          return true if self.class.allow_unconfirmed_access_for.nil?
+          return false if self.class.allow_unconfirmed_access_for == 0.days
+
+          confirmation_sent_at && confirmation_sent_at.utc >= self.class.allow_unconfirmed_access_for.ago
         end
 
         # Checks if the user confirmation happens before the token becomes invalid
--- a/test/models/confirmable_test.rb
+++ b/test/models/confirmable_test.rb
@@ -240,6 +240,16 @@ class ConfirmableTest < ActiveSupport::TestCase
     refute user.active_for_authentication?
   end
 
+  test 'should not be active when confirm period is set to 0 days' do
+    Devise.allow_unconfirmed_access_for = 0.days
+    user = create_user
+
+    Timecop.freeze(Time.zone.today) do
+      user.confirmation_sent_at = Time.zone.today
+      refute user.active_for_authentication?
+    end
+  end
+
   test 'should be active when we set allow_unconfirmed_access_for to nil' do
     swap Devise, allow_unconfirmed_access_for: nil do
       user = create_user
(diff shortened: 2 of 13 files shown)


stanhu:
Full context of the bug is here: https://gitlab.com/gitlab-org/gitlab-ee/issues/2362#note_29004709

stanhu:
@carlosantoniodasilva Would you mind reviewing? This bug took hours to uncover.

lucasmazza:
@stanhu can you please add a test case that cover this case? It will help to prevent any regressions regarding this scenario.

stanhu:
@lucasmazza Yes, done.

stanhu:
@lucasmazza Could you take a look at this?

test/models/confirmable_test.rb
```diff
@@ -225,6 +225,11 @@ def setup
     user = create_user
     user.confirmation_sent_at = Time.zone.today
     refute user.active_for_authentication?
+
+    Timecop.freeze(Time.zone.today) do
```
tegon: Can you create a new test block for these new assertions? It would be easier to read.

lib/devise/models/confirmable.rb
```diff
@@ -209,7 +209,10 @@ def confirmation_required?
         #   confirmation_period_valid?   # will always return true
         #
         def confirmation_period_valid?
-          self.class.allow_unconfirmed_access_for.nil? || (confirmation_sent_at && confirmation_sent_at.utc >= self.class.allow_unconfirmed_access_for.ago)
+          return false if self.class.allow_unconfirmed_access_for == 0.days
+          return true if self.class.allow_unconfirmed_access_for.nil?
+
+          (confirmation_sent_at && confirmation_sent_at.utc >= self.class.allow_unconfirmed_access_for.ago)
```
tegon: Since this was moved to a new line, the parenthesis can be removed.

lib/devise/models/confirmable.rb
```diff
@@ -209,7 +209,10 @@ def confirmation_required?
         #   confirmation_period_valid?   # will always return true
         #
         def confirmation_period_valid?
-          self.class.allow_unconfirmed_access_for.nil? || (confirmation_sent_at && confirmation_sent_at.utc >= self.class.allow_unconfirmed_access_for.ago)
+          return false if self.class.allow_unconfirmed_access_for == 0.days
+          return true if self.class.allow_unconfirmed_access_for.nil?
```
tegon: WDYT about moving the `nil` check to the top? I know it will work the same way, but it seems reasonable to return first when the value is `nil`, without the need to check whether is equal to `0.days` or not.

tegon:
@stanhu Thanks for your contribution.

Your code looks fine to me, I just wanted to ask: have you experienced this in production? I saw your issue in GitLab, and it seems like this would only happen when the `confirmation_sent_at` is set at the same second as the `#confirmation_period_valid?` method is called.

stanhu:
> Your code looks fine to me, I just wanted to ask: have you experienced this in production? I saw your issue in GitLab, and it seems like this would only happen when the confirmation_sent_at is set at the same second as the #confirmation_period_valid? method is called.

@tegon No, it only happened in our tests.

PR #4552, opened 2017-06-01
Generate a valid fixture by default

### LINKED ISSUE #4475: Default installation crashes tests
By following the instructions from the Readme on a rails 5.0 project created with `rails new myapp`, you end up with the following error:
```
Error:
HomeControllerTest#test_should_get_index:
ActiveRecord::RecordNotUnique: SQLite3::ConstraintException: UNIQUE constraint failed: admins.email: INSERT INTO "admins" ("created_at", "updated_at", "id") VALUES ('2017-03-23 16:00:12.355353', '2017-03-23 16:00:12.355353', 298486374)
```

This can be fixed by removing the admins.yml (or whatever you called your model), but it's not immediately obvious that that's how to fix it. Or at least it wasn't to me, given my limited experience with rails.

eldano:
This reduces the default fixture generated by devise creating only one instance of User (or whatever the name of the model is) instead of two, since that breaks the tests by default.
This fixes #4475

### DIFF
--- a/lib/generators/active_record/devise_generator.rb
+++ b/lib/generators/active_record/devise_generator.rb
@@ -18,7 +18,8 @@ module ActiveRecord
       end
 
       def generate_model
-        invoke "active_record:model", [name], migration: false unless model_exists? && behavior == :invoke
+        invoke "active_record:model", [name], migration: false, fixture: false unless model_exists? && behavior == :invoke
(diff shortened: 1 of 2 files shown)


tegon:
Hello, @eldano, thanks for your interesting in contributing to Devise!
Your code looks fine, but as I mentioned [here](https://github.com/plataformatec/devise/issues/4475#issuecomment-350014534) I don't think this is something we should add to the code. 
Sorry about that, I hope you understand my point.

Thank you again!

PR #4563, opened 2017-06-19
added #index action back into Sessions Controller.

nathantech2005:
I was receiving error #index action missing for Sessions Controller. The index action should be in the sessions controller because it's the action that loads when someone goes to load the sign in form.


rafaelfranca:
Thank you for the pull request. The action that loads when someone goes to load the sign in form in the `new` not `index`. See here https://github.com/plataformatec/devise/blob/master/lib/devise/rails/routes.rb#L49

nathantech2005:
I understand. But the #index is getting loaded when I trace the steps using
pry debugger.

rafaelfranca:
Is your application overriding the devise controller somehow?  Can you please provide a sample application that reproduces the error?

tegon:
`SessionsController` shouldn't have an `index` action. If you're getting an error about this, please open an issue so we can discuss what's happening.

Thanks.

PR #4565, opened 2017-06-20
Ensure string literals can be frozen.

pat:
These minor changes will allow Devise to work when MRI's string literals are frozen by default. This feature of Ruby is documented here: https://wyeworks.com/blog/2015/12/1/immutable-strings-in-ruby-2-dot-3

Will over a second patch in a moment for testing this to ensure regressions are avoided.

### DIFF
--- a/lib/generators/devise/devise_generator.rb
+++ b/lib/generators/devise/devise_generator.rb
@@ -8,7 +8,7 @@ module Devise
       namespace "devise"
       source_root File.expand_path("../templates", __FILE__)
 
-      desc "Generates a model with the given NAME (if one does not exist) with devise " <<
+      desc "Generates a model with the given NAME (if one does not exist) with devise " \
            "configuration plus a migration file and devise routes."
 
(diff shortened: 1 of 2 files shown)


lib/generators/devise/devise_generator.rb
```diff
@@ -8,15 +8,15 @@ class DeviseGenerator < Rails::Generators::NamedBase
...
+      desc "Generates a model with the given NAME (if one does not exist) with devise " \
            "configuration plus a migration file and devise routes."
 
       hook_for :orm
 
       class_option :routes, desc: "Generate routes", type: :boolean, default: true
 
       def add_devise_routes
-        devise_route  = "devise_for :#{plural_name}"
+        devise_route  = String.new "devise_for :#{plural_name}"
```
rafaelfranca: Can we use `dup` here?

pat: Done 👍

PR #4566, opened 2017-06-20
Test frozen string literals on Travis CI.

pat:
This PR adds a script `bin/literals` which will add the pragma setting `frozen_string_literal: true` to the top of all Ruby files in `app`, `lib`, and `test`. Errors in the test suite will then be raised if any string literals in these files are modified.

This script will only add the pragma comment if MRI 2.4 or newer is being used. Previous versions of Ruby either do not support this feature (MRI <= 2.2), or do not work with pragmater (MRI 2.3).

Related to #4565.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,9 @@ env:
 
 before_install: "rm ${BUNDLE_GEMFILE}.lock"
 
-before_script: "bundle update"
+before_script:
+  - "bundle update"
+  - ./bin/literals
 
 script: "bin/test"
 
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Thanks for the pull request. I'm more included to add the magic comments in all the files than adding a new step in the build. Mind to do it?

pat:
Sure, I can do that in a separate PR - though it's worth noting that this means anyone adding new files to the project will also need to add the magic comment as well. Having something like this step in the build is a more reliable process for catching any regressions.

Your project, your call, but just want to make sure the differences to the approaches are understood.

tegon:
Hi @pat, thanks for your pull request.
I agree with @rafaelfranca. We can ensure the new files have the magic comments during the code review.

I wanted to check with you: are you willing to send us a patch adding the comments?

pat:
Sure thing - have just submitted a new PR with the pragma comments. Closing this in favour of #4725.

Issue #4576, opened 2017-07-03
Typo in README


santiago-su:
On Getting Started you should migrate the database and it says
`Then run rails db:migrate`
It should say
`Then run rake db:migrate`

asnad:
@santiago-su In rails 5 you can run rails db:migrate. rake commands will be removed in future.

rafaelfranca:
Correct. It works in Rails 5 and we recommend you to use the newest version of the framework.

PR #4577, opened 2017-07-03
Typo in README #4576

santiago-su:


jasnow:
@santiago-su - "rails db:migrate" works on Rails 5 projects?
Reference: https://www.sitepoint.com/whats-new-rails-5

santiago-su:
@jasnow didn't know that, I should have checked better  before creating a pull request, maybe it should specify `rake db:migrate` on Rails < 5 and `rails db:migrate` for Rails 5? It's nothing too important either way it might confuse beginners.

rafaelfranca:
The readme always match the latest version of the framework and since Rails 4 is not officially supported anymore I don't think we need to change the README to talk about it. Thank you for the PR

PR #4592, opened 2017-07-20
Update links to use new path helpers

ybart:


tegon:
Hi @ybart, thanks for your pull request.
Your code looks ok to me, but IMHO the old version is simpler to understand - because we don't need to rely on metaprogramming - and we also don't need to create the intermediate `path_helper` variable.
That's why I'm going to close this, I hope you don't mind.

Thanks again!

ybart:
No problem. Thanks for reviewing!

Issue #4095, opened 2016-05-08
"Invalid Login or password." is grammatically incorrect


connorshea:
This was changed in PR #4014 and released with 4.1.0.

With an incorrect password the user sees "Invalid Login or password." "Login" shouldn't be capitalized in this case, as it's not a proper noun and isn't at the beginning of the sentence.

This is because of the following i18n string:

``` yaml
invalid: "Invalid %{authentication_keys} or password."
```
#4014 "humanized" the `authentication_keys` variable, thus causing this issue.

connorshea:
I'm not sure how this should be resolved, different languages will have different requirements with regards to capitalization.

ulissesalmeida:
Before `4.1` the keys was not "humanized", and we have something like `email or password`.
With the humanized version we have now something like `E-mail or password`.
If we keep this way, the `password` is not consistent with `E-mail`.
I think we should provide a nice default for the english language.

For other idioms, they can simply change the entire message of the `yml` file of the specific language with their specific rules. Without rely on `authentication_keys`.

```
invalid: "The idiom custom message here"
```

connorshea:
@ulissesalmeida I think I'm not following, are you saying we should not humanize them for English, or we should replace it with our own message that doesn't include `authentication_keys`?

ulissesalmeida:
@connorshea I think we should humanize them for English since it is the Devise default(downcasing the keys). For other languages that do not follow this default, I think they should change in their application for a proper translation that's doesn't include the `authentication_keys`.

crawfoal:
@ulissesalmeida Are there still plans to change this back to using all lowercase (but keeping the humanize change)?

bernardogalindo:
I think humanize authentication keys is good for English but why `password` is not part of ?

tegon:
@crawfoal @bernardogalindo We can follow with the downcase solution (keeping the humazine attribute). If someone is willing to open a PR for this, I'd be happy to accept it.

tegon:
Closed via https://github.com/plataformatec/devise/pull/4834

PR #4598, opened 2017-07-25
Fix capitalization of "Invalid Email or password"

fny:
Currently, `authentication_keys` is humanized which causes its first character to be capitalized. This causes awkward looking messages such as "Invalid Email or password." By moving `authentication_keys`, we can create a correct sentence for the english local without disrupting the behavior of the other locales.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -9,10 +9,10 @@ en:
     failure:
       already_authenticated: "You are already signed in."
       inactive: "Your account is not activated yet."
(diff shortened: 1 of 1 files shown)


fsinisi90:
I just had that problem today! I solved it by capitalizing the whole error message on my view, but it would be nice if it works properly by default.

tegon:
We intend to follow another solution. See https://github.com/plataformatec/devise/issues/4095#issuecomment-368010499

PR #4604, opened 2017-08-02
wrong number of arguments (given 1, expected 0) during password reset…

terryyin:
…. Rails 5.1.2

Since Rails 5.1.2, the default headers in ActionMailer behave differently if they are Proc. It doesn't take any parameter.

### DIFF
--- a/lib/devise/mailers/helpers.rb
+++ b/lib/devise/mailers/helpers.rb
@@ -51,7 +51,7 @@ module Devise
       def mailer_sender(mapping, sender = :from)
         default_sender = default_params[sender]
         if default_sender.present?
-          default_sender.respond_to?(:to_proc) ? instance_eval(&default_sender) : default_sender
+          default_sender.respond_to?(:to_proc) ? default_sender.call : default_sender
         elsif Devise.mailer_sender.is_a?(Proc)
           Devise.mailer_sender.call(mapping.name)
         else


jbourassa:
I am getting this too, but I think there are a couple things that are not working with the proposed solution:

- If the Proc has an arity greater than 1, this will crash on prior Rails versions (you said 5.1.2, but I think it has been merged in 5.1)
- Rails now uses `instance_exec` – simply `call`ing the method will not yield the same results.

Do you want to update this PR or should I create a new one?

terryyin:
You can create another PR, just wanted to point out the problem.

Actually in other places I simply add a default parameter to the Proc (_=nil).

But the question is how can we remember to remove this after some time?

jbourassa:
>But the question is how can we remember to remove this after some time?

Not sure that's a major issue. According to the PR I opened on Rails, they may want to keep both 1 and 0 arg.

terryyin:
In that case, isn't a default parameter good enough solution?

jbourassa:
>In that case, isn't a default parameter good enough solution?

It's good enough to use in your mailers, but Devise should work with more than just Rails 5.1, so to support both version you have to do this trickery.

tegon:
Hello @terryyin, thanks for your report.
I've just tested here with a new Rails 5.1.4 application and I haven't got this error during password reset.
Can you provide us a sample application that reproduces the issue in isolation?
That would help us find the issue.

Thank you!

tegon:
Closing in favor of #4627

PR #4608, opened 2017-08-08
Test with Rails 5.1

aried3r:


rafaelfranca:
Thank you for the pull request. It is already being testes with Rails 5.1 with the Gemfile in the root of the repository.

PR #4613, opened 2017-08-13
Use form_with instead of form_for

matthias-g:
Rails 5.1 introduced `form_with` as the new default to create forms (see http://edgeguides.rubyonrails.org/5_1_release_notes.html#unification-of-form-for-and-form-tag-into-form-with).
This PR updates the views to use this new helper.

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -1,11 +1,11 @@
 <h2>Resend confirmation instructions</h2>
 
-<%= form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: { method: :post }) do |f| %>
+<%= form_with model: resource, scope: resource_name, url: confirmation_path(resource_name), method: :post, local: true do |f| %>
   <%= devise_error_messages! %>
 
   <div class="field">
     <%= f.label :email %><br />
-    <%= f.email_field :email, autofocus: true, value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) %>
+    <%= f.email_field :email, autofocus: true, id: "#{resource_name}_email", value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) %>
   </div>
 
   <div class="actions">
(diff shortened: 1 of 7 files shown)


rafaelfranca:
Thank you for the PR. Devise supports versions of Rails that don't have `form_with` so we need to generate `form_with` only on Rails 5.1. Can you update the PR to do that?

matthias-g:
I see.
How do you want me to do that? Duplicate the views and implement a switch in the generator?

rafaelfranca:
I think we can add the conditional around the `form_for` calls

matthias-g:
As the `form_for` calls make up most of the views this would effectively result in duplicating these, wouldn't it?

```ruby
<% if Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 1 || Rails::VERSION::MAJOR > 5 %>
  <%= form_with ...%>
  ...
  <% end %>
<% else %>
  <%= form_for ...%>
  ...
  <% end %>
<% end %>
```

rafaelfranca:
You don't need to duplicate the body. You can do:

```ruby
<% if Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR >= 1 || Rails::VERSION::MAJOR > 5 %>
<%%= form_with ...%>
<% else %>
<%%= form_for ...%>
  ...
<%% end %>
<%% end %>
```

But we would need to change the `erb` file to be a template, not only a copy.

I wonder if it is worth to use `form_with` since `form_for` is still a valid method in Rails 5.1.

@lucasmazza toughts?

lucasmazza:
@rafaelfranca if Rails is sticking with `form_for` for the time being I don't think we should add more version conditionals around the codebase - when someone decides to rotate the dependency to support 5.1+ only (maybe when 5.2 drops?) we can apply this change.

rafaelfranca:
Yes. `form_for` is not going aways any time soon. Agree we should wait the 5.1+ support to change devise to use `form_with`.

@matthias-g Thank you for working on this! ❤️

PR #4616, opened 2017-08-14
Refactor the devise_error_messages! helper to render a partial

prograhamer:
This should make it more intuitive and easier for developers using the project to implement their own layout for errors.


tegon:
It seems ok to me, but I'm not sure about the backwards compatibility. WDYT @rafaelfranca?

PR #4627, opened 2017-08-29
Fix wrong arity in AM defaults procs

jbourassa:
Support Rails 5.1 + 5.0 by using the right arity for the proc.
Rails 5.1 changed the arity from 0 to 1.

See rails/rails#30391 and #4604 for details.

### DIFF
--- a/lib/devise/mailers/helpers.rb
+++ b/lib/devise/mailers/helpers.rb
@@ -51,7 +51,7 @@ module Devise
       def mailer_sender(mapping, sender = :from)
         default_sender = default_params[sender]
         if default_sender.present?
-          default_sender.respond_to?(:to_proc) ? instance_eval(&default_sender) : default_sender
+          compute_default_mapping_sender(default_sender)
         elsif Devise.mailer_sender.is_a?(Proc)
           Devise.mailer_sender.call(mapping.name)
         else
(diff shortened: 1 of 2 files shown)


tegon:
Hello @jbourassa, thanks for your report.
I've just tested here with a new Rails 5.1.4 application and I haven't got this error during password reset.
Can you provide us a sample application that reproduces the issue in isolation?
That would help us find the issue.

Thank you!

jbourassa:
Hey @tegon,

Here's a project where you can reproduce the issue: https://github.com/jbourassa/devise-4627-repro

It's a stock Rails 5.1.4 + Devise project, with the following changes:
- Generated user model
- Removed Devise's default sender, instead using the [app's default sender proc](https://github.com/jbourassa/devise-4627-repro/blob/master/config/application.rb#L14)

Steps to reproduce the issue:
- Create a user (`http://localhost:3000/users/sign_up`)
- Sign out (delete cookies?)
- Try to reset the password (`http://localhost:3000/users/password/new`)

tegon:
@jbourassa I get it now, thank you for the test repo, it really helps.
Your building is failing with the same error (`ArgumentError: wrong number of arguments (1 for 0)`) in some Rails versions, can you take a look at it?

Thanks!

jbourassa:
This is fixed in Rails 5.1.5 btw.

rafaelfranca:
Good to know. So I'm closing this PR since it was a bug in the Rails framework.

Thanks

PR #4655, opened 2017-09-28
Allow frozen options in as_json

guigs:
The test uses `as_json` instead of `to_json` because `to_json` does `#dup` on `options` before it reaches `#serializable_hash` and the test would pass without the fix.

See also this related change in Rails: https://github.com/rails/rails/pull/27945


tegon:
Hi @guigs, thanks for your pull request.
Can you rebase your branch with master please?

guigs:
@tegon Sure, done.

PR #4656, opened 2017-09-28
MongoDB is not required anymore to run tests

guigs:
I was able to run tests without having MongoDB server running in my system. So I guess it is not required anymore. Or am I missing something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ We hope that you will consider contributing to Devise. Please read this short ov
 
 https://github.com/plataformatec/devise/wiki/Contributing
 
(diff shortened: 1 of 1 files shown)


tegon:
Hello, @guigs thanks for your pull request.
You're right, if you run `rake test` the tests will run only for `active_record`. You can see in the output which `orm` is being used for the tests:
```
> rake test
```
But, in order to run the tests for all the environments, a `mongodb` instance is needed. It is possible to switch to `mongoid` by setting the `DEVISE_ORM` variable:
```
> DEVISE_ORM=mongoid rake test
==> Devise.orm = :mongoid
```
We're going to update the readme with more details of how to run our test suite - including how to switch rails/ruby versions too, so I will close this one.

Thanks again!

Issue #4673, opened 2017-10-16
Devise ignores (some) validations on signup


pcragone:
tl;dr: During the signup process, I'm guessing only if trackable is turned on, Devise will call a `save(validate: false)` on the model, causing it to save despite any failed validations, due to: (https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/models/trackable.rb#L33)[https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/models/trackable.rb#L33]

While trying to figure out what change I had made that broke things, I discovered my application was saving with failed validations during the signup process.  In my specific case, I have an `after_create` callback that calls create on an associated model.  However, that ends up being invalid, and so does not save (which led my tests to fail).  Yet if I tried it through the Devise signup process, everything went through and that associated model saved anyway.  I eventually tracked this down to the above referenced line within devise, causing my associated model to save despite being invalid.

FWIW, this doesn't actually affect how my application ends up performing, as I need that model created despite being invalid, so it actually works in my favor (I have an after-signup profile creation process users through, so that's why the validation exists in the first place).  I don't know if this is necessarily a bug you want fixed, or just something you want documented.

tegon:
Fixed in #4674

carlosantoniodasilva:
Just linking a comment on the final related fix here: https://github.com/heartcombo/devise/pull/4796#issuecomment-812659563.

I'm not sure I can see how Devise would be trying to save an invalid record that wasn't already saved before.

PR #4674, opened 2017-10-17
Fix missing validations on Signup

AshleyFoster:
This commit fixes issue
https://github.com/plataformatec/devise/issues/4673

This removes `validate: false` from saving a record when `Trackable` is
in use.

Alternatively, we could use `update_attribute` to update each field individually to skip validations.
That would have the downside of making multiple database calls though.

I'd also love any other ideas on how to solve this.

### DIFF
--- a/lib/devise/models/trackable.rb
+++ b/lib/devise/models/trackable.rb
@@ -30,7 +30,7 @@ module Devise
 
       def update_tracked_fields!(request)
         update_tracked_fields(request)
-        save(validate: false)
+        save
       end
     end
   end
--- a/test/models/trackable_test.rb
+++ b/test/models/trackable_test.rb
@@ -38,4 +38,13 @@ class TrackableTest < ActiveSupport::TestCase
     assert_nil user.last_sign_in_at
     assert_equal 0, user.sign_in_count
   end
+
+  test 'update_tracked_fields should run model validations' do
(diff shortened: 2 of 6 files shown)


rafaelfranca:
Thanks! I think your solution is good. Could you add a test case?

tegon:
Hi, @AshleyFoster thanks for your contribution!
Are you going to add a test case for this?

AshleyFoster:
Thanks for reviewing! I've just added a test case.

tegon:
@AshleyFoster thank you, the test looks good to me.
Looks like your test is failing when `mongoid` is the orm. I think that if you add the class `UserWithValidations` for `mongoid` the build will pass.
Would mind doing this change?

AshleyFoster:
I've added the `UserWithValidations` class for `mongoid`. I wasn't able to run the tests locally because I was getting `uninitialized constant User::Mongoid (NameError)` even on master.

I didn't look too much into it but maybe some info on running the tests with `mongoid` could be a good addition to the README?

tegon:
@AshleyFoster You're right, we have plans to improve our documentation on how to run our test suite locally using different ORMs.

dhiemstra:
Please see https://github.com/plataformatec/devise/issues/4790

PR #4678, opened 2017-10-23
Include Turbolinks if it is installed

wnm:
Turbolinks is overwriting the redirect_to method of ActionController::Redirecting.

Devise only includes ActionController::Redirecting

So, if turbolinks is installed, it doesn't work will with devise.

This should fix it.

### DIFF
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -284,7 +284,7 @@ Devise.setup do |config|
   # ==> Turbolinks configuration
   # If your app is using Turbolinks, Turbolinks::Controller needs to be included to make redirection work correctly:
   #
-  # ActiveSupport.on_load(:devise_failure_app)
+  # ActiveSupport.on_load(:devise_failure_app) do
   #   include Turbolinks::Controller
   # end
 end
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -280,4 +280,11 @@ Devise.setup do |config|
   # When using OmniAuth, Devise cannot automatically set OmniAuth path,
   # so you need to do it manually. For the users scope, it would be:
   # config.omniauth_path_prefix = '/my_engine/users/auth'
+
+  # ==> Turbolinks configuration
+  # If your app is using Turbolinks, Turbolinks::Controller needs to be included to make redirection work correctly:
+  #
+  # ActiveSupport.on_load(:devise_failure_app)
+  #   include Turbolinks::Controller
+  # end
 end
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -260,5 +260,7 @@ module Devise
     def relative_url_root?
       relative_url_root.present?
     end
+
+    ActiveSupport.run_load_hooks(:devise_failure_app, self)
   end
 end
--- a/test/failure_app_test.rb
+++ b/test/failure_app_test.rb
@@ -337,4 +337,10 @@ class FailureTest < ActiveSupport::TestCase
       end
     end
   end
+
+  context "Lazy loading" do
+    test "loads" do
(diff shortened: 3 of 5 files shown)


rafaelfranca:
The failure app should not know about turbolinks. I know it will redirect without turbolinks in the case the authentication fails, but devise should not know anything about turbolinks and fully rendering a failure page is not a performance problem to justify this coupling. Thank you for the pull request.

wnm:
@rafaelfranca it's not a performance issue, no. But using Turbolinks with the wrappers for Android and iOS requires Turbolinks redirects. The Webview doesn't follow normal redirects. 

But I understand you don't want to include this. 

For anyone stumbling opon this issue, put this in your devise config:

```
module Devise
  class FailureApp
    include Turbolinks::Controller
  end
end
```

rafaelfranca:
I see. We could take a different approach. Instead of adding that in the controller we could allow people to extend the devise failure app without monkey patching. Something like:

```ruby
ActiveSupport.run_load_hooks(:devise_failure_app)
```

in the controller and

```ruby
ActiveSupport.on_load(:devise_failure_app)
  include Turbolinks::Controller
end
```

tegon:
I agree we can provide a lazy load hook. And it seems it shouldn't be much hard to do it.
If I understood it right, we just need to add `ActiveSupport.run_load_hooks(:devise_failure_app, self)` to the end of the [failure app](https://github.com/plataformatec/devise/blob/master/lib/devise/failure_app.rb) class.

@wnm If you're willing to change this implementation to use a lazy load hook, we'd happily accept the pull request. Otherwise we can implement it in another one.

Some related resources:

http://api.rubyonrails.org/classes/ActiveSupport/LazyLoadHooks.html
https://simonecarletti.com/blog/2011/04/understanding-ruby-and-rails-lazy-load-hooks/

wnm:
@tegon I updated the pull request:

- Added `ActiveSupport.run_load_hooks(:devise_failure_app, self)` to the end of failure app class.
- Included config to load Turbolinks::Controller in devise initilizer template (but commented out)

Hope I understood everything correctly?

lib/generators/templates/devise.rb
```diff
@@ -280,4 +280,11 @@
   # When using OmniAuth, Devise cannot automatically set OmniAuth path,
   # so you need to do it manually. For the users scope, it would be:
   # config.omniauth_path_prefix = '/my_engine/users/auth'
+
+  # ==> Turbolinks configuration
+  # If your app is using Turbolinks, Turbolinks::Controller needs to be included to make redirection work correctly:
+  #
+  # ActiveSupport.on_load(:devise_failure_app)
```
tegon: I believe we're missing a `do` block here
```diff
- # ActiveSupport.on_load(:devise_failure_app)
+ # ActiveSupport.on_load(:devise_failure_app) do

**Comment by tegon** [comment]
@wnm Thanks! I think that's basically it. I just think we should have a test for this.
One way we can test it is to create a simple module with a method and include it in the failure app using this hook, then we can check if the failure app responds to that method.
If you need help with this test, let me know.

**Comment by wnm** [comment]
@tegon not sure if the tests are in the right place. let me know if they are not...

**Comment by tegon** [comment]
@wnm Looks good, thanks again!

**Comment by dsandstrom** [comment]
This doesn't seem to be required for Rails 6 and may cause confusion. Was trying to troubleshoot and tried to toggle this config on, but it errors for me.  My issue ended up being a conflict with my existing user routes.

I'm not using the turbolinks gem so there's no `Controller` to include. Instead, I installed turbolinks with yarn and it gets loaded in application.js.

Error:
```
uninitialized constant Turbolinks
```

More info about rails js: https://prathamesh.tech/2019/08/26/understanding-webpacker-in-rails-6/

Just an fyi. Thanks for the PR/gems.

Issue #4275, opened 2016-08-26
Rails 5 API doesn't work with devise, throws undefined local variable or method `flash'


johndel:
The new [rails 5 API](http://edgeguides.rubyonrails.org/api_app.html) doesn't work with devise (v4.2.0). To reproduce:
1. `rails new my_api --api`
2. Follow steps of installing devise (adding gem, running rake commands).
3. Try accessing `sign_in` with a valid user with something like that:
   `curl -X POST -F "user[email]=example@example.com" -F "user[password]=example" localhost:3000/users/sign_in`

It throws an error like this:

```
NameError (undefined local variable or method `flash' for #<Devise::SessionsController:0x007fb8b8814250>)
```

Due to this [line of code](https://github.com/plataformatec/devise/blob/v4.2.0/app/controllers/devise/sessions_controller.rb#L18). 

Something as simple as this commit, makes it work: https://github.com/johndel/devise/commit/c870213ee57b71ac580784979b46c7aab8863a2a

**Update:**
Seems that if you call it with `.json` at the end it works (like this `curl -X POST -F "user[email]=example@example.com" -F "user[password]=example" localhost:3000/users/sign_in.json`). 

**Update 2:**
Seems like adding the `.json` at the end of the `sign_in.json` url worked, but in other endpoints didn't, like `/password/edit.json`.

kreba:
I agree with John that this is an issue. The `PasswordsController#edit` does not respect navigational formats. (I did not check if this concerns other controllers, too.) 

This could be fixed by using `set_flash_message!` instead of `set_flash_message` [here](https://github.com/plataformatec/devise/blob/v4.2.0/app/controllers/devise/passwords_controller.rb#L64).

Ikariusrb:
I will chime in that I suspect this is more of an issue that it appears at first blush.

I'm attempting to use devise with a rails --api (5.0.0.1) application, and while it does the correct thing if an end-user submits a request with the correct type, it throws a 500 if a user tries to do a simple html format GET from one of my endpoints without providing authorization, rather than returning a proper 401.

I could limit routes to specific formats (btw, anyone yet figure out how to use route constraints to limit routes to jsonapi format?),  which will then throw an unknown route error,  but I don't control what formats users may throw requests at my server in;  I'd like to respond appropriately.  

Perhaps a config flag to never use flash would be appropriate?

Ikariusrb:
[...]

PR #4680, opened 2017-10-24
set_flash_method! instead of set_flash_method in omniauth callbacks controller

saichander17:


rafaelfranca:
Thank you for the pull request. Can you explain why we need this change?

tegon:
Hello @saichander17, thanks for your pull request.
Would you mind telling us the motivation behind it?

KelseyDH:
Would this finally be a solution to https://github.com/plataformatec/devise/issues/4275 ?

rafaelfranca:
This doesn't change anything related to #4275. I mean, it doesn't try to access the flash message if you don't request using a navigational format, but if you do you will get the same problem. The right way to fix #4275 is to not do a HTML request (navigational) in a API application, like the two updates in the issue description shows.

KelseyDH:
@rafaelfranca  Thanks for responding.  While I get that its not ideal for a Rails API to serve HTML, with `devise_token_auth` I struggle to see how there is any other way of doing a password reset.

As I added to this [mega thread of people pulling their hair out trying to get a sane password reset system working with devise_token_auth](https://github.com/lynndylanhurley/devise_token_auth/issues/604#issuecomment-351180401), an API Rails-only email password reset flow doesn't work if the client apps that need to be opened from a password reset link sent by email are iOS and Android:  custom app scheme redirects are blocked by Gmail, and it appears the proper solution to getting around this is to handle these scheme redirects [within an app endpoint first, to get around Gmail's stripping of custom scheme URL's.](https://github.com/EddyVerbruggen/Custom-URL-scheme/issues/81#issuecomment-97701155)

So in light of how Google treat's custom scheme URL's in password resets... I am unsure of how I can achieve a seemless password reset redirect to a mobile app without serving some form of HTML page from my Rails API.  (When you look at app's like Slack, or any other app where there is a custom app scheme redirect from an email, it's almost always done via a link that first goes to an app-generated HTML page first.)

On password reset email links `devise_token_auth` users seem caught in a catch 22:  don't use html and have their iOS/android custom app scheme URL's break in gmail, or add devise to get a web form for their password reset portion of their app, and have it completely break when using Rails api due to design decisions made by `devise`.

(Any advice on solving this would be much appreciated!!)

rafaelfranca:
If you are rendering HTML you Rails application can't be an API-only application anymore. You application now need flash, sessions, CSRF protection, etc. As your application now needs to render html. I'd recommend you to use `ActionController::Base` instead of `ActionController::API` as the superclass of you reset password controller. Doing that you will have the `flash` method again and that code will not raise.

KelseyDH:
@rafaelfranca   Thank you, subclassing my controllers to create two ApplicationControllers, one renamed to `ApiController` pointing to Rails API as `ApiController < ActionController::Api`, and the other to `ApplicationController < ActionController::Base`, appears to have allowed me to get devise html up and running.  (Though not sure if doing this is a bad idea or may produce unknown side effects.)

PR #4681, opened 2017-10-26
Provide config info into the translation files

kyoshidajp:
Enable to reference config info from the translation files.

For example, if the paremeter `maximum_attempts` value is default(20), and the translation file is like this.

```yml
en:
  devise:
    failure:
      locked: "Your account is locked. Maximum amount of failed attempts(%{resource_maximum_attempts}) was reached."
```

Then, the result is translated would be the following.

```
Your account is locked. Maximum amount of failed attempts(20) was reached.
```

To inform the detail reason why user failed to log in. It might be a good idea to specify the config values as part of error message. In that case, it will be easy to change the code.

### DIFF
--- a/app/controllers/devise_controller.rb
+++ b/app/controllers/devise_controller.rb
@@ -176,11 +176,20 @@ MESSAGE
     options
   end
 
+  def devise_config_options(options)
+    Devise.class_variables.each do |k|
+      key_name = k.to_s.sub('@@', '')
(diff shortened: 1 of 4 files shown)


rafaelfranca:
Thank you for the pull request but I don't think we need to add more things to the devise i18n. If you want to provide this information to your users you already able to show that information changing the message yourself. Of course it would not be dynamic, but it doesn't need to.

kyoshidajp:
All right. Thanks.

PR #4682, opened 2017-10-27
Removed gem minitest

haneru:
I removed `gem "minitest", "< 5.10.2"` from Gemfile, because Rails 5.1.1 has already been released


rafaelfranca:
Thanks!

PR #4684, opened 2017-10-27
Fix typo in README

fhinson:
`rails db:migrate` should be `rake db:migrate`


rafaelfranca:
Thank you for the pull request. `rails db:migrate` is a valid command in Rails 5. Since Rails 5 is the only supported version of Rails right now our README is reflecting that.

PR #4688, opened 2017-11-01
Replaced Double quotes with Single Quotes: prefer single-quoted strin…

anoobbava:
…gs when not string interpolation

### DIFF
--- a/app/helpers/devise_helper.rb
+++ b/app/helpers/devise_helper.rb
@@ -6,7 +6,7 @@ module DeviseHelper
   # This method is intended to stay simple and it is unlikely that we are going to change
   # it to add more behavior or options.
   def devise_error_messages!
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thanks for the pull request but as you can see in this same file we uses double quote for other strings and there is no preference for single quotes in our codebase. That being said we are not willing to accept this kind of cosmetic change.

anoobbava:
@rafaelfranca  I believe there is a meaning for single quotes and double quotes. Here we are breaking that rule. We are supposed to be clearing if there is an issue, please explain what is the issue with that.

rafaelfranca:
> Here we are breaking that rule.

Which rule? We don't follow rules define by some project. The style rules of the devise codebase is: we don't care about which quotes are being used.

PR #4690, opened 2017-11-01
make methods inline and adjusted the space in devise controller

anoobbava:


rafaelfranca:
Thank you for the pull request but we are not willing to change this codebase code style to use rubocop's default styles.

Issue #3069, opened 2014-06-13
undefined local variable or method `current_user' when devise routes are defined within a scope


Sephi-Chan:
Hello,

I build a multitenant app and I have specific logic for authentication in two tenants so I defined the devises routes in my constrained scope (in order to have two sets of namespaced routes, with their own controllers):

``` ruby
scope(as: 'geea', module: 'geea', constraints: TenantConstraint.new('geea')) do
  devise_for(:users, {
    encryptor: :md5,
    controllers: {
      confirmations:      'geea/users/confirmations',
      sessions:           'geea/users/sessions',
      registrations:      'geea/users/registrations',
      omniauth_callbacks: 'geea/users/omniauth_callbacks',
      passwords:          'geea/users/passwords'
    }
  })

  devise_scope :user do
    namespace :backoffice do
      get    'signin',  to: 'sessions#new',     as: 'new_user_session'
      post   'signin',  to: 'sessions#create',  as: 'user_session'
      delete 'signout', to: 'sessions#destroy', as: 'destroy_user_session'
    end
  end
end
```

The problem is that causes Devise helpers not to be available in my controllers.

```
undefined local variable or method `current_user' for #<Geea::Frontoffice::HomeController:0x007fd14d4a36e0>
```

How could I deal with it?

Thanks!

Issue #2840, opened 2014-01-22
devise_for vs devise_scope routes


michaelirey:
```
Using devise (3.2.2)
Using rails (4.0.2)
```

Working on developing an API which only has the need for only certain routes. I have the following in my `routes.rb`:

```
devise_for :users, path: '/users', controllers: {
  sessions: 'v1/custom_devise/sessions',
  passwords: 'v1/custom_devise/passwords'
}
```

Great! Now we have all these routes:

```
              Prefix Verb    URI Pattern                    Controller#Action
    new_user_session GET     /users/sign_in(.:format)       v1/custom_devise/sessions#new
        user_session POST    /users/sign_in(.:format)       v1/custom_devise/sessions#create
destroy_user_session DELETE  /users/sign_out(.:format)      v1/custom_devise/sessions#destroy
       user_password POST    /users/password(.:format)      v1/custom_devise/passwords#create
   new_user_password GET     /users/password/new(.:format)  v1/custom_devise/passwords#new
  edit_user_password GET     /users/password/edit(.:format) v1/custom_devise/passwords#edit
                     PATCH   /users/password(.:format)      v1/custom_devise/passwords#update
                     PUT     /users/password(.:format)      v1/custom_devise/passwords#update
```

But I do not want or need these routes:

```
    new_user_session GET     /users/sign_in(.:format)       v1/custom_devise/sessions#new
   new_user_password GET     /users/password/new(.:format)  v1/custom_devise/passwords#new
  edit_user_password GET     /users/password/edit(.:format) v1/custom_devise/passwords#edit
```

So I thought I will just create my own routes as per the documentation. So I changed my `routes.rb`:

```
devise_scope :user do
  post   '/users/sign_in'  => 'custom_devise/sessions#create',  as: :user_session
  delete '/users/sign_out' => 'custom_devise/sessions#destroy', as: :destroy_user_session

  post  '/users/password'  => 'custom_devise/passwords#create', as: :user_password
  put   '/users/password'  => 'custom_devise/passwords#update', as: nil
  patch '/users/password'  => 'custom_devise/passwords#update', as: nil
end
```

Now my routes look perfect, just the way I want them:

```
              Prefix Verb    URI Pattern                   Controller#Action
        user_session POST    /users/sign_in(.:format)      v1/custom_devise/sessions#create
destroy_user_session DELETE  /users/sign_out(.:format)     v1/custom_devise/sessions#destroy
[...]

PR #4694, opened 2017-11-06
Provide additional info on devise_scope usage

rpbaptist:
After using `devise_scope` instead of `devise_for` I found I no longer had access to Devise helper`current_user`.

After searching I found this issue to be the answer to the problem:: https://github.com/plataformatec/devise/issues/2840

The issue is also reporter here: https://github.com/plataformatec/devise/issues/3069

I think this simple comment will clarify the issue and help other users in case they run into the same issue.


rafaelfranca:
Thanks

Issue #4127, opened 2016-05-22
Routes issue when devise engine is mounted in another engine


bilby91:
I'm building an engine that uses devise inside for user authentication. I have followed the instructions in the wiki (https://github.com/plataformatec/devise/wiki/How-To:-Use-devise-inside-a-mountable-engine) but I couldn't make it work as expected. It seems that `router_name` is being ignored. Maybe I'm miss understanding something.

I have created a simple engine with a dummy app that reproduces my issue. You can find it here: https://github.com/bilby91/devise_mountable_engine_issue. In the `spec/dummy` app there is a authenticated route under `/` path. When user is redirected to the sign_in path, the route doesn't build correctly and ignores the `router_name`.

I could make it work by providing a custom FailureApp to warden that handles the redirect.

```
module MyEngine

  class FailureApp < Devise::FailureApp

    def redirect_url
      engine_route = Rails
        .application
        .routes
        .routes
        .select { |r| r.app.app == MyEngine::Engine }
        .first

      "#{engine_route.path.spec.to_s}/users/sign_in"
    end
     #
    #  # You need to override respond to eliminate recall
    def respond
      if http_auth?
        http_auth
      else
        redirect
      end
    end

   end

end
```

I really don't like this solution :(. Any help is more than welcome!

Thanks!

bilby91:
Any help ?

petrosp:
+100 for this. I have exactly the same issue and ended up using a `FailureApp` which is not really solution but a workaround.

Is anybody having a look at this?

petrosp:
@bilby91, have you heard anything on this issue?

bilby91:
@petrosp nothing yet. 🙁

bilby91:
I know lots of people with this same problem. It would be awesome to have some feedback :)

fmh:
+1, I think this wiki https://github.com/plataformatec/devise/wiki/How-To:-Use-devise-inside-a-mountable-engine is no longer works ... or is outdated

sognatore24431:
The same..... .o_0     
-100 rep for devise.. ..

a-barbieri:
I encountered the same problem after upgrading from Rails `5.0` to `5.1` and moving from Devise `4.1.1` to `4.3.0`. That [wiki](https://github.com/plataformatec/devise/wiki/How-To:-Use-devise-inside-a-mountable-engine) needs to be updated or something needs to be fixed.

a-barbieri:
[...]

PR #4700, opened 2017-11-17
FIX plataformatec/devise#4127

a-barbieri:
I've been able to fix the issue for https://github.com/plataformatec/devise/issues/4127#issuecomment-309112353 though I'd like some help to create the test. 

It's pretty simple: the engine should point to a directory inside the engine, in order to trigger the redirect. Then the test would be:

```ruby
test 'redirect to login page when browsing an internal route without being authenticated' do
    get '/mountable_engine/some_route'
    follow_redirect!
    assert_response :success
    assert_contain 'Rendered content of MyMountableEngine'
end
```

@basiszwo, how can I add to [your code](https://github.com/plataformatec/devise/blob/cbbe932ee22947fb7fc741a4da3e6783091c88b0/test/integration/mounted_engine_test.rb#L3-L7) an internal route `some_route ` in order to test it?

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -142,10 +142,14 @@ module Devise
 
       opts[:format] = request_format unless skip_format?
 
-      opts[:script_name] = relative_url_root if relative_url_root?
-
       router_name = Devise.mappings[scope].router_name || Devise.available_router_name
       context = send(router_name)
+
+      if relative_url_root?
+        opts[:script_name] = relative_url_root
+      elsif defined? context.routes
+        opts[:script_name] = context.routes.url_helpers.root_path
+      end
 
       if context.respond_to?(route)
         context.send(route, opts)
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -148,7 +148,8 @@ module Devise
       if relative_url_root?
         opts[:script_name] = relative_url_root
       elsif defined? context.routes
-        opts[:script_name] = context.routes.url_helpers.root_path
+        rootpath = context.routes.url_helpers.root_path
+        opts[:script_name] = rootpath.chomp('/') unless rootpath.length <= 1
       end
 
       if context.respond_to?(route)
--- a/test/integration/mounted_engine_test.rb
+++ b/test/integration/mounted_engine_test.rb
@@ -1,9 +1,22 @@
 require 'test_helper'
 
-class MyMountableEngine
-  def self.call(env)
-    ['200', { 'Content-Type' => 'text/html' }, ['Rendered content of MyMountableEngine']]
+module MyMountableEngine
+  class Engine < ::Rails::Engine
+    isolate_namespace MyMountableEngine
   end
+  class TestsController < ActionController::Base
+    def index
+      render plain: 'Root test successful'
+    end
+    def inner_route
+      render plain: 'Inner route test successful'
+    end
+  end
+end
+
+MyMountableEngine::Engine.routes.draw do
+  get 'test', to: 'tests#inner_route'
+  root to: 'tests#index'
 end
 
 # If disable_clear_and_finalize is set to true, Rails will not clear other routes when calling
@@ -13,7 +26,7 @@ Rails.application.routes.disable_clear_and_finalize = true
 
 Rails.application.routes.draw do
   authenticate(:user) do
-    mount MyMountableEngine, at: '/mountable_engine'
+    mount MyMountableEngine::Engine, at: '/mountable_engine'
   end
 end
 
@@ -31,6 +44,15 @@ class AuthenticatedMountedEngineTest < Devise::IntegrationTest
     get '/mountable_engine'
 
     assert_response :success
-    assert_contain 'Rendered content of MyMountableEngine'
+    assert_contain 'Root test successful'
+  end
+
+
+  test 'renders a inner route of the mounted engine when authenticated' do
+    sign_in_as_user
+    get '/mountable_engine/test'
+
+    assert_response :success
+    assert_contain 'Inner route test successful'
   end
 end
--- a/test/integration/mounted_engine_test.rb
+++ b/test/integration/mounted_engine_test.rb
@@ -55,4 +55,12 @@ class AuthenticatedMountedEngineTest < Devise::IntegrationTest
     assert_response :success
     assert_contain 'Inner route test successful'
   end
+
+  test 'respond properly to a non existing route of the mounted engine' do
+    sign_in_as_user
+    
+    assert_raise ActionController::RoutingError do
+      get '/mountable_engine/non-existing-route'
+    end
+  end
 end


tegon:
Hello @a-barbieri, thanks for your pull request.
Did you manage to write a test for this?

a-barbieri:
Hi @tegon, I manage to create the test. Pretty straight forward. Let me know if it's ok on your side.

a-barbieri:
Apparently fixing the issue generated by the [previous PR](https://github.com/plataformatec/devise/pull/4081) we brought back an even older issue which led to an infinite loop on Rails 4.2.

Locally I cannot reproduce the error and test are ok. How can I test just that version of Rails?

tegon:
@a-barbieri you can use the `BUNDLE_GEMFILE` environment variable. Here's what I usually do:
```bash
rbenv local 2.2.8 # or rvm use 2.2.8
BUNDLE_GEMFILE=gemfiles/Gemfile.rails-4.2-stable bundle install
BUNDLE_GEMFILE=gemfiles/Gemfile.rails-4.2-stable rake test
```

a-barbieri:
Thanks @tegon, running 
```
BUNDLE_GEMFILE =gemfiles/Gemfile.rails-4.2-stable rake test
``` 
was throwing a error:
```
/path/to/devise/test/rails_app/gemfiles/Gemfile.rails-4.2-stable not found
```
I managed to make it work adding the local path to Devise repo
```
BUNDLE_GEMFILE=/path/to/devise/gemfiles/Gemfile.rails-4.2-stable rake test
```
---
**Anyway nothing fails.**
```
Finished in 19.669069s, 41.3848 runs/s, 110.2747 assertions/s.
814 runs, 2169 assertions, 0 failures, 0 errors, 0 skips
```
Can you try on your machine? Or run again Travis to confirm the failure? 
I really don't know how to reproduce this failing test in order to debug it.

a-barbieri:
The only thing that I noticed is that there is this message between the dots:
```
Expected string default value for '--test-framework'; got false (boolean)
```
What is the meaning of that? Any hint?

tegon:
@a-barbieri Indeed, I triggered another build and this time everything passed. I'll review your PR later, I still need to read the related issues to fully understand the problem.

Thanks again for your contribution!

a-barbieri:
Hi @tegon, it's been some time since I fixed this issue so I'm trying to catch up from where I left it. 

When I first came across it I was using Rails 5.1 and 4.3. See https://github.com/plataformatec/devise/issues/4127#issuecomment-308959262

---

> Notice that the engine app is in the url: http://localhost:3000/blorgh/users/sign_in. As far as I understood, this is the correct behavior or am I missing something?

You are right. Back then it didn't work because `relative_url_root ` would not bring `blorgh` to `opts`. Mounting the engine `relative_url_root ` didn't work. If it does now it means someone did fix it. See https://github.com/plataformatec/devise/issues/4127#issuecomment-345236599

Can you check that `relative_url_root ` contains `blorgh` with Devise 4.5 and 5.2?

You should debug `relative_url_root ` and the `opts` hash at those lines: 
https://github.com/plataformatec/devise/blob/48aa20897f2a2022c970398c67495cfb04be5f2a/lib/devise/failure_app.rb#L145
https://github.com/plataformatec/devise/blob/cbbe932ee22947fb7fc741a4da3e6783091c88b0/lib/devise/failure_app.rb#L157

---

Regarding

> undefined method `root_path`

Can you just double check that you have a `root` set within your routes? My `if` statement was actually pretty loose.

---

TL;DR

If Rails has fixed `relative_url_root` on a mounted engine we don't need this PR anymore.

feliperenan:
> Can you check that relative_url_root contains blorgh with Devise 4.5 and 5.2?

Both `relative_url_root?` and  `relative_url_root ` returns nil. 

> Can you just double check that you have a root set within your routes? My if statement was actually pretty loose.

Ohh, I had not the `root_path` configured in the Blorgh Engine 🙈. Now your code works well, even in Rails 4.2

For some reason it works without your patch on Rails 5.2 (even with `relative_url_root`), but I don't know if something has changed inside Rails to get it working on that version.

Here the app that I used to test it: https://github.com/feliperenan/engine-test-app

feliperenan:
I think I've found the issue.

`ActionDispatch` has changed the way it finds the `script_name` in [Rails 5.1](https://github.com/rails/rails/blob/5-1-stable/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L34), we can see the different implementations among Rails 4.2 and Rails 5.2:

[Rails 4.2 - router_proxy.rb](https://github.com/rails/rails/blob/7a48f5869e7b9723cf94b5366ccaeb1e17060807/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L32)
[Rails 5.0 - router_proxy.rb](https://github.com/rails/rails/blob/5a0fce7d4ecd152e883877bc5a6a11af54207fde/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L32)
[Rails 5.1 - router_proxy.rb](https://github.com/rails/rails/blob/bb22fe9d4a6102d2a28cb1adfd6fe9d38fc9bb22/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L34)
[Rails 5.2 - router_proxy.rb](https://github.com/rails/rails/blob/52af51c08a48ea7c6e0df09e2470a66a3141087d/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L37)


### Debugging Devise with Rails 5.2
Debugging into Rails 5.2 app, I can see that `context.send(route, opts)` returns the correct path even with `{script_name: nil}`

```Ruby
context.send(rout, {script_name: nil})
# => "http://localhost:4000/blorgh/users/sign_in"

context.send(rout, {})
# => "http://localhost:4000/blorgh/users/sign_in"
```

The `/blorgh` path is return by `@script_namer.call(options)` and after merged into options.

[building script name on Rails 5.1+](https://github.com/rails/rails/blob/52af51c08a48ea7c6e0df09e2470a66a3141087d/actionpack/lib/action_dispatch/routing/routes_proxy.rb#L38)

### Debugging Devise with Rails 4.2
On the other hand, on Rails 4.2, when `context.send(route, opts)` call happens with `{script_name: nil}`, it returns the wrong path.

```Ruby
context.send(rout, {script_name: nil})
# => "http://localhost:4000/users/sign_in"

context.send(rout, {})
# => "http://localhost:4000/my_engine/users/sign_in"
```

Looking into Rails 4.2 code, on the way, they check if the `options` has a `script_name` and if it does they call `super` and returns the wrong path since `script_name` is nil.

```Ruby
script_name = find_script_name options
# => "http://localhost:4000/users/sign_in"

options.delete(:script_name)
script_name = find_script_name options
# => "http://localhost:4000/my-engine/users/sign_in"
```

[find_script_name call](https://github.com/rails/rails/blob/7a48f5869e7b9723cf94b5366ccaeb1e17060807/actionpack/lib/action_dispatch/routing/route_set.rb#L790)
[find_script_name method definition](https://github.com/rails/rails/blob/7a48f5869e7b9723cf94b5366ccaeb1e17060807/actionpack/lib/action_dispatch/routing/mapper.rb#L631)

### Where context.send is called on Devise code
https://github.com/plataformatec/devise/blob/c000b58c565f46cb63d52ade0807294943b20cd3/lib/devise/failure_app.rb#L153

### TL;DR
We should apply your patch considering the change you've mentioned (checking for `root_path`), but also adding a comment to remove these checks after we remove support of Rails 5.0 or we could add these checks only for Rails versions lower than Rails 5.0.

lib/devise/failure_app.rb
```diff
@@ -142,11 +142,16 @@ def scope_url
...
       opts[:format] = request_format unless skip_format?
 
-      opts[:script_name] = relative_url_root if relative_url_root?
-
       router_name = Devise.mappings[scope].router_name || Devise.available_router_name
       context = send(router_name)
 
+      if relative_url_root?
+        opts[:script_name] = relative_url_root
+      elsif defined? context.routes
```
feliperenan: We could create a private method to ask if the `root_path` is defined.

```ruby
module Devise
  class FailureApp < ActionController::Metal
    ...

    private

    def root_path_defined?(context)
      defined?(context.routes) && context.routes.url_helpers.root_path.present?
    end
  end
end
```

Wdyt?

lib/devise/failure_app.rb
```diff
@@ -142,11 +142,16 @@ def scope_url
...
+        rootpath = context.routes.url_helpers.root_path
+        opts[:script_name] = rootpath.chomp('/') unless rootpath.length <= 1
```
feliperenan: My brain doesn't process `unless` very well 🙈. I always prefer an `if` with the positive case, for me, it's easier to read. So I'd suggest to change it to:
```ruby
opts[:script_name] = rootpath.chomp('/') if rootpath.length > 1
```

Wdyt?

a-barbieri: I do agree. `if` is to be preferred, so please go with it.

PR #4711, opened 2017-11-28
Test on Rails 5.2.0.rc1

jcoyne:


tegon:
Hello @jcoyne, thanks for your pull request.
Could you do a rebase with master, please?

test/rails_app/config/application.rb
```diff
@@ -31,8 +31,10 @@ class Application < Rails::Application
     # config.assets.enabled = false
 
     config.action_mailer.default_url_options = { host: "localhost", port: 3000 }
-
-    if DEVISE_ORM == :active_record && (Rails::VERSION::MAJOR >= 4 && Rails::VERSION::MINOR >= 2)
+    rails_version = Gem::Version.new(Rails.version)
```
tegon: Would you mind explaining why do we need this change?

jcoyne: `raise_in_transactional_callbacks` was deprecated in rails 5.0 and removed in 5.2 https://github.com/rails/rails/commit/8029f779b8a1dd9848fee0b7967c2e0849bf6e07

jcoyne:
@tegon I've rebased this on master.

test/rails_app/config/application.rb
tegon: I see. The strange thing is that some specs are failing on CI with an error that seems to be related to this
```
`method_missing': undefined method `raise_in_transactional_callbacks=' for ActiveRecord::Base:Class
```
Could you check those errors?

jcoyne:
@tegon I've got this passing in its entirety except for the mongoid builds.  I have no idea about what's going on with mongoid. Can anyone offer help?

test/rails_app/config/application.rb
jcoyne: This is now fixed.

tegon:
@jcoyne It seems that the problem is because you didn't add the `mongoid` gem to your `Gemfile`.

jcoyne:
@tegon I see, I copied the rails 5.0 gemfile, which also has mongoid commented out, so I added an exclusion to the matrix as rails 5.0 does.

test/rails_app/config/application.rb
gregoirenovel: @jcoyne is it fixed on `master`? I still encounter this error

tegon:
@jcoyne You're right, I don't know why it's excluding `mongoid` but since it's already doing this, I'll take a look at this later.
The build still isn't passing, though. Can you take a look, please?

jcoyne:
@tegon i have looked, I don't know what that is about... it seems to be something related to how the testing framework is stubbing rails?

tegon:
I'm lost on this one too. @rafaelfranca do you know what might be causing this error?

jasnow:
Look at this: https://github.com/rails/rails/pull/29699

rafaelfranca:
`ActiveRecord::Migrator` don't have a `migrate` method anymore.

We need to change this

https://github.com/plataformatec/devise/blob/master/test/orm/active_record.rb#L7

to:

`MigrationContext.new(File.expand_path("../../rails_app/db/migrate/", __FILE__).migrate`

tegon:
@jcoyne Can you make this change, please?
I think we'll have to keep both syntaxes with a branch condition so that it works on multiple Rails versions.

jcoyne:
@tegon Yes, I'll get to this presently.

jcoyne:
@tegon that seems to have fixed the problem

@rafaelfranca Thank you for your assistance.

PR #4725, opened 2017-12-15
Add the frozen_string_literal pragma comment to all Ruby files.

pat:
As requested by @rafaelfranca and @tegon in #4566. I think I've got every relevant file covered :)


pat:
Thanks for merging :)

PR #4668, opened 2017-10-12
Fix syntax for MRI 2.5.0-preview1.

pat:
This ensures the tests are green when run on MRI 2.5.0-preview1 :)

PR #4748, opened 2018-01-09
Fix syntax error in ruby 2.5.0

schovi:
When trying to update rails app to ruby 2.5.0 it fails when booting on following error

```
gems/activesupport-5.1.4/lib/active_support/dependencies.rb:476:in `load': gems/devise-4.4.0/app/controllers/devise/sessions_controller.rb:5: syntax error, unexpected '{', expecting keyword_end (SyntaxError)
...ion only: [:create, :destroy] { request.env["devise.skip_tim...
...                              ^
gems/devise-4.4.0/app/controllers/devise/sessions_controller.rb:5: syntax error, unexpected '}', expecting keyword_end
..."devise.skip_timeout"] = true }
...                              ^
```

This fix it.


tegon:
This was fixed on #4668

PR #4752, opened 2018-01-10
Fix `ActiveRecord` check on `Confirmable`

tegon:
As pointed out by @dark-panda in #4302, the condition for an `ActiveRecord` model is wrong inside the `Confirmable` initialization block.

https://github.com/plataformatec/devise/pull/4302#issuecomment-355103489

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -48,7 +48,7 @@ module Devise
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
         after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
-        if defined?(ActiveRecord) && self.is_a?(ActiveRecord::Base) # ActiveRecord
+        if defined?(ActiveRecord) && self < ActiveRecord::Base # ActiveRecord
           after_commit :send_on_create_confirmation_instructions, on: :create, if: :send_confirmation_notification?
           after_commit :send_reconfirmation_instructions, on: :update, if: :reconfirmation_required?
         else # Mongoid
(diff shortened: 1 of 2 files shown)


rafaelfranca:
It makes sense, but since no test were failing we are missing some tests.

tegon:
@rafaelfranca I agree, but I'm struggling to figure out which kind of test would make sense here 🤔 
We have an [integration test](https://github.com/plataformatec/devise/blob/master/test/integration/confirmable_test.rb) that covers the desired behavior - for example, a confirmation email should be sent after the user is registered. This test doesn't fail because even though we fall on the `else` condition, we still define a callback to send the email - `after_create` instead of `after_commit` - but the email is sent either way.
Do you think we need to test ActiveRecord's internal methods to check what callbacks are being defined?

rafaelfranca:
oh, I see. Right now the only way to test this is to check the callbacks yes.

tegon:
Got it, I've added a test.

lib/devise/models/confirmable.rb
```diff
@@ -48,7 +48,7 @@ module Confirmable
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
         after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
-        if defined?(ActiveRecord) && self.is_a?(ActiveRecord::Base) # ActiveRecord
+        if defined?(ActiveRecord) && self < ActiveRecord::Base # ActiveRecord
```
feliperenan: I think the comment `#ActiveRecord` isn't necessary. What do you think to remove it?

PR #4759, opened 2018-01-17
Update sample code in Devise::Models::Authenticatable.

fredngo:
- DRY up some of the sample code by refactoring into smaller methods.
- ‘Namespace’ method names to reduce chances of conflict.
- Mark `send_pending_devise_notifications`, `pending_devise_notifications`, and `render_and_send_devise_message` methods as `private` since they are only used internally.
- Update comments.


fredngo:
Thank you!

PR #4775, opened 2018-02-04
Add an option to disable GET requests on /users/auth/provider to address RECONNECT vulnerability

JackMc:
This PR allows users of Devise's Omniauth support to mitigate the RECONNECT vulnerability specifically for Facebook OAuth discussed here: https://sakurity.com/blog/2015/03/05/RECONNECT.html

Specifically we address the vulnerability by adding an option `disable_omniauth_get` which allows the library user to specify a list of providers for which they do not wish to have a GET route available for the `/users/auth/:provider` endpoint. This mitigates the vulnerability by adding CSRF at the application level before it even reaches Facebook.

cc @rafaelfranca

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -292,6 +292,10 @@ module Devise
   mattr_accessor :token_generator
   @@token_generator = nil
 
+  # List of providers to disable GET-based oauth passthru for
+  mattr_accessor :disable_omniauth_get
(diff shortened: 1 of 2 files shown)


JackMc:
Never mind, turns out in a RECONNECT like situation this endpoint is never called and the fix should go into `omniauth-facebook`.

PR #4776, opened 2018-02-07
Added password_regex validation to validatable

adi-pen:
We use Validatable instead of the custom validations as mentioned in https://github.com/plataformatec/devise/wiki/How-To:-Upgrade-to-Devise-4.4.0.

We still have a custom validation to check the password complexity which can be easily added to Validatable.


rafaelfranca:
Thank you for the pull request but we don't believe validating format of password is a good idea. That said we prefer to not add support for this by default in the library, but you should be still able to validate the format of the password by yourself defining the validation in the mode.

Issue #4540, opened 2017-05-17
Can you add a method to get remote IP for Trackable module?


hegwin:
I'm using SLB(Server Load Balancing) serivers that forword requests to my rails servers. In this condition `request.remote_ip` doesn't refers to the real IP of a user but the IP of the SLB machines.

And after I look up into Trackable module, I find I need to define the whole `update_tracked_fields` methods just for a single line.

```ruby
  def update_tracked_fields(request)
    old_current, new_current = self.current_sign_in_at, Time.now.utc
    self.last_sign_in_at     = old_current || new_current
    self.current_sign_in_at  = new_current

    old_current, new_current = self.current_sign_in_ip, (request.headers['X-Forward-For'] || request.remote_ip)
    self.last_sign_in_ip     = old_current || new_current
    self.current_sign_in_ip  = new_current

    self.sign_in_count ||= 0
    self.sign_in_count += 1
  end
```

So can you separate the process to get remote ip to a single method?

Or can I do a fork-and-pull-request? Is that welcome?

tegon:
Hello @hegwin, thanks for your report.
I don't see any problem in doing this. If you're willing to tackle this one, feel free to do it.

Thank you!

camilleroux:
Hi!
Using Cloudflare, I've the same need. Cloudflare use the header `X-Forwarded-For` and can contain IPv6 addresses like:
`X-Forwarded-For: 
2a21:ce1d:824e:900:7529:daaa:b207:a897, 134.156.21.145`

mckramer:
FYI, I tossed together a quick proposal as #4777

RouL:
Just seeing this. Maybe we should only allow this for a list of trusted proxy IPs or when explicitly enabled. If there is no Proxy in front filtering the header, then malicious users could set it for making sure their real ip is not tracked. 

Just a thought.

tegon:
@RouL You mean this should be handled in the application, right? 
When you implement the method to get the IP you can do the security checks that you need.

tegon:
Closed via #4777

PR #4777, opened 2018-02-14
Split IP resolution from update_tracked_fields

mckramer:
This permits users to easily customize where the ip address
should be resolved.  When fronting the application with a webserver or
load balancer, the ip address may be the server and not be the user.

E.g. consider the IP address is passed as the header: "X-Forwarded-For".

```ruby
class User
  devise :trackable

  protected
    def extract_ip_from(request)
      request.headers["X-Forwarded-For"]
    end
end
```

See discussion in #4540 and https://groups.google.com/forum/#!topic/plataformatec-devise/aCuFLef16Sk

Edit: updated example based on feedback changes (#get_ip => #extract_ip_from)

### DIFF
--- a/lib/devise/models/trackable.rb
+++ b/lib/devise/models/trackable.rb
@@ -22,7 +22,7 @@ module Devise
         self.last_sign_in_at     = old_current || new_current
         self.current_sign_in_at  = new_current
 
-        old_current, new_current = self.current_sign_in_ip, request.remote_ip
+        old_current, new_current = self.current_sign_in_ip, extract_ip_from(request)
         self.last_sign_in_ip     = old_current || new_current
         self.current_sign_in_ip  = new_current
 
@@ -39,6 +39,13 @@ module Devise
         update_tracked_fields(request)
         save(validate: false)
       end
+
+      protected
+
+      def extract_ip_from(request)
+        request.remote_ip
+      end
+
     end
   end
 end
--- a/test/models/trackable_test.rb
+++ b/test/models/trackable_test.rb
@@ -59,4 +59,22 @@ class TrackableTest < ActiveSupport::TestCase
 
     assert_not user.update_tracked_fields!(request)
   end
+
+  test 'extract_ip_from should be overridable' do
+    class UserWithOverride < User
+      protected
+        def extract_ip_from(request)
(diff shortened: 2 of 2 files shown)


lib/devise/models/trackable.rb
```diff
@@ -34,6 +34,13 @@ def update_tracked_fields!(request)
         update_tracked_fields(request)
         save
       end
+
+      protected
+
+        def get_ip(request)
```
tegon: Thanks for the pull request @mckramer!

I have some minor comments, they're more about code style:

- Leave the method definition at the same indentation level as the `protected` keyword;
- Change the method name to `extract_ip_from`

mckramer: Thanks for the feedback.  I made the changes.

tegon:
@mckramer Thanks!
Can you please do a rebase with `master`. I had to push a bug fix last week that also changed some code in the `Trackable` module, which apparently caused a conflict here.

mckramer:
@tegon, rebased.

tegon:
@mckramer Thanks!

lib/devise/models/trackable.rb
kuzminvn2010: > Thanks for the feedback. I made the changes.

Why did you choose to use 'protected' method instead of 'private' ?

tegon: @kuzminvn2010 The idea was to able to override this method.

mechnicov: Hi @tegon
But you can also override the private method.  No worse than protected.

```
module A
  def public_a_in_a
    a
  end

  private

  def a
    1
  end
end

class B
  include A
  
  def public_a_in_b
    a
  end

  private

  def a
    super + 1
  end
end

B.new.public_a_in_a #=> 2
B.new.public_a_in_b #=> 2
```

mechnicov: @tegon ping

tegon: Even though you're able to, it shows our intention. If we had chosen to make it `private`, that would mean that is a method people should be careful when using it, because it can change in patch and minor versions.

PR #4779, opened 2018-02-22
Allow test rails 6

Fudoshiki:


rafaelfranca:
Rails 6 is not released yet, why would we support it right now?

Fudoshiki:
We should not support it, but allowed to test with it

rafaelfranca:
The way it is right now it support it. If we change to `< 6.0` we will allow to test without supporting it.

PR #4784, opened 2018-02-26
Fixes #4783 - preserve content_type for unauthenticated test responses

### LINKED ISSUE #4783: In Rails4, unauthenticated controller tests lost their content_type
## Environment

- Ruby 2.4.1
- Rails 4.2.8
- Devise 4.4.1

## Current behavior

In Rails 4, controller tests which trigger the failure app lose their content_type (`response.content_type` is nil).

````
test "returns the content type of a failure app" do
  get :index, params: { format: :xml }
  assert response.content_type.include?('application/xml') # <= content_type is nil!
end
````

This started with bb44d42 (Devise 4.2), which changed how `Devise::Test::ControllerHelpers` simulates failure responses.

Previously ControllerHelpers called `render` with a content-type. Now, values are set on `@controller.response` manually but content_type is skipped.

This doesn't impact Rails5, as [`ActionDispatch::Response#content_type`](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/response.rb#L246) parses the content-type header on demand. But in Rails4, without a call to [#content_type=](https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_dispatch/http/response.rb#L62) it will remain nil.

## Expected behavior

Controller tests which trigger the failure app should be able to inspect `response.content_type`. It should match the content type of the original request.

gmcnaughton:
In Rails 4, for unauthenticated controller tests which trigger the failure app, ensure that the simulated failure response includes a content_type.

This works in Rails5, which parses the content-type header on-demand, but not in Rails4 which requires setting the response's content_type explicitly.

Fixes #4783

### DIFF
--- a/lib/devise/test/controller_helpers.rb
+++ b/lib/devise/test/controller_helpers.rb
@@ -139,6 +139,7 @@ module Devise
 
           status, headers, response = Devise.warden_config[:failure_app].call(env).to_a
           @controller.response.headers.merge!(headers)
+          @controller.response.content_type = headers["Content-Type"] unless Devise::Test.rails5?
           @controller.status = status
           @controller.response.body = response.body
           nil # causes process return @response
--- a/test/test/controller_helpers_test.rb
+++ b/test/test/controller_helpers_test.rb
@@ -100,6 +100,11 @@ class TestControllerHelpersTest < Devise::ControllerTestCase
     assert_equal response.body, "<html><body>You are being <a href=\"http://test.host/users/sign_in\">redirected</a>.</body></html>"
   end
 
+  test "returns the content type of a failure app" do
+    get :index, params: { format: :xml }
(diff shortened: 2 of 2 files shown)


sbleon:
It seems like something might be missing from this commit. Making an unauthenticated controller request now results in `NoMethodError: undefined method `rails5?' for Devise::Test:Module`

Where is the `rails5?` method defined? I can't find it.

Thanks!

tegon:
@sbleon Yeah, this method is defined in our test suite's rails app and it shouldn't be used outside of it. I've already opened a PR to fix it: https://github.com/plataformatec/devise/pull/4809

sbleon:
Thanks for the explanation and the link, Leonardo!
- Leon

PR #4705, opened 2017-11-23
Yield resource before saving in RegistrationsController

vincentwoo:
The golden-path way of extending Devise functionality is to subclass things like RegistrationController and call super with a block. However, if you try something like:

```ruby
class RegistrationsController < Devise::RegistrationsController
  def create
    super { |user| user.skip_confirmation! if some_reason? }
   end
end
```

This doesn't actually prevent the confirm email from being sent, because RegistrationsController always saves before yielding the resource to the block.

PR #4785, opened 2018-02-27
Add #resource_validation_context on Devise::RegistrationsController a…

toommz:
…s POC

As a side feature of configuring permitted params in Devise::RegistrationsController, it might be a valid use case to be able to configure the validation context on resource registration.

Linked PR is just a POC. If you show some interest in this feature, I would be glad to investigate more in code quality (and adding tests).

Possibly related to #4705

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -16,7 +16,7 @@ class Devise::RegistrationsController < DeviseController
   def create
     build_resource(sign_up_params)
 
-    resource.save(context: resource_validation_context)
+    resource.save(context: resource_validation_context_on_create)
     yield resource if block_given?
     if resource.persisted?
       if resource.active_for_authentication?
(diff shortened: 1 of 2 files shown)


tegon:
Seems reasonable, but I think we might include the action on the validation context name, because people might want a different one to the update action, for example 🤔

toommz:
You’re right. I might even implement it on every controller that performs persistence on resources in order to keep it consistent across the gem.

tegon:
Hey @toommz, just checking with you to see if you're still going to work on this. Otherwise, I might leave it for someone to grab on Hacktoberfest.

toommz:
Hi, sorry for the late response. You can assign it to the Hacktoberfest
feature list as I’m not working on it at the moment.

Le ven. 28 sept. 2018 à 23:01, Leonardo Tegon <notifications@github.com> a
écrit :

toommz:
@tegon Don't know if some work has been done on this PR during Hacktoberfest. I found that applying a context on update was more complicated than on create. I think that merging the context in the params is far from an ideal implementation, but I was thinking about backward compatibility for people using update_resource outside of Devise::RegistrationsController.

Otherwise I can be more drastic in the changes for the sake of code cleanliness.

feliperenan:
Hi @toommz

I also think that is a good idea since it would make Devise more flexible and allow developers to validate the resource according to the context. But I don't know if we should add this feature to Devise right now because I don't see much use case for since you can override the `create` or `update_resource` methods case you want something different.

I think we should close it for now and reopen case people ask for this feature in the future.

Thanks for the pull-request

PR #4787, opened 2018-02-27
Fix error when params is not a hash

b0nn1e:
Hello!

When we send to devise controller string or nil instead hash in params we have error **undefined method permit' for "string":String**
This crash Rails app with 500 error and creates unnecessary records in errbit.

Example clean Rails app with Devise - https://shrouded-brook-83924.herokuapp.com
Course code -  https://github.com/b0nn1e/devise_test

**Normal case:**
`/users/sign_in?user[email]=test@example.com`
**Cases with error:** 
`/users/sign_in?user=test@example.com`
`/users/sign_in?user`

Link to reproduce bug:
-  https://shrouded-brook-83924.herokuapp.com/users/sign_in?user=string
-  https://shrouded-brook-83924.herokuapp.com/users/sign_in?user

rails 5.1.5
devise 4.4.1
ruby 2.4.2

### DIFF
--- a/lib/devise/parameter_sanitizer.rb
+++ b/lib/devise/parameter_sanitizer.rb
@@ -135,7 +135,19 @@ module Devise
     end
 
     def default_params
-      @params.fetch(@resource_name, {})
+      if hashable_resource_params?
+        @params.fetch(@resource_name)
+      else
+        empty_params
+      end
+    end
+
+    def hashable_resource_params?
+      @params[@resource_name].respond_to?(:permit)
+    end
+
+    def empty_params
+      ActionController::Parameters.new({})
     end
 
     def permit_keys(parameters, keys)
--- a/test/parameter_sanitizer_test.rb
+++ b/test/parameter_sanitizer_test.rb
@@ -16,6 +16,34 @@ class ParameterSanitizerTest < ActiveSupport::TestCase
     assert_equal({ 'email' => 'jose' }, sanitized)
   end
 
+  test 'permits empty params when received not a hash' do
+    sanitizer = sanitizer({ 'user' => 'string' })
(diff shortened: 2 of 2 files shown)


lib/devise/parameter_sanitizer.rb
```diff
@@ -135,7 +135,19 @@ def cast_to_hash(params)
...
-      @params.fetch(@resource_name, {})
+      if hashable_resource_params?
+        @params.fetch(@resource_name)
+      else
+        empty_params
+      end
+    end
+
+    def hashable_resource_params?
+      @params.to_unsafe_h[@resource_name].is_a?(Hash)
```
rafaelfranca: We don't need to call  `to_unsafe_h` we can just `@params[@resource_name].respond_to?(:permit)`

b0nn1e: Good idea, thanks.
Fix it.

tegon:
@b0nn1e Thanks for your pull request. Can you explain what's your use case for this?

In your request example, you're sending the email value inside a `user` parameter, and you're also not sending a password.

rafaelfranca:
The problem this is solving is that instead of a 422 error because the information sent was not correct it is returning a 500 error because the permit method is being called in a string.

b0nn1e:
@tegon Its not usable case, its example for reproduce error. 
But i think it's incorrect behavior when any user can raise 500 error in my rails app.

tegon:
Oh ok, I get it now. When I saw the description I thought this PR was adding support to it, sorry about that 😄

b0nn1e:
@tegon I did not exactly explain. Sorry for my English :)

b0nn1e:
@rafaelfranca Can you merge, please?

Issue #4536, opened 2017-05-15
Resend confirmation is allowing confirmed users to login


luismfonseca:
If we set `config.confirm_within = 1.days` and `config.allow_unconfirmed_access_for = 1.days`, then a user can just trigger a `resend_confirmation_instructions` which will cause the update field `confirmed_sent_at` to be updated.

And since that gets set to `Time.now.utc`, the user is now able to login for the next 24h, without confirming the email address.

The natural response to this would be to bump `confirm_within`, but that does not solve the actual problem, it only defers it.

PS: I've searched for similar tickets and I haven't found any, but feel free to close this if a ticket already existed.

domangi:
Hi @luismfonseca ,
actually devise uses the same field, confirmation_sent_at timestamp, to check if a confirmation token is still valid (confirm_within) and to check if a user can authenticate without confirmation (allow_unconfirmed_access_for).

Just not updating the confirmation_sent_at timestamp when a new confirmation email is requested would not work because a new confirmation token would be considered always invalid.

The only way I see to limit the unconfirmed access to the period after the first confirmation is sent, is to add another timestamp field first_confirmation_sent_at, to be used just to check how many days passed after the first confirmation was sent.

I have implemented this solution here https://github.com/domangi/devise/tree/allow-unconfirmed-access-only-for-one-period

### Test Code

```
test 'should be not active when first confirmation sent is overpast' do
    swap Devise, allow_unconfirmed_access_for: 5.days, confirm_within: 5.days do
      Devise.allow_unconfirmed_access_for = 5.days
      
      user = create_user
      user.update_attribute(:confirmation_sent_at, 7.days.ago)
      old_confirmation_sent_at = user.confirmation_sent_at
      old_token = user.confirmation_token
      refute user.active_for_authentication?

      user = User.find(user.id)
      user.resend_confirmation_instructions
     
      assert_equal old_confirmation_sent_at, user.confirmation_sent_at
      assert_not_equal old_token, user.confirmation_token
      
      refute user.active_for_authentication?
    end
  end
```

The method that checks if unconfirmed access is allowed becomes

```
def confirmation_period_valid?
    self.class.allow_unconfirmed_access_for.nil? || 
    (first_confirmation_sent_at && 
     first_confirmation_sent_at.utc >= self.class.allow_unconfirmed_access_for.ago)
end
```

Where first_confirmation_sent_at is set only the first time a confirmation is send.

[...]

PR #4792, opened 2018-03-02
confirmation_period_valid? method logic

elgolpe:
Activation without confirmation should be checked using created_at, because checking by confirmation_sent_at creates bugs which allows users to signin after every resend of email confirmation.

Additionaly, I changed 'confirmation_period_valid?' method name to more precious 'allow_unconfirmed_access?'

### DIFF
--- a/test/models/confirmable_test.rb
+++ b/test/models/confirmable_test.rb
@@ -202,7 +202,7 @@ class ConfirmableTest < ActiveSupport::TestCase
   test 'confirm time should fallback to devise confirm in default configuration' do
     swap Devise, allow_unconfirmed_access_for: 1.day do
       user = create_user
-      user.confirmation_sent_at = 2.days.ago
+      user.created_at = 2.days.ago
       refute user.active_for_authentication?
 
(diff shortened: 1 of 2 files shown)


test/models/confirmable_test.rb
```diff
@@ -215,10 +215,10 @@ def setup
       Devise.allow_unconfirmed_access_for = 5.days
       user = create_user
 
-      user.confirmation_sent_at = 4.days.ago
+      user.created_at = 4.days.ago
       assert user.active_for_authentication?
 
-      user.confirmation_sent_at = 5.days.ago
+      user.created_at = 5.days.ago
```
charlesoft: I see an opportunity to refactor this test here. The expected behavior here is to check if the user is active to authentication as it was done in line 219. But below, we test a behavior when the user is not active for authentication. Wdyt about creating a separated test for this case?

elgolpe:
Yes, I will do it

tegon:
Maybe this is also related to #4536 🤔

PR #4793, opened 2018-03-02
Passing :scope option to uniqueness validator of :email

one-more-alex:
Uniqueness validator of :email must have :scope option setting to solve many difficulties like for:
- per subdomain auth;
- soft deletion and paranoid like support;
- even example from https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-sign-in-using-their-username-or-email-address.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -118,6 +118,10 @@ module Devise
   mattr_accessor :password_length
   @@password_length = 6..128
 
+  # Option :scope for Uniqueness Validator of email
+  mattr_accessor :email_scope
+  @@email_scope = nil
+
   # The time the user will be remembered without asking for credentials again.
   mattr_accessor :remember_for
   @@remember_for = 2.weeks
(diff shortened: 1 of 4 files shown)


one-more-alex:
I can add some tests, but I can't do t at near time. Maybe at weekend...
What do you think about some wider solution like following?
<pre>
validates_uniqueness_of :email, {allow_blank: true, if: :will_save_change_to_email?}.merge(other_options_scope_case_etc)
</pre>

To be able add arbitrary options for validator.

tegon:
No problem, you can do when you're able to.

I think your suggestion is a good idea, but since we already have the `email_regexp` setting, I think it's ok to continue with this approach for now.
Maybe we can introduce the hash of options in a major release.

tegon:
@one-more-alex Just to be sure, are you still going to work on this or can I assign it to someone else?

tegon:
@one-more-alex No problem, I just wanted to know if you still wanted to work on this.

> I made an attempt to add tests, but not get deep into tests structure yet.

If you need help, let me know. You can add a WIP commit here with your progress so far so I can help, or send your questions.

> I not understood if it is possible to edit this PR commits somehow

You mean how to remove the commit from https://github.com/plataformatec/devise/pull/4794?

tegon:

Thank you!

PR #4794, opened 2018-03-02
Generator templates indexes condition correction

one-more-alex:
Indexes for tokens should have WHERE ...token IS NOT NULL construction.
Else we are admitted to create only one user.
Unfortunatelly, in worse cases people just remove these indexes...
Initially faced with it on MS SQL server (not checked other adapters
        though).
Inspired by https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/153

Probably need to add test case for generated migrations with tokens

### DIFF
--- a/lib/generators/active_record/templates/migration.rb
+++ b/lib/generators/active_record/templates/migration.rb
@@ -13,8 +13,8 @@ class DeviseCreate<%= table_name.camelize %> < ActiveRecord::Migration<%= migrat
     end
 
     add_index :<%= table_name %>, :email,                unique: true
-    add_index :<%= table_name %>, :reset_password_token, unique: true
-    # add_index :<%= table_name %>, :confirmation_token,   unique: true
-    # add_index :<%= table_name %>, :unlock_token,         unique: true
+    add_index :<%= table_name %>, :reset_password_token, unique: true, where: '([reset_password_token] IS NOT NULL)'
+    # add_index :<%= table_name %>, :confirmation_token,   unique: true, where: '([confirmation_token] IS NOT NULL)'
+    # add_index :<%= table_name %>, :unlock_token,         unique: true, where: '([unlock_token] IS NOT NULL)'
   end
 end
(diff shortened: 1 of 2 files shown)


tegon:
Hey @one-more-alex, thanks for your pull request.

I don't get it why we need this though. You said

> Else we are admitted to create only one user.

Those files are created using a generator, which assumes you are creating another table or new columns to a existing table. But I don't think we ever intended to cover the case where these columns already existed in a table but the indexes didn't exist - which is the case I've understood you PR is targeting, but please correct me if I'm wrong.

Anyway, can you explain your use case for this?

one-more-alex:
Thank you for feedback.
Sorry for unclear description. Now I reread it and understood only form second attempt ).

First of all, it is only MS SQL specific issue. I checked in MySQL - no issue.

The issue may be reproduced by:
- install new rails and generate device files/migrations;
- we have table index with reset_password_token 'unique: true';
- create one user and see all is fine;
- attempt to create second user and see fail like 
<pre>
Cannot insert duplicate key row in object 'dbo.users' with unique index 'reset_password_token'. The duplicate key value is (<NULL>).
</pre>

So newbies where disappointed here and finds receipts on stack overflow like "just drop the index".

I now realize that it is Microsoft only bug or violation of standards (see references in https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/153#issue-2460030). Not sure if it possible to have where condition add only for MS SQL Server.
Personally I reproduced the issue with SQL Server Express 2017.

So it will be unfair to add index condition commonly. PR need to be targeted on engine type if you will decide to merge it.
Or maybe even better to do some monkey patch from activerecord-sqlserver-adapter only.
But... even this may not solve the issue, because schema may be generated for normal engine and then be applied on MS SQL Server.
Not sure what to do here. Probably do fat warnings in activerecord-sqlserver-adapter instead.
Please decide, because I have not so much experience.
PR likely to be rejected.

tegon:
Now I got it, thanks for your explanation.

Since it's an MS SQL issue only, it doesn't make sense for devise to do some hack to support this. Ideally, this can be handled at the `activerecord-sqlserver-adapter` gem.
But since people are facing this when using devise, I think a Wiki page explaining this might be helpful. 
If you're willing to, I suggest you [create](https://github.com/plataformatec/devise/wiki/_new) a new Wiki page that explains the issue and the solution you used. Then maybe this can help others in the future.

Thanks!

s0nspark:
I added a Wiki page for this - feel free to edit if it is not clear ;-)

https://github.com/plataformatec/devise/wiki/SQL-Server-token-column-indexes-and-null-values

PR #4795, opened 2018-03-04
Use `update` instead of `update_attributes`

koic:
`update_attributes` and `update_attributes!` will be deprecated in Rails 6.

Refer https://github.com/rails/rails/commit/5645149d3a27054450bd1130ff5715504638a5f5


tegon:
Thanks, @koic!

Issue #4790, opened 2018-03-01
Model's validations are called on EVERY user's request


dhiemstra:
https://github.com/plataformatec/devise/pull/4674 made a breaking change to the `update_tracked_fields!` method by removing `validate: false` from the `save()` call.

Basically this results in all the validations being called on every requested. Some companies run extended validations to ensure the validity of the email, phone number or vat number and often a 3rd party service / api is used to check this.

For us this meant an unexpected bill from Mailgun of more than $3500 because a massive amount of request were done in the last month to their validations API.

When looking at the original issue why this change was introduced I also don't think this is the proper fix. Why is `update_tracked_fields!` (with bang) even called when the user doesn't exist in the database? And, why is it even attempting to update tracked fields for a `new_user?`.

I suggest adding back the `validate: false` asap and fixing the original issue otherwise.

tegon:
When we merged #4674, we didn't know that it would change the contract of `Trackable` that much, that was definitely a mistake, so I'm sorry for this.
I'll work on another solution for the problem since it's causing that much trouble. I haven't before because the other issues (#4742 for example) could be solved it by making small changes in the validations. But you have a valid point: it seems weird that we're calling the `update_tracked_fields!` on a record that isn't persisted yet.

I'd advise that those calls to external services would better be done in places where you can control how and when they're going to be executed, especially if you're paying for each request. Model validations can be called in many situations we aren't aware of and cause surprises such as this one.

Anyway, I'm definitely going to work on another way of doing this, so please downgrade for now.

Thanks for your report and sorry again for the trouble.

PR #4796, opened 2018-03-05
Check if the record is persisted in `update_tracked_fields!`

tegon:
In some cases, invalid records could be created during the signup process because we were calling `save(validate: false)` inside the `update_tracked_fields!` method. See https://github.com/plataformatec/devise/issues/4673 for more information.
This was fixed on https://github.com/plataformatec/devise/pull/4674 by calling `save` directly, but it caused some trouble and confusion since it changed Devise's behavior significantly.
We talked about on https://github.com/plataformatec/devise/issues/4790 and it doesn't even make sense to call `save` on an object that isn't persisted yet, so I've added a guard clause to the `update_tracked_fields!` method.

### DIFF
--- a/lib/devise/models/trackable.rb
+++ b/lib/devise/models/trackable.rb
@@ -31,8 +31,13 @@ module Devise
       end
 
       def update_tracked_fields!(request)
+        # We have to check if the user is already persisted before running
+        # `save` here because invalid users can be saved if we don't.
+        # See https://github.com/plataformatec/devise/issues/4673 for more details.
+        return if new_record?
+
         update_tracked_fields(request)
-        save
+        save(validate: false)
       end
     end
   end
--- a/test/integration/authenticatable_test.rb
+++ b/test/integration/authenticatable_test.rb
@@ -3,6 +3,12 @@
 require 'test_helper'
 
 class AuthenticationSanityTest < Devise::IntegrationTest
+  test 'sign in should not run model validations' do
+    sign_in_as_user
+
+    refute User.validations_performed
+  end
+
   test 'home should be accessible without sign in' do
     visit '/'
     assert_response :success
--- a/test/integration/trackable_test.rb
+++ b/test/integration/trackable_test.rb
@@ -3,6 +3,11 @@
 require 'test_helper'
 
 class TrackableHooksTest < Devise::IntegrationTest
+  test "trackable should not run model validations" do
+    sign_in_as_user
+
+    refute User.validations_performed
+  end
 
   test "current and last sign in timestamps are updated on each sign in" do
     user = create_user
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -13,7 +13,7 @@ PATH
     devise (4.4.1)
       bcrypt (~> 3.0)
       orm_adapter (~> 0.1)
-      railties (>= 4.1.0, < 5.2)
+      railties (>= 4.1.0, < 6.0)
       responders
       warden (~> 1.2.3)
 
--- a/test/integration/http_authenticatable_test.rb
+++ b/test/integration/http_authenticatable_test.rb
@@ -3,6 +3,12 @@
 require 'test_helper'
 
 class HttpAuthenticationTest < Devise::IntegrationTest
+  test 'sign in with HTTP should not run model validations' do
+    sign_in_as_new_user_with_http
+
+    refute User.validations_performed
+  end
+
   test 'handles unverified requests gets rid of caches but continues signed in' do
     swap ApplicationController, allow_forgery_protection: true do
       create_user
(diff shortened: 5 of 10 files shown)


lib/devise/models/trackable.rb
```diff
@@ -31,8 +31,13 @@ def update_tracked_fields(request)
       end
 
       def update_tracked_fields!(request)
+        # We have to check if the user is already persisted before running
+        # `save` here because invalid users can be saved if we don't.
+        # See https://github.com/plataformatec/devise/issues/4673 for more details.
+        return if new_record?
+
         update_tracked_fields(request)
```
rafaelfranca: Should not the `update_tracked_fields` still be called when it is a new record?

tegon: When the signup succeeds the record would already be persisted here. This is being called from a Warden's `after_set_user` [callback](https://github.com/plataformatec/devise/blob/master/lib/devise/hooks/trackable.rb#L9). The callback is triggered when we call `set_user` from the `sign_in` [helper](https://github.com/plataformatec/devise/blob/master/lib/devise/controllers/sign_in_out.rb#L50), which is called after the record is created in [registrations controller](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/registrations_controller.rb#L24).

rafaelfranca: Makes sense. Is there any way we can test this to make sure it doesn't happen anymore?

tegon: The test introduced in https://github.com/plataformatec/devise/pull/4674/files#diff-8aa58f219abef0280c8857f9161c087bR42 is still passing, so I think we're good to go.

rafaelfranca: But how about the tests to make sure that every request don't run validations?

tegon: I see, I'm not sure of how we should test this 🤔 
Do you think we should check here if the validations were run or in the sessions controller?

rafaelfranca: One way to test this is write an integration test, add a validation to the model that sets a global value in the class if the validation is run. This way we can test if a request to any action sets this global value.

charlesoft: I think it's worth to write a unit test to the `Trackable` module to check if the record is saved not running the validations 🤔 .

test/integration/authenticatable_test.rb
```diff
@@ -3,6 +3,12 @@
 require 'test_helper'
 
 class AuthenticationSanityTest < Devise::IntegrationTest
+  test 'sign in should not run model validations' do
+    sign_in_as_user
+
+    refute $validations_performed
```
rafaelfranca: When I said a global I was thinking more in a class variable inside the model

tegon: I get it now 😅 
I've changed to use a class variable

test/integration/trackable_test.rb
```diff
@@ -3,6 +3,12 @@
 require 'test_helper'
 
 class TrackableHooksTest < Devise::IntegrationTest
+  test "sign in with HTTP should not run model validations" do
+    create_user
```
charlesoft: Is `create_user` already called when we use `sign_in_as_user ` method? 🤔

carlosantoniodasilva:
Hey folks, taking a look at this, I'm not sure I see how/why a record would get to this call via a warden hook while not being persisted. This must be something custom someone is doing along the way, since as you [pointed out](https://github.com/heartcombo/devise/pull/4796#discussion_r172348087) @tegon, the registrations controller calls sign up which will in turn delegate to warden's set_user eventually, but it calls `save` before any of that: https://github.com/heartcombo/devise/blob/e8e0c275999dd98150197cab03acb5509cb16b6a/app/controllers/devise/registrations_controller.rb#L19-L24.

So, unless we're talking about some custom code in the registration that's not calling save and/or checking for the errors before calling `set_user` (or `sign_up`), I'm not sure I can see a path for us to save a record not being persisted here, thus I'm not sure we need the `new_record?` check in trackable.

If I remove that check, only the model tests fail, the integration ones don't, which is more proof that the workflow as a whole is working I think.

tegon:
@carlosantoniodasilva It has been a long time, but I think you're right, it can't happen with Devise's default setup. If I recall correctly, we ended up adding this patch because it was affecting multiple applications out there. I believe this could happen if the apps had other validations on their models apart from the Devise ones. Since the trackable module runs apart from registrations, that's the way we prevented it from happening in there.

carlosantoniodasilva:
Makes sense. I still think the contract should be that you're not running the hooks for an unsaved model, I'll see about that in a future major version or something. Thanks!

PR #4797, opened 2018-03-08
[bugfix] [refactoring] Sanitize parameters in find_or_initialize_with_errors

rlue:
### Context

The `Authenticatable::find_or_initialize_with_errors` class method loads a record based on given attributes, or instantiates a new one if it does not exist. It depends on `ParameterFilter` to filter/sanitize the input attributes as appropriate (_e.g.,_ downcasing case-insensitive keys or stripping whitespace).

### Problem

In its current form, it only does this parameter filtering when loading an existing record (by way of `::find_first_by_auth_conditions`) — that is, it does not filter input attributes when instantiating a new resource.

### Solution

This PR ensures that parameters are filtered for new records as well. I also took the liberty of doing some refactoring.

### DIFF
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -276,28 +276,20 @@ module Devise
 
         # Find or initialize a record with group of attributes based on a list of required attributes.
         def find_or_initialize_with_errors(required_attributes, attributes, error=:invalid) #:nodoc:
-          attributes = if attributes.respond_to? :permit!
-            attributes.slice(*required_attributes).permit!.to_h.with_indifferent_access
-          else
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Can you write a test for what doesn't work today?

rlue:
Done. This is my first experience with minitest, so if you have any comments on code style (_e.g.,_ lines are too long, or comments are unnecessary), please let me know.

rlue:
@rafaelfranca, is there anything keeping this from being merged? There's a bug on another project I'd like to file a PR for that depends on this one.

tegon:
Your changes look good, thanks @rlue!

PR #4809, opened 2018-03-15
Fix `Devise::Test::ControllerHelpers`

### LINKED ISSUE #4808: undefined method `rails5?' for Devise::Test:Module in 4.4.2
## Environment

- Ruby 2.3.5
- Rails 5.1.5
- Devise 4.4.2

## Current behavior
Controller specs where the user is not authenticated fail with this exception:
```
undefined method `rails5?' for Devise::Test:Module in 4.4.2
```
An example stack trace is:
```
NoMethodError:
       undefined method `rails5?' for Devise::Test:Module
     # /Users/moncefbelyamani/.gem/ruby/2.3.5/gems/devise-4.4.2/lib/devise/test/controller_helpers.rb:142:in `_process_unauthenticated'
     # /Users/moncefbelyamani/.gem/ruby/2.3.5/gems/devise-4.4.2/lib/devise/test/controller_helpers.rb:118:in `_catch_warden'
     # /Users/moncefbelyamani/.gem/ruby/2.3.5/gems/devise-4.4.2/lib/devise/test/controller_helpers.rb:35:in `process'
     # /Users/moncefbelyamani/.gem/ruby/2.3.5/gems/actionpack-5.1.5/lib/action_controller/test_case.rb:392:in `get'
     # /Users/moncefbelyamani/.gem/ruby/2.3.5/gems/rails-controller-testing-1.0.2/lib/rails/controller/testing/integration.rb:12:in `block (2 levels) in <module:Integration>'
     # ./spec/controllers/application_controller_spec.rb:103:in `block (4 levels) in <top (required)>'
```

## Expected behavior
Tests that used to pass with Devise 4.4.1 continue to pass when upgrading Devise to 4.4.2

tegon:
This was broken on https://github.com/plataformatec/devise/commit/3e23371b01443d6b8bdbfcbdcd33d2883847e39d#diff-bafaaa60fc003e648eb4981c9add523eR142
because of a call to the method `Devise::Test.rails5?` which is only defined in our test suite's rails app https://github.com/plataformatec/devise/blob/a45bbe1783e4b46af15825a0bb2eb5917a9b8726/test/rails_app/config/boot.rb#L19
This commits checks for the rails version directly.

Since this works inside our test suite, we haven't noticed this bug. It only happens when other projects require the controller helpers outside devise. I'm not sure how we could add an automated test for this, I accept suggestions. But I believe a better solution would be to change this class namespace for our test suite's rails app so we don't face this kind of confusion anymore.

Fixes #4808

### DIFF
--- a/lib/devise/test/controller_helpers.rb
+++ b/lib/devise/test/controller_helpers.rb
@@ -139,7 +139,7 @@ module Devise
 
           status, headers, response = Devise.warden_config[:failure_app].call(env).to_a
           @controller.response.headers.merge!(headers)
-          @controller.response.content_type = headers["Content-Type"] unless Devise::Test.rails5?
+          @controller.response.content_type = headers["Content-Type"] unless Rails.version.start_with?('5')
           @controller.status = status
           @controller.response.body = response.body
           nil # causes process return @response


gmcnaughton:
Oh no, I'm sorry about that! Thanks for the fix ❤️

NARKOZ:
🚢

PR #4812, opened 2018-03-23
cleanup in ToC

hiphapis:
these contents are not exists, so i did remove it.
but if you will add these contents, just close my PR

ToC updated commit: https://github.com/plataformatec/devise/commit/42b68d53a707dca655e27ab63c6c3f0b583c8bbe#diff-04c6e90faac2675aa89e2176d2eec7d8
@shivabhusal @tegon


tegon:
Great catch @hiphapis! 👏

PR #4818, opened 2018-03-26
Use before_action instead of before_filter

edenthecat:
The integration test template was using the deprecated `before_filter` rather than `before_action`


tegon:
Thanks, @edenthecat!

PR #4831, opened 2018-04-04
Improve passwords autocomplete

gssbzn:
Use `autocomplete="new-password"` or `autocomplete="current-password"` to better signal browsers and password managers how to handle the password fields.

This feature is especially useful for “change your password” and “new user” forms

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ### Unreleased
 
+* enhancements
+  * Add `autocomplete="new-password"` to new password fields (by @gssbzn)
+  * Add `autocomplete="current-password"` to current password fields (by @gssbzn)
+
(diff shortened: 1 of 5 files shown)


tegon:
Thanks for your pull request, @gssbzn!

Personally, I don't know those autocomplete options. Can you explain what the differences from a user perspective are? 
From a quick Google search, it seems like the behavior is similar to the `off` option, so I'm not sure if there are meaningful advantages.

gssbzn:
@tegon like you say they are similar to `off`, the main difference will be to users using password managers, and specially to pages where you need to change your password, with fields for new and current password in the same form.

This prevents cases where browsers or password managers will try to auto complete the new password with existing/stored password (really common case) and let's password managers suggest new passwords or autocomplete where needed

Issue #4356, opened 2016-11-18
Inconsistent checking of navigational format


BookOfGreg:
Why does [after_sending_reset_password_instructions_path_for](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/passwords_controller.rb#L57) check [is_navigational_format?](https://github.com/plataformatec/devise/blob/master/lib/devise/controllers/helpers.rb#L262)  but [after_sign_up_path_for](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/registrations_controller.rb#L112) does not?

The behaviour I expected is in [an example repo](https://github.com/BookOfGreg/devise-navigational-example).
This repo makes use of JQuery_UJS and turbolinks with [create.js.erb](https://github.com/BookOfGreg/devise-navigational-example/blob/master/app/views/devise/registrations/create.js.erb) rendering forms.
There are two failing tests based around the [password controller](https://github.com/BookOfGreg/devise-navigational-example/blob/master/test/controllers/user_password_test.rb#L4), one is to do with using the wrong resource which I will fix but the other failing test is devise returning nil for a named path for is_navigational_format not believing :js is real.

There is a counterpoint test for [registration controller](https://github.com/BookOfGreg/devise-navigational-example/blob/master/test/controllers/user_registration_test.rb#L4) that passes as it does not bother to check is_navigational_format.

I'm not looking for devise to change it's behaviour but I would like to understand why these two similar controllers are different in this way.

tegon:
Hello @BookOfGreg, thanks for your report.
Unfortunately, I don't have an exact answer to your question. What you said makes sense to me, and the [commit](https://github.com/plataformatec/devise/commit/43f7cf0f1283fd2dc526ebbe7cf17528b42f3884) that introduced the checks for `is_navigational_format?` also modified the registrations controller. But somehow, the latest [commit](https://github.com/plataformatec/devise/commit/8c9cf7d67164c956597fed93ae0aecd945bc3247) that touched this code on master is the previous one - without this check.
My guess is that at some point a `git reset` might have removed this code from the registrations controller and it got lost forever.
Anyway, it seems reasonable to me to keep the same behavior. Are you willing to open a pull request for this change?

Thank you!

iorme1:
I made a PR for this issue.   Please let me know if there are any issues with it. Thanks!

PR #4833, opened 2018-04-04
issue 4356 adds is_navigational_format? check to after_sign_up_path_for

iorme1:
Added  is_navigational_format? check to after_sign_up_path_for,  as the issue #4356 requested.  Tests are passing.

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -112,7 +112,7 @@ class Devise::RegistrationsController < DeviseController
   # The path used after sign up. You need to overwrite this method
   # in your own RegistrationsController.
   def after_sign_up_path_for(resource)
-    after_sign_in_path_for(resource)
+    after_sign_in_path_for(resource) if is_navigational_format?
   end
 
   # The path used after sign up for inactive accounts. You need to overwrite


app/controllers/devise/registrations_controller.rb
```diff
@@ -112,7 +112,7 @@ def sign_up(resource_name, resource)
   # The path used after sign up. You need to overwrite this method
   # in your own RegistrationsController.
   def after_sign_up_path_for(resource)
-    after_sign_in_path_for(resource)
+    after_sign_in_path_for(resource) if is_navigational_format?
```
feliperenan: Could we cover this change with tests?

iorme1: I looked up all the current methods in Devise that use 'is_navigational_format?' checks and I couldn't seem to find anything in those tests that specifically incorporate a test for that navigational check. In my previous pull requests for Devise, I wrote my tests based off similar structured methods and their corresponding tests. However, since I cannot find how the other methods using 'is_navigational_format?' are testing it,  I am not sure how to
incorporate additional tests for the method I added it to. Based on the issue conversation, i was under the impression that the method I changed used to have a 'is_navigational_format?' check and that it somehow got deleted by accident from a git reset.  So i simply
added it back to the method.

tegon: Yep, we don't have tests for any of these so I think it's ok for now.

tegon:
Thank you @iorme1! 🎉

PR #4834, opened 2018-04-05
Downcase authentication keys and humanize error message

### LINKED ISSUE #4095: "Invalid Login or password." is grammatically incorrect

grantzau:
To fix #4095

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -103,11 +103,11 @@ module Devise
         options[:scope] = "devise.failure"
         options[:default] = [message]
         auth_keys = scope_class.authentication_keys
-        keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class.human_attribute_name(key) }
+        keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class.human_attribute_name(key).downcase }
         options[:authentication_keys] = keys.join(I18n.translate(:"support.array.words_connector"))
         options = i18n_options(options)
 
-        I18n.t(:"#{scope}.#{message}", options)
+        I18n.t(:"#{scope}.#{message}", options).humanize
       else
         message.to_s
       end
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -106,8 +106,11 @@ module Devise
         keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class.human_attribute_name(key).downcase }
         options[:authentication_keys] = keys.join(I18n.translate(:"support.array.words_connector"))
         options = i18n_options(options)
+        translated_message = I18n.t(:"#{scope}.#{message}", options)
 
-        I18n.t(:"#{scope}.#{message}", options).humanize
+        # only call `#humanize` when the message is `:invalid` to ensure the original format
+        # of other messages - like `:does_not_exist` - is kept.
+        message == :invalid ? translated_message.humanize : translated_message
       else
         message.to_s
       end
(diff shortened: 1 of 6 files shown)


lib/devise/failure_app.rb
```diff
@@ -103,11 +103,11 @@ def i18n_message(default = nil)
         options[:scope] = "devise.failure"
         options[:default] = [message]
         auth_keys = scope_class.authentication_keys
-        keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class.human_attribute_name(key) }
+        keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class.human_attribute_name(key).downcase }
```
feliperenan: Could we add tests for it? 🤔

grantzau: Edited the existing tests. They check, that the downsizing are done. There's not really a test to check the added humanize method though, not sure, if that's necessary?

feliperenan: I would add to ensure that it's the expected behavior.

grantzau:
Great! I will look at it and squash it during the weekend. Was editing
directly on Github from the browser, which wasn't a good idea ..!

fre. 6. apr. 2018 kl. 03.22 skrev Felipe Renan <notifications@github.com>:

test/failure_app_test.rb
```diff
@@ -153,20 +153,30 @@ def call_failure(env_params={})
...
+
+    test 'humanizes the flash message' do
+      call_failure('warden' => OpenStruct.new(message: :invalid))
+      assert_equal @request.flash[:alert], @request.flash[:alert].humanize
+    end
+
     test 'uses custom i18n options' do
       call_failure('warden' => OpenStruct.new(message: :does_not_exist), app: FailureWithI18nOptions)
-      assert_equal 'User Steve does not exist', @request.flash[:alert]
+      assert_equal 'User steve does not exist', @request.flash[:alert]
```
tegon: I'm not sure about this one, I think it should have stayed as it was before. This is a custom message defined here: https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/test/support/locale/en.yml#L5 and we're passing the user's name `Steve` in a capitalized form here: https://github.com/grantzau/devise/blob/cca5a7b8e9f699de4299e001d95e3e269470f47f/test/failure_app_test.rb#L27, which gets lost when we call `#humanize`
```ruby
[3] pry(main)> "User Steve does not exist".humanize
=> "User steve does not exist"
```
I'm not sure what's the best solution here. Maybe we should only call `#humanize` when the message is `:invalid` 🤔

smoyth:
Note that you can control i18n of parameters to failure messages by doing something like this:

In your models:
```
module People
  module Users
    # Overriding the i18n_options method in the Devise FailureApp so that we can have control
    # over how the `authentication_keys` parameter to the messages under devise.failure is translated.
    # By default the `human_attribute_name` of the attribute is used, but those are typically capitalized,
    # while the messages call for a lowercase value.
    # This custom failure app must be activated in the config/initializers/devise.rb file.
    class CustomDeviseFailureApp < Devise::FailureApp
      def i18n_options(options)
        options.merge(authentication_keys: I18n.t("devise.authentication_keys.email"))
      end
    end
  end
end
```

and then in `config/initializers/devise.rb`:

```
  config.warden do |manager|
    manager.failure_app = People::Users::CustomDeviseFailureApp
  end
```

and then in your `devise.yml` file:

```
  devise:
    ...
    authentication_keys:
      email: "email"
```

Ideally Devise would support an optional `authentication_keys` key in the YML file out of the box, so you could have control over these params without a custom failure app, but this works right now and is not particularly hacky. There is support for this method in the tests even, see [here](https://github.com/plataformatec/devise/blob/master/test/failure_app_test.rb#L171).

Hope this helps!

tegon:
@grantzau are you still going to work on this?

Issue #4399, opened 2017-01-18
Redirect to new_password_path if reset_password_token is expired


jeremywadsack:
The current flow of `PasswordController#update` re-renders the `edit` view whenever there is an error resetting the password. But if the error is that the token has expired the message says "Reset password token has expired, please request a new one" — with no instructions on how to do that. It would make more sense to me to have devise redirect this to the `new_password_path` in this case. That way the user has the form in front of them to make that request.

Is there a reason not to do this? If this seems like a reasonable change, I'm happy to open a pull request with the change.

tegon:
Seems reasonable to me, but I think we would have to introduce this in a major version.

victorhazbun:
This is not a simple change.

tegon:
@victorhazbun Can you explain your concerns?

victorhazbun:
@tegon I was looking into the code that handles the redirection/rendering in the controller and I saw that there was a method which sets errors for the reset_password_token https://bit.ly/2pu8vXN my concern is the way the controller responds when the resource has errors, it does not classifies which kind of error is, it always responds with the resource https://bit.ly/2FVHuXj so in order to implement this feature there should be a class that determines what to do with the resource depending on each error (It seems like the only possible error is reset_password_token). So, I believe this is what makes it not so simple, I do not know how you guys do it in this library but I can only think about a different class that has the responsibility to handle the resource errors and depending on the error it return a different action: If errors on reset_password_token then return a redirect to x_path, and so on for each kind of error.

jeremywadsack:
We patched by replacing [line 49](https://github.com/plataformatec/devise/blob/f39c6fd92774cb66f96f546d8d5e8281542b4e78/app/controllers/devise/passwords_controller.rb#L49) with this for Rails 4.2:

```
      if resource.errors[:reset_password_token].any? { |error| error =~ /expired/ }
        redirect_to new_password_path(resource_name), alert: resource.errors.full_messages.join("\n")
      else
        respond_with resource
      end
```

In Rails 5 we can inspect the validator key rather than inspecting the message which is a better i18n approach:

```
      if resource.errors.details[:reset_password_token].any? { |error| error[:error] == :expired }
```

tegon:
Since this is going to be included in a major change, we can support only the Rails 5 version.
@jeremywadsack would you be willing to open a PR for this?

jeremywadsack:
Yeah, I can do that. Might be a week or two, but I'll try to get something soon.

[...]

PR #4837, opened 2018-04-08
When password reset token expires, redirect to new password path

jeremywadsack:
Addresses issue #4399.

The flow of `PasswordController#update` re-rendered the edit view whenever there was an error resetting the password. But if the error was that the token has expired the message said "Reset password token has expired, please request a new one" — with no instructions on how to do that. 

This pull requests changes the behavior such that if the error is an expired password token, devise redirects to the `new_password_path`. That way the user has the form in front of them to make that request.

### DIFF
--- a/app/controllers/devise/passwords_controller.rb
+++ b/app/controllers/devise/passwords_controller.rb
@@ -47,7 +47,12 @@ class Devise::PasswordsController < DeviseController
       respond_with resource, location: after_resetting_password_path_for(resource)
     else
       set_minimum_password_length
-      respond_with resource
+
+      if expired_token_error(resource)
+        redirect_to new_password_path(resource_name), alert: resource.errors.full_messages.join("\n")
+      else
+        respond_with resource
+      end
     end
   end
 
@@ -79,5 +84,12 @@ class Devise::PasswordsController < DeviseController
 
     def translation_scope
       'devise.passwords'
+    end
+
+  private
+    def expired_token_error(resource)
+      (resource.errors.respond_to?(:details) &&
+          resource.errors.details[:reset_password_token].any? { |error| error[:error] == :expired }) ||
+          resource.errors[:reset_password_token].any? { |error| error =~ /expired/ }
     end
 end
--- a/app/controllers/devise/passwords_controller.rb
+++ b/app/controllers/devise/passwords_controller.rb
@@ -49,7 +49,8 @@ class Devise::PasswordsController < DeviseController
       set_minimum_password_length
 
       if expired_token_error(resource)
-        redirect_to new_password_path(resource_name), alert: resource.errors.full_messages.join("\n")
+        alert = resource.errors.full_messages_for(:reset_password_token).join("\n")
+        redirect_to new_password_path(resource_name), alert: alert
       else
         respond_with resource
       end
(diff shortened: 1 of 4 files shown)


app/controllers/devise/passwords_controller.rb
```diff
@@ -46,7 +46,12 @@ def update
       respond_with resource, location: after_resetting_password_path_for(resource)
     else
       set_minimum_password_length
-      respond_with resource
+
+      if expired_token_error(resource)
+        redirect_to new_password_path(resource_name), alert: resource.errors.full_messages.join("\n")
```
tegon: I'm not sure about showing the errors in a flash message. The errors were supposed to be displayed inside `app/views/password/new`, when we call [`devise_error_messages!`](https://github.com/plataformatec/devise/blob/16b3d6d67c7e017d461ea17ed29ea9738dc77e83/app/views/devise/passwords/new.html.erb#L4). Is this not working?

jeremywadsack: How would resource errors be carried through a redirect?

yordis: @jeremywadsack maybe use specific i18n key for this rather than using `full_messages` because it could contains other error messages and the only reason you would go to `new_password_path` for this case is because your token expired.

So probably change it to use `i18n`.

runephilosof: it might be
```
        redirect_to new_password_path(resource_name), alert: resource.errors.full_messages_for(:reset_password_token).join("\n")
```

jeremywadsack:
@runephilosof Thanks for the code suggestion (a long time ago)! @tegon Thanks for the nudge.

I've updated with the requested change. If this looks good I can squash commits. Let me know if there are any other changes you see.

tegon:
@jeremywadsack I would go with the specific I18n key like @yordis suggested. Do you see any downsides in following that approach?

tegon:
@jeremywadsack I meant to create a new I18n key and call it directly (outside `#errors`):
```yml
# config/locales/en.yml
en:
  devise:
    passwords:
      expired_token: "The password recovery link expired. Please request a new one."
```
```rb
# app/controllers/devise/passwords_controller.rb
if expired_token_error(resource)
  redirect_to new_password_path(resource_name), alert: I18n.t("devise.passwords.expired_token")
else
  respond_with resource
end
```

jeremywadsack:
@tegon  Thanks for making that change. Somehow this dropped off my radar in the last two months!

app/controllers/devise/passwords_controller.rb
```diff
@@ -80,4 +85,11 @@ def unlockable?(resource)
     def translation_scope
       'devise.passwords'
     end
+
+  private
+    def expired_token_error(resource)
+      (resource.errors.respond_to?(:details) &&
```
feliperenan: Could we break this check into smaller methods to improve readability? Also, would be good have some documentation explaining how we check whether the token is expired or not.

tegon: I remembered that Devise 5 will drop Rails 4 support, so we don't need the conditional anymore. I guess it's simpler now and self-explanatory. WDYT?

Issue #4816, opened 2018-03-24
Require password for account deletion by default


mnlse:
## Current behavior
Devise requires current password in order to change account, but does not require current password for account deletion.

## Expected behavior
Devise should require current password in order to delete account.

## Manual solution
~~~
def destroy
  if resource.destroy_with_password(params[:user][:current_password])
    flash[:notice] = "Your account has been deleted"
    redirect_to root_path    
  else                       
    flash[:alert] = "Wrong password"
    render :edit, layout: 'application'                                                 
  end  
end
~~~

tegon:
I'm ok with supporting this feature, but we can't require it by default since it isn't backward compatible.
We can add a setting and then check for it in the destroy method.
@mnlse If you're willing to open a pull request for this, we'd be happy to accept it.

Thanks!

iorme1:
I'll be working on a PR for this.

iorme1:
I finished the feature and submitted a PR.

lancecarlson:
@tegon @iorme1 This looks good to me. If you want this to default to off, I would make the generated config commented out and assigned to true. If you want this to default to on, then I would leave the config uncommented, but change its value to true.

collimarco:
I am really interested in this feature. It doesn't make sense that a user cannot change the email, but can destroy the account without password!

Can you merge the PR?

tegon:
@collimarco The PR still needs some fixes, that's why I haven't merged yet.

PR #4840, opened 2018-04-12
Adds feature require_password_to_destroy

iorme1:
This feature addresses issue #4816. I added a setting for requiring a user's current password in order to destroy their account.  This setting is set to false by default.  Please let me know if I went about coding this feature according to your typical structure. I tried to follow Devise's code structure as closely as possible.  All of Devise's tests including my test additions are passing. I also tried this feature out on my own test app with no issues.

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -66,11 +66,23 @@ class Devise::RegistrationsController < DeviseController
 
   # DELETE /resource
   def destroy
-    resource.destroy
-    Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
-    set_flash_message! :notice, :destroyed
-    yield resource if block_given?
-    respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+    if resource.class.require_password_to_destroy
+      resource_destroyed = destroy_resource(resource, account_destroy_params)
+      if resource_destroyed
+        Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+        set_flash_message! :notice, :destroyed
+        yield resource if block_given?
+        respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+      else
+        render :edit
+      end
+    else
+      resource.destroy
+      Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+      set_flash_message! :notice, :destroyed
+      yield resource if block_given?
+      respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+    end
   end
 
   # GET /resource/cancel
@@ -95,6 +107,10 @@ class Devise::RegistrationsController < DeviseController
   # You can overwrite this method in your own RegistrationsController.
   def update_resource(resource, params)
     resource.update_with_password(params)
+  end
+
+  def destroy_resource(resource, params)
+    resource.destroy_with_password(params)
   end
 
   # Build a devise resource passing in the session. Useful to move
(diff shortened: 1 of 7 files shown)


mftaff:
Hey guys!
Any news on this PR?

This is a feature I would love to have in my rails app!

app/views/devise/registrations/edit.html.erb
```diff
@@ -38,6 +38,21 @@
 
 <h3>Cancel my account</h3>
 
-<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?" }, method: :delete %></p>
+<% if resource.class.require_password_to_destroy %>
+  <%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :delete }) do |f| %>
```
tegon: This form is exactly the same as the previous one. This will generate invalid HTML - multiple elements with the same ID.
screen shot 2018-12-04 at 18 41 36

One solution that I see is to rename this form so that it won't cause conflicts with the previous one.

app/controllers/devise/registrations_controller.rb
```diff
@@ -97,6 +109,10 @@ def update_resource(resource, params)
     resource.update_with_password(params)
   end
 
+  def destroy_resource(resource, params)
+    resource.destroy_with_password(params)
```
tegon: Here the `params` parameter is actually only the literal `current_password`, right? If that's the case, I'd either rename it or change it here to receive the full params hash and then pass only the `current_password` to `resource#destroy_with_password`

app/controllers/devise/registrations_controller.rb
```diff
@@ -66,11 +66,23 @@ def update
...
+        Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+        set_flash_message! :notice, :destroyed
+        yield resource if block_given?
+        respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+      else
+        render :edit
+      end
+    else
+      resource.destroy
+      Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
```
tegon: WDYT about creating a method to extract this sign out logic? That way we can use it in both scenarios, without the need to duplicate it.

test/integration/registerable_test.rb
```diff
@@ -238,6 +238,38 @@ def user_sign_up
...
+      within(".destroy-password") do
+        fill_in 'current password', with: '87654321'
+      end
+
+      click_button "Cancel my account"
+      assert_contain "Current password is invalid"
+
+      assert_current_url '/users'
+    end
+    assert_equal "user@test.com", User.to_adapter.find_first.email
```
tegon: I think this assertion can be removed. The email wasn't changed in this test.

test/integration/registerable_test.rb
```diff
@@ -238,6 +238,38 @@ def user_sign_up
     assert User.to_adapter.find_all.empty?
   end
 
+  test 'a signed in user should be able to cancel their account with current password if password required' do
+    swap Devise, require_password_to_destroy: true  do
+      sign_in_as_user
+      get edit_user_registration_path
+
+      within(".destroy-password") do
```
tegon: I already mentioned the changes needed in the form, but just to add a comment here that ideally, for integration tests we won't include much implementation details - like a CSS class, for example - because that would made the test more fragile - e.g. if we change the CSS class the test would break, even though the system is still functioning the same way.

tegon:
@iorme1 are you still going to work on this?

feliperenan:
Closing in favor https://github.com/plataformatec/devise/pull/5029

PR #4851, opened 2018-04-24
Update simple form templates

gssbzn:
Update `simple_form` templates with changes from 16b3d6d67c7e017d461ea17ed29ea9738dc77e83 and 6260c29a867b9a656f1e1557abe347a523178fab

### DIFF
--- a/lib/generators/templates/simple_form_for/passwords/edit.html.erb
+++ b/lib/generators/templates/simple_form_for/passwords/edit.html.erb
@@ -7,7 +7,12 @@
   <%= f.full_error :reset_password_token %>
 
   <div class="form-inputs">
-    <%= f.input :password, label: "New password", required: true, autofocus: true, hint: ("#{@minimum_password_length} characters minimum" if @minimum_password_length) %>
+    <%= f.input :password,
+                label: "New password",
+                required: true,
+                autofocus: true,
+                hint: ("#{@minimum_password_length} characters minimum" if @minimum_password_length),
+                input_html: { autocomplete: "new-password" } %>
     <%= f.input :password_confirmation, label: "Confirm your new password", required: true %>
   </div>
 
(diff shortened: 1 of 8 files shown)


tegon:
Thank you @gssbzn!

lib/generators/templates/simple_form_for/passwords/edit.html.erb
```diff
@@ -7,7 +7,12 @@
...
 
   <div class="form-inputs">
-    <%= f.input :password, label: "New password", required: true, autofocus: true, hint: ("#{@minimum_password_length} characters minimum" if @minimum_password_length) %>
+    <%= f.input :password,
+                label: "New password",
+                required: true,
+                autofocus: true,
+                hint: ("#{@minimum_password_length} characters minimum" if @minimum_password_length),
+                input_html: { autocomplete: "new-password" } %>
     <%= f.input :password_confirmation, label: "Confirm your new password", required: true %>
```
amangano-privy: Several of the other password_confirmation fields in this PR were updated to include autofocus, should this one have been updated as well?

tegon: I went to look at the original code and found this comment: https://github.com/plataformatec/devise/commit/6260c29a867b9a656f1e1557abe347a523178fab#r31284249

So yeah, it looks like we forgot. Are you willing to take a stab on this one?

amangano-privy: Sorry, but I don't think I'll have the time to take this on. I just wanted to call it out in case it was an oversight so that it could be fixed later.

tegon: Sure thing, no problem. Thanks for pointing this out!

Issue #4763, opened 2018-01-25
Use same string on failed login regardless of whether account exists when in paranoid mode


JasonBarnabe:
Previously reported in #4568. devise is using `devise.failure.invalid` when you attempt to log in with an existing identifier but a wrong password, while it uses `devise.failure.not_found_in_database` when the identifier does not exist. In the default `en.yml` provided by devise, these strings are the same, but if you are using a different localization file, they may not be (see https://github.com/tigrish/devise-i18n/issues/223).

If you have enabled paranoid mode, the same localization key should be used for both cases.

tegon:
I haven't understood what the problem was on the previous issue, but now I do 😅 . Thanks, @JasonBarnabe.
This change seems reasonable to me, but I think we would have to do it in a major version since it isn't backward compatible.
Would you be willing to work on this?

tegon:
Closed via https://github.com/plataformatec/devise/pull/4854

PR #4854, opened 2018-04-28
Issue 4763: Use same string on failed login regardless of whether account exists when in paranoid mode

TonyMK9068:
Addresses issue #4763.

If paranoid mode is enabled, Devise will use `devise.failure.invalid` whether or not the user is attempting to log in with an existing identifier.

Previously, log in attempts in paranoid mode were using `devise.failure.not_found_in_database` if identifier did not exist, or `devise.failure.invalid` if identifier existed but password was wrong.


tegon:
Thank you @TonyMK9068!

Issue #4849, opened 2018-04-23
Support for GDPR


melsatar:
Do you have any plans to support GDPR? or how we can support that using the current setup

fakenine:
I've been wondering about this too. An option to opt out the fields that are personal data when installing devise would be nice. i.e: `rails generate devise:install --gdpr true` would

- remove the IP address field in the migration
- remove the email address field in the migration and replace it with a username field (thus changing the signin process)

rafaelfranca:
Devise is a generic authentication solution and has no plan or need to change to conform specific legislations of countries/states. That is under the responsibilities of the applications since it is their that have the context about which information is being stored.

fakenine:
GDPR is not a specific legislation of one or few countries/states. It concerns you as soon as you have users that are citizens of an European country (there are around 353 million internet users in Europe).

Even if you are in the US and have a little app, with one EU user, you are concerned by GDPR.

GDPR is complicated to understand, even more for junior developers/developers that build their first app. It would be a shame to have legal issues just because you left something like the tracking/registering of the user IP address in your app when installing devise.

Actually, I think we should build and have **privacy-by-default** tools, to build and have **privacy-by-default** apps.

Why not remove those sensitive fields (IP address, signin count ?) in the default devise generator (unlike what I said previously) ? If someone really needs them, they can add them via some option.

We can't just say "Oh, there's some new law that concerns everyone, it is very complicated and will change the way you develop applications. But we won't help you with that."

rafaelfranca:
> Why not remove those sensitive fields (IP address, signin count ?) in the default devise generator (unlike what I said previously) ?

I think that is a good compromise. If you want to do it please open a PR, but I don't think we should include an option and implement code specifically for gdpr.

melsatar:
It is important to include that in the future release and this should be soon. All service providers around the world started to align with GDPR. As @fakenine mentioned it is not like any specific legislation for a specific country.

I hope if you can reconsider this again.

rafaelfranca:
I just did in the comment above and I'm asking for help to implement. Just to be clear, can someone send a PR with the required changes? We will make it the new default.

fakenine:
@rafaelfranca Thank you! I will look into this this week.

mrsweaters:
@fakenine Would this also require encrypting the email address?

melsatar:
[...]

PR #4857, opened 2018-05-02
remove :trackable from the default modules

fakenine:
Hello 👋 

This refers to the discussion in #4849.

With the upcoming arrival of GDPR, I thought it would be preferable to remove the `:trackable` module from the default generator.

An IP address is considered as personal data, and storing it in the database along with tracking the user would require explicit consent from them.

I thus thought it would be a shame that a developer would get in legal trouble just because they left the `trackable` option when they first installed devise on their projects/applications (especially when using it for the first time, that's not the first thing we check or notice).

### DIFF
--- a/lib/generators/active_record/devise_generator.rb
+++ b/lib/generators/active_record/devise_generator.rb
@@ -54,11 +54,11 @@ module ActiveRecord
       t.datetime :remember_created_at
 
       ## Trackable
-      t.integer  :sign_in_count, default: 0, null: false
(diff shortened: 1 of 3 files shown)


rafaelfranca:
Should we also comment the columns on the migration generator?

fakenine:
Yes, my bad, I should have done that to stay consistent with the non-defaults `confirmable`, `lockable` for instance. I will do the required changes.

fakenine:
Done!

PR #4858, opened 2018-05-03
Fix in rdoc of DeviseController#_prefixes

lookageek:
Removing an additional "is" in the rdoc for DeviseController#_prefixes


nashby:
@lookageek thanks!

PR #4869, opened 2018-05-09
find auto-generated secret_key_base in development

### LINKED ISSUE #4864: Cannot find auto-generated secret_key_base in development
A friend and I started a new project on Ruby 2.5.1, Rails 5.2.0, Devise 4.4.3

When I cloned the project, I was unable to run any rails/rake tasks, since Devise was not able to find the key.

The intent is that Rails will automatically set `SECRET_KEY_BASE` for you in dev and test: [link to merge](https://github.com/rails/rails/pull/30067/files#diff-580e183141be706162224ef696bdb6c7R413)

I see that [activity](
https://github.com/plataformatec/devise/pull/4712) has been made in the past few months around hunting down the `SECRET_KEY_BASE` but it seems that there is still a small bit more to go:

Currently Devise's [Secret Key Finder](https://github.com/plataformatec/devise/blob/master/lib/devise/secret_key_finder.rb) is unable to find the autogenerated key. Rails only does "auto generation" of the key in `Rails.application.secret_key_base` which is not currently checked.

@jcoyne mentioned this in https://github.com/plataformatec/devise/issues/4710#issuecomment-347624819 but it seems that the current implementation doesn't take into account `Rails.application.secret_key_base` -- the only autogenerated version in dev/test for Rails 5.2

```
[7] pry(main)> Rails.application.credentials.secret_key_base
=> nil
[8] pry(main)> Rails.application.secrets.secret_key_base
=> nil
[9] pry(main)> Rails.application.config.secret_key_base
=> nil
[10] pry(main)> Rails.application.secret_key_base
=> "d56814a3585e59378bbcabc1467d2693"
```

- [Rails.application.secret_key_base - not checked in devise](https://github.com/rails/rails/blob/master/railties/lib/rails/application.rb#L428)

The [help message for this error mentions to hardcode a secret key string into the initializer](https://github.com/plataformatec/devise/blob/f220b992c338122226f6fd396056d5a1adf28df8/lib/generators/templates/devise.rb#L11), which would affect all environments. For anyone who comes across this issue until this is better handled, a more appropriate option is to use

```ruby
```

gencer:
With this fix, we will try latest changes in Rails 5.2 together with standard auto-generated secret_key_base in development as a fallback.

If no specified key found, auto-generated value will be used instead.

This fixes #4864

### DIFF
--- a/test/secret_key_finder_test.rb
+++ b/test/secret_key_finder_test.rb
@@ -32,6 +32,24 @@ class Rails52Config
   end
 end
 
+class Rails52SecretKeyBase
+  def credentials
+    OpenStruct.new(secret_key_base: nil)
+  end
+
+  def secrets
+    OpenStruct.new(secret_key_base: nil)
+  end
+
+  def config
+    OpenStruct.new(secret_key_base: nil)
+  end
+  
+  def secret_key_base
+    'secret_key_base'
+  end
+end
+
 class Rails41Secrets
   def secrets
     OpenStruct.new(secret_key_base: 'secrets')
@@ -75,6 +93,12 @@ class SecretKeyFinderTest < ActiveSupport::TestCase
     secret_key_finder = Devise::SecretKeyFinder.new(Rails52Config.new)
 
     assert_equal 'config', secret_key_finder.find
+  end
+
+  test "rails 5.2 uses secret_key_base when config is empty" do
+    secret_key_finder = Devise::SecretKeyFinder.new(Rails52SecretKeyBase.new)
+
+    assert_equal 'secret_key_base', secret_key_finder.find
   end
 
   test "rails 4.1 uses secrets" do
--- a/lib/devise/secret_key_finder.rb
+++ b/lib/devise/secret_key_finder.rb
@@ -13,6 +13,8 @@ module Devise
         @application.secrets.secret_key_base
       elsif @application.config.respond_to?(:secret_key_base) && key_exists?(@application.config)
         @application.config.secret_key_base
+      elsif @application.respond_to?(:secret_key_base) && key_exists?(@application)
+        @application.secret_key_base
       end
     end
 


gencer:
@tegon failed job is something about networking and/or i/o side on Travis. Maybe reschedule will fix the problem.

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
> The build has been terminated

tegon:
@gencer thanks! Can you add a new test case for this?

gencer:
@tegon I've added test case for `secret_key_base`, squashed them and pushed back.

ttilberg:
@tegon @gencer I referenced that branch from my Gemfile and it worked as expected.

The only thing I would suggest is the following consideration:

It seems to be that the Rails team views `Rails.application.secret_key_base` as the authoritative source for this information based on the fact that this method [already aggregates](https://github.com/rails/rails/blob/master/railties/lib/rails/application.rb#L430) the [other configuration locations](https://github.com/rails/rails/blob/master/railties/lib/rails/application.rb#L401), in addition to the new dev/test default. As such, I think this should be the first location checked, with the others being the fallbacks to preserve backwards compatibility. 

Even in the case where it does have to fall back, it's fewer checks and iterations happening as it's not going to be responding to that method in the first place.

Either way, this is successful in Rails 5.2.

Thanks for doing the PR on this. I understood what needed to be done, but am a little intimidated of the testing side of it.

test/secret_key_finder_test.rb
```diff
@@ -32,6 +32,25 @@ def config
...
+  def secrets
+    OpenStruct.new(secret_key_base: nil)
+  end
+
+  def config
+    OpenStruct.new(secret_key_base: nil)
+  end
+  
+  def secret_key_base
+    'secret_key_base' if Rails.env.development? || Rails.env.test?
```
tegon: Why do we need this `if` statement here?

gencer: Oh! How did I missed that! I was testing something else and I just left them there. I'll fix and squash now.

gencer:
@tegon the line you were referring was absolutely unnecessary as its already in dev/test environment. I removed that in addition, I had one more blank line which I removed that too.

Squashed&Pushed now.

issei-m:
Where is this fix released? I'm currently facing on this same problem on Devise 4.4.3.
Thanks.

tegon:
@issei-m it was released in `4.4.3`. Please move the discussion to #4879, which is an open issue where we're checking this. Also, it'd be good if you provide us your Rails version and the output of the following commands:
```
Rails.application.credentials.secret_key_base

Rails.application.secrets.secret_key_base

Rails.application.config.secret_key_base

**Comment by issei-m** [comment]
@tagon Oh really? It seems to me, `v4.4.3` [has no commits introduced in this PR](https://github.com/plataformatec/devise/blob/v4.4.3/lib/devise/secret_key_finder.rb#L10-L16).
And I cloned this repository and I hit the some commands to know who has these commits right now, then only `master` branch seems to have, while v4.4.3 doesn't:

```
$ git branch --contains 6c91648
* master

$ git tag --contains 6c91648

```

tegon:
@issei-m Ouch, you're right. I've mistaken this with https://github.com/plataformatec/devise/commit/962cea2039c72a92691af734ebbd8495dd5c0501, my bad. Thanks for pointing that out.
I'll see which commits we have unreleased and if there's something else to include in the next release and then bump a new version.
Just to make sure, if you use Devise from master it works, right?

tegon:
@issei-m thanks, I'll prepare a release as soon as I can.

tegon:
Released in `v4.5.0`

swrobel:
@gencer @tegon what about [ttilberg's comment](https://github.com/plataformatec/devise/pull/4869#issuecomment-387874576) which points out the incongruity between Devise behavior & Rails.

For example, on Heroku, they set `ENV['SECRET_KEY_BASE']` if it isn't already set for Rails apps. As a result, Rails ends up using that key instead of `Rails.application.credentials.secret_key_base`, although Devise will prefer the latter, resulting in different values for each.

issei-m:
@swrobel Simply, you can remove `secret_key_base` from your credentials.yml.inc, so `ENV['SECRET_KEY_BASE']` will be used always in Heroku environment.

swrobel:
@issei-m I still don't get why devise should behave differently from Rails

tegon:
@swrobel Devise's `secret_key_base` doesn't necessarily need to be the same as Rails's. We only try to get it from Rails for convenience, so that you don't need to manually set it to get started.
When I looked at @ttilberg comment, I thought it was only a matter of order - e.g. where we look first for a key - I didn't realize that Rails looked for the environment variable too.
We can support this in a new version, but this PR was mostly about fixing a bug we introduced in a previous version. Would you mind opening an issue asking to support this?

ttilberg:
@tegon My comment wasn't as much about "where to look first" for the sake of looking in many places, but instead a solution to leverage the configuration of modern rails apps while maintaining backwards compatibility. There's a few links to the relevant parts of the rails source in both my comment above and the original issue this closes that I thought were helpful to understand the "current rails configuration strategy" -- it felt a little all over the place but seems to get unified at `Rails.application.secret_key_base` accounting for all expected possibilities, including the new dynamic generation for dev/test.

What is the best way to ask for this support in a new issue? Something along the lines of "Feature: Prefer Rails default SECRET_KEY_BASE configuration?" It feels a bit superfluous now, but I think it's a good future-proofing thing to do.

gencer:
What I get so far that we need to also check `ENV` for `SECRET_KEY_BASE` (if possible at first) and regardless what we've found so far, we will use this Environment variable as default secret key.

Simply, `ENV["SECRET_KEY_BASE"]` has higher priority and overrides all (or If found, no need to check further places)

We need an another **PR** for this. I'll have a look in spare time (No promises) and create PR with tests (if no one contribute in this time of course). But, We still need an issue to track those changes.

ttilberg:
Hey @gencer -- I was already in the middle of taking a stab at refactoring the key finder just now, I'll add that consideration (to check ENV first) as well and post a PR in a moment here. In addition, it was about checking the correct "default location in Rails" first, instead of last. Thanks for your help working through this!

ttilberg:
@tegon @gencer Can you guys take a look at PR I just posted above? I don't have a lot of experience/confidence with contributing to such a serious project, so I'd appreciate feedback on:

- Was this an appropriate way to contribute (going straight to PR without an issue, etc)
- The refactor decisions

I added a test for the `ENV` addition @gencer mentioned, and all tests are still passing.

md5:
> @swrobel Simply, you can remove `secret_key_base` from your credentials.yml.inc, so `ENV['SECRET_KEY_BASE']` will be used always in Heroku environment.

This is not the case. If `Rails.application.credentials` exists at all, Devise will try to get `secret_key_base` from `Rails.application.credentials.secret_key_base`. If it does not exist, it doesn't try the other sources. So if you want to set `SECRET_KEY_BASE` from an environment variable and still use `config/credentials.yml.enc` for other values, that won't work.

I think @ttilberg is correct that `Rails.application.secret_key_base` should be called first if it exists to avoid this problem.

md5:
Or rather, removing `secret_key_base` from `credentials.yml.enc` causes `ENV['SECRET_KEY_BASE']` to be used by Rails, but not by Devise.

PR #4871, opened 2018-05-10
Set FailureApp #response to ActionDispatch::Response

delner:
In the `FailureApp`, it wraps a call to the recall app (aka controller), and sets the recall response as . its own, which is `[200, Hash, ActionDispatch::Response::RackBody]`. Normally this `response` value in an ActionController is an `ActionDispatch::Response` object, not an `Array`.

This is problematic for anyone who derives from `FailureApp` or otherwise relies on `response` to be an `ActionDispatch::Response` object, as would reasonably be expected of an `ActionController::Metal` instance. More specifically, this manifested [here](https://github.com/DataDog/dd-trace-rb/pull/420) when some instrumentation expected to receive an `ActionDispatch::Response` object instead of an `Array`.

My suggestion would be to set the `FailureApp`'s response as the `ActionDispatch::Response` from the recall app. I'm not sure my current implementation is the best way of achieving that, but I figured it might be a good place to start a discussion and illustrate what I'd like to accomplish. This should work for all versions of Rails and Ruby supported by Devise.

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -71,8 +71,7 @@ module Devise
       end
 
       flash.now[:alert] = i18n_message(:invalid) if is_flashing_format?
-      # self.response = recall_app(warden_options[:recall]).call(env)
-      self.response = recall_app(warden_options[:recall]).call(request.env)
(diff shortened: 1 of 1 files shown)


tegon:

Thank you!

Issue #4782, opened 2018-02-24
Unauthenticated DELETE /users/sign_out returns 204 status


christophweegen:
While working with a gem called [`devise-jwt`](https://github.com/waiting-for-dev/devise-jwt), i noticed that an unauthenticated request to `DELETE /users/sign_out` returns a `204 No-Content`. 

The same as an authenticated request, whereas an unauthenticated one should return something like `401 Unauthorized`.

You can see the issue in detail here https://github.com/waiting-for-dev/devise-jwt/issues/71

Tracking down the problem, it turned out that this problem is actually coming from `devise`, a plain `devise` app returns a `204` on an unauthorized request too.

You can try this out with sending a simple `DELETE /users/sign_out` to a plain `devise` app without providing a session cookie.

Additionally while testing that, i recognized that if you have a `http_basic_authenticate_with` in your `ApplicationController` it returns a `204` too, while it actually should be intercepted before by the `basic_auth`, or not? This wasn't an error on my side because maybe the `basic_auth` was still authenticated by an earlier request. Tried it on a `GET` request before too and got a `401` and a `basic_auth` message. After that rejected request i directly tried the `DELETE` again and it returned a `204` again.

I'm not sure what caused this behaviour of the `http_basic_authenticate_with`, maybe it's actually a bug in `rails`? Not sure if i should open an issue there right now too.

@rafaelfranca since you're in the rails team and contribute a lot to devise, do you maybe have an idea?

Thanks!

tegon:
I see no problem in returning a 401 if the user isn't signed in, we actually already [check if there's any user in the session](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/sessions_controller.rb#L61) before calling sign out, so if you're willing to work on this, a PR would be welcomed.
I'm also curious, is this causing a problem in your application? For me, it seems like we only hit this request if we know upfront that a user is signed in.

christophweegen:
Hey thanks for your reply :-)

Actually it isn't really a problem. I just discovered it while using [`devise-jwt`](https://github.com/waiting-for-dev/devise-jwt) for a rails api-only application while i was making some requests while testing. The idea was to show flash-like messages in the frontend like "You signed out successfully" or "You are already signed out" smth. like that, depending on the http status. This obviously doesn't work out since both possibilities return the same status code. I just thought making a distinction would adhere more to http semantics too, like described above.

I will work smth. out when i got time. :-)

aamarill:
Hello,

[...]

PR #4878, opened 2018-05-18
Modified sessions controller to return 401 for destroy action with no user signed in

aamarill:
This PR is for issue #4782  @tegon
To clarify the 3 tests added:

### 1. Authenticated non-navigational request - returns 204.
- This was the original behavior before I made any changes. I simply added this test to ensure this was not broken after my implementation.  

### 2. Unauthenticated non-navigational request - returns 401.
- This is the new behavior added.

### 3. Unauthenticated navigational request -  returns 302.
- This was the original behavior before I made any changes. I simply added this test to ensure this was not broken after my implementation.

### DIFF
--- a/app/controllers/devise/sessions_controller.rb
+++ b/app/controllers/devise/sessions_controller.rb
@@ -28,7 +28,7 @@ class Devise::SessionsController < DeviseController
     signed_out = (Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name))
     set_flash_message! :notice, :signed_out if signed_out
     yield if block_given?
-    respond_to_on_destroy
+    respond_to_on_destroy(status: :no_content)
   end
 
   protected
@@ -62,10 +62,7 @@ class Devise::SessionsController < DeviseController
     if all_signed_out?
       set_flash_message! :notice, :already_signed_out
 
-      respond_to do |format|
-        format.all {head :unauthorized}
-        format.any(*navigational_formats) { redirect_to after_sign_out_path_for(resource_name) }
-      end
+      respond_to_on_destroy(status: :unauthorized)
     end
   end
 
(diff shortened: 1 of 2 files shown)


aamarill:
Looking into what went wrong with Travis CI.
I did forget to test using Mongoid and different Gemfiles, my apologies. Hoping that's all I need to account for.

test/controllers/sessions_controller_test.rb
```diff
@@ -88,6 +88,30 @@ class SessionsControllerTest < Devise::ControllerTestCase
     assert_equal 204, @response.status
   end
 
+  test "#destroy returns 204 status if the requested format is not navigational" do
```
feliperenan: This behavior is already tested in the test above. IMHO changing the test name would be enough.

`#destroy doesn't set the flash and returns 204 status if the requested format is not navigational`

test/controllers/sessions_controller_test.rb
```diff
@@ -88,6 +88,30 @@ class SessionsControllerTest < Devise::ControllerTestCase
...
+    delete :destroy, format:'json'
+    assert_equal 204, @response.status
+  end
+
+  test "#destroy returns 401 status if user is not signed in and the requested format is not navigational" do
+    delete :destroy, format: 'json'
+    assert_equal 401, @response.status
+  end
+
+  test "#destroy returns 302 status if user is not signed in and the requested format is navigational" do
```
feliperenan: 👏

app/controllers/devise/sessions_controller.rb
```diff
@@ -62,7 +62,10 @@ def verify_signed_out_user
     if all_signed_out?
       set_flash_message! :notice, :already_signed_out
 
-      respond_to_on_destroy
+      respond_to do |format|
```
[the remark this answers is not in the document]
feliperenan: Maybe the default `:no_content` isn't a good idea. I like the idea to have the status explicit in the caller.

```ruby
respond_to_on_destroy(status: :no_content)
respond_to_on_destroy(status: :unauthorized)
```

tegon:
@aamarill are you still going to work on this?

aamarill:
@tegon please go ahead and assign to someone else. Thank you

tegon:
@aamarill Since the changes were small, I committed them directly to your branch. Thanks for your contribution!

PR #4891, opened 2018-06-19
Add dependency version and SemVer badges to README

greysteil:
Would you be up for adding a badge that shows how SemVer compliant / bug free new releases are? I was looking through the data we gather at Dependabot and realised we could put one together, so threw together the below. I've also added a gem version badge, as I always look for that when figuring out what version to use in my Gemfile.

[Gem Version](https://rubygems.org/gems/devise) [SemVer compatibility](https://dependabot.com/compatibility-score.html?dependency-name=devise&package-manager=bundler&version-scheme=semver)

If you click through then there's a description of how it's calculated - basically it takes all of the relevant updates Dependabot has done for projects that use devise and checks what percentage of the time specs pass on the upgrade PR.

The score isn't 100% because the 4.4.2 release broke a few people's tests, but it should steadily climb back up, assuming you're following SemVer.

In future, I'm also planning to build a tool that lets you view the Dependabot data for pre-release updates, so you'd be able to see whether they break anyone's tests before releasing a full release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ By [Plataformatec](http://plataformatec.com.br/).
 
 [![Build Status](https://api.travis-ci.org/plataformatec/devise.svg?branch=master)](http://travis-ci.org/plataformatec/devise)
 [![Code Climate](https://codeclimate.com/github/plataformatec/devise.svg)](https://codeclimate.com/github/plataformatec/devise)
+[![Gem Version](https://badge.fury.io/rb/devise.svg)](https://rubygems.org/gems/devise)
+[![SemVer compatibility](https://api.dependabot.com/badges/compatibility_score?dependency-name=devise&package-manager=bundler&version-scheme=semver)](https://dependabot.com/compatibility-score.html?dependency-name=devise&package-manager=bundler&version-scheme=semver)
 
 This README is [also available in a friendly navigable format](http://devise.plataformatec.com.br/).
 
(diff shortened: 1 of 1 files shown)


tegon:
Hello @greysteil, thanks for your pull request.

I'm ok with the gem version badge, but I'm not sure about the SemVer one. Your app seems interesting, but there are tough questions to answer when SemVer is the subject. For example, a bugfix that changes an API, or a private API change that people are using - which would break some application's tests.

tegon:
I'm closing this because it had no recent activity.

Thank you!

greysteil:
Oh man, sorry, I don't know how I missed this!

I think those are valid points - Dependabot does a lot of Devise updates, so has quite a bit of data to work with, but there are absolutely cases where someone is tapping into private APIs and will be broken by a patch release. It's very hard for Dependabot to understand those cases, although we do now link to failing pull requests (if they're on public repos), so users can figure out what went wrong for themselves (see, for example, the [4.4.1 -> 4.4.2 upgrade](https://dependabot.com/compatibility-score/?dependency-name=devise&new-version=4.4.2&package-manager=bundler&previous-version=4.4.1)). I've been encouraging as many open source apps as possible to use it so that we get good data there.

On the badge, I've also put together this version:

[SemVer compatibility](https://dependabot.com/compatibility-score.html?dependency-name=devise&package-manager=bundler&version-scheme=semver&new-version=latest)

The idea with the above is that any breaks from previous versions are excluded (so in this case, it isn't affected by the CI failures that 4.4.2 saw, since those "breaks" have since been fixed). The folks at [webpack](https://github.com/webpack/webpack) are the ones who recommended that version of the badge, and I think it's an improvement - what do you reckon?

PR #4892, opened 2018-06-20
update devise to use latest minor bcrypt version

muhammadn:
bcrypt fails when it is being used with libxcrypt instead of glib libcrypt.

reference:
https://gitlab.com/gitlab-org/gitlab-ce/issues/48149

The latest bcrypt (3.1.12) supports libxcrypt which is being used in fedora linux distributions. fedora does not use libcrypt and this update will fix that.

### DIFF
--- a/devise.gemspec
+++ b/devise.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency("warden", "~> 1.2.3")
   s.add_dependency("orm_adapter", "~> 0.1")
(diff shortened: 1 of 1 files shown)


Qwertie-:
Is it possible to get this merged soon? devise has been broken on fedora for about 2 months now.

rafaelfranca:
Thank you for the pull request but the current code allow you to install 3.1. Why do we need to change this?

Qwertie-:
@rafaelfranca I just checked and indeed it is possible to install the latest bcrypt version so it seems this PR isn't needed.

PR #4897, opened 2018-07-03
Corrects "brakes" -> "breaks" when talking about tests

philnash:


tegon:
Thank you, @philnash!

Issue #4542, opened 2017-05-18
undefined method `will_save_change_to_email?' for User


mtoribio:
Hi,
I have problems with:
- Devise 4.3
- Rails 5.1
- Mongoid 6.1
```
NoMethodError (undefined method `will_save_change_to_email?' for #<User:0x007f80b1778ec0>):
  
activesupport (5.1.1) lib/active_support/callbacks.rb:413:in `block in make_lambda'
activesupport (5.1.1) lib/active_support/callbacks.rb:177:in `block (2 levels) in halting_and_conditional'
activesupport (5.1.1) lib/active_support/callbacks.rb:177:in `each'
activesupport (5.1.1) lib/active_support/callbacks.rb:177:in `all?'
activesupport (5.1.1) lib/active_support/callbacks.rb:177:in `block in halting_and_conditional'
activesupport (5.1.1) lib/active_support/callbacks.rb:507:in `block in invoke_before'
activesupport (5.1.1) lib/active_support/callbacks.rb:507:in `each'
activesupport (5.1.1) lib/active_support/callbacks.rb:507:in `invoke_before'
activesupport (5.1.1) lib/active_support/callbacks.rb:130:in `run_callbacks'
mongoid (6.1.0) lib/mongoid/interceptable.rb:138:in `run_callbacks'
activesupport (5.1.1) lib/active_support/callbacks.rb:825:in `_run_validate_callbacks'
activemodel (5.1.1) lib/active_model/validations.rb:405:in `run_validations!'
activemodel (5.1.1) lib/active_model/validations/callbacks.rb:110:in `block in run_validations!'
activesupport (5.1.1) lib/active_support/callbacks.rb:131:in `run_callbacks'
mongoid (6.1.0) lib/mongoid/interceptable.rb:138:in `run_callbacks'
activesupport (5.1.1) lib/active_support/callbacks.rb:825:in `_run_validation_callbacks'
activemodel (5.1.1) lib/active_model/validations/callbacks.rb:110:in `run_validations!'
activemodel (5.1.1) lib/active_model/validations.rb:335:in `valid?'
mongoid (6.1.0) lib/mongoid/validatable.rb:97:in `valid?'
activemodel (5.1.1) lib/active_model/validations.rb:372:in `invalid?'
mongoid (6.1.0) lib/mongoid/persistable/creatable.rb:115:in `prepare_insert'
mongoid (6.1.0) lib/mongoid/persistable/creatable.rb:23:in `insert'
[...]

PR #4910, opened 2018-07-28
Confirmable and rails apps using both ActiveRecord and Mongoid

sandboxws:
## Problem Statement

The `confirmable` module doesn't work as expected when the parent Rails app is using both ActiveRecord and Mongoid.

## Solution

Assign a default `orm` attribute to the Devise config with a default value of `activerecord`. This value then is overridden/changed to `mongoid` if `lib/devise/orm/mongoid` is required by the parent Rails application.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -61,6 +61,10 @@ module Devise
   # True values used to check params
   TRUE_VALUES = [true, 1, '1', 't', 'T', 'true', 'TRUE']
 
+  # ORM used by Devise. Allowed values [:activerecord, :mongoid]
+  mattr_accessor :orm
+  @@orm = :active_record
+
   # Secret key used by the key generator
   mattr_accessor :secret_key
   @@secret_key = nil
@@ -298,7 +302,15 @@ module Devise
   end
 
   def self.activerecord51? # :nodoc:
-    defined?(ActiveRecord) && ActiveRecord.gem_version >= Gem::Version.new("5.1.x")
+    self.activerecord? && (defined?(ActiveRecord) && ActiveRecord.gem_version >= Gem::Version.new("5.1.x"))
+  end
+
+  def self.activerecord? # :nodoc:
+    Devise.orm.to_sym == :active_record
+  end
+
+  def self.mongoid? # :nodoc:
+    Devise.orm.to_sym == :mongoid
   end
 
   # Default way to set up Devise. Run rails generate devise_install to create
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -48,7 +48,7 @@ module Devise
       included do
         before_create :generate_confirmation_token, if: :confirmation_required?
         after_create :skip_reconfirmation_in_callback!, if: :send_confirmation_notification?
-        if defined?(ActiveRecord) && self < ActiveRecord::Base # ActiveRecord
+        if Devise.activerecord? && defined?(ActiveRecord) && self < ActiveRecord::Base # ActiveRecord
           after_commit :send_on_create_confirmation_instructions, on: :create, if: :send_confirmation_notification?
           after_commit :send_reconfirmation_instructions, on: :update, if: :reconfirmation_required?
         else # Mongoid
--- a/lib/devise/orm/active_record.rb
+++ b/lib/devise/orm/active_record.rb
@@ -2,6 +2,7 @@
 
 require 'orm_adapter/adapters/active_record'
 
+Devise.orm = :active_record
 ActiveSupport.on_load(:active_record) do
   extend Devise::Models
 end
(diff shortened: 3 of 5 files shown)


tegon:
Hello @sandboxws, thanks for your pull request.

> The confirmable module doesn't work as expected when the parent Rails app is both ActiveRecord and Mongoid.

Can you explain with more details what the issue is?

sandboxws:
@tegon Consider the following use case. A Ruby on Rails application uses both Mongoid and ActiveRecord. Devise is used with Mongoid, i.e. `require 'devise/orm/mongoid'` is included in the devise intializer and the `User` model includes the confirmable module.

And this is when the following line breaks the `unconfirmed_email` functionality of this module since the conditon assumes that if `ActiveRecord` is loaded then that means the paretn Rails app is using `ActiveRecord` as the devise ORM which is not true in this use case.

https://github.com/plataformatec/devise/blob/master/lib/devise/models/confirmable.rb#L51

Please let me know if you need any further details on the issue.

tegon:
@sandboxws the next condition was supposed to fix this `self < ActiveRecord::Base`. So, this is not working, right?

sandboxws:
@tegon correct.

tegon:
@sandboxws Ok, got it. We're not covering this in our test suite, we switch which ORM is going to be used, and we never use them together, so we can't be sure it's working.
I'm not sure this is the best solution. First, I'm concerned about using a global accessor for this, if it can cause problems in multi-thread environments.
Second, the accessor is changed every time a ORM file is loaded. Won't the last ORM loaded always be used throught the application?

I think we would need to correctly setup a test for this case - both ORMs used thogheter - so we can perform more expensive tests and find other alternatives. If you can help creating a sample application that uses both AR and Mongoid that would be great, since you have more context with this problem. I know some hacks are necessary to make Rails work with two databases, but I don't remember exactly how to do it.

sandboxws:
@tegon 100% agree, which ORM is being used is something that most definitely needs to be covered by tests. Please find my comments below on your questions.

> First, I'm concerned about using a global accessor for this, if it can cause problems in multi-thread environments.

Valid concern, this value needs to be set in a thread-safe manner so to speak.


> Second, the accessor is changed every time a ORM file is loaded. Won't the last ORM loaded always be used throught the application?

How is devise currently handling such case? If I load both `lib/devise/orm/mongoid.rb` and `lib/devise/orm/active_record.rb` under the devise config initializer?


> If you can help creating a sample application that uses both AR and Mongoid that would be great

Sure thing, I'll have a sample application demonstrating the issue as soon as possible.

tegon:
@sandboxws thanks! The sample application will certainly help a lot.
About what you said here:
> which ORM is being used is something that most definitely needs to be covered by tests

Just to make sure it's clear, we do cover *which* ORM is being used. It works if you use Active Record *or* Mongoid. The problem is if you use **both** at the same time.

sandboxws:
@tegon here is a sample app demonstrating the problem. https://github.com/sandboxws/devise_mongoid_ar

And here is an example of the issue…

User model devise's settings

```ruby
devise :database_authenticatable,
         :registerable,
         :recoverable,
         :validatable,
         :confirmable,
         :lockable,
         :trackable
```

```shell
2.3.4 :005 > u = User.first
 => #<User _id: BSON::ObjectId('5b74a74dd5cc2e563e3072c3'), created_at: Wed, 15 Aug 2018 22:21:01 UTC +00:00, email: "foobar@gmail.com", first_name: "Foo", last_name: "Bar", updated_at: Wed, 15 Aug 2018 22:26:31 UTC +00:00>
2.3.4 :006 > u.email
 => "foobar@gmail.com"
2.3.4 :007 > u.unconfirmed_email
 => nil
2.3.4 :008 > u.confirmed?
 => true
2.3.4 :009 > u.email = 'test@gmail.com'
 => "test@gmail.com"
2.3.4 :010 > u.save
 => true
2.3.4 :011 > u = User.first
 => #<User _id: BSON::ObjectId('5b74a74dd5cc2e563e3072c3'), created_at: Wed, 15 Aug 2018 22:21:01 UTC +00:00, email: "test@gmail.com", first_name: "Foo", last_name: "Bar", updated_at: Wed, 15 Aug 2018 22:27:08 UTC +00:00>
2.3.4 :012 > u.email
 => "test@gmail.com"
2.3.4 :013 > u.confirmed?
 => true
2.3.4 :014 > u.unconfirmed_email
 => nil
2.3.4 :015 >
```

As you can see, the user model was saved successfully; however, the `unconfirmed_email` attribute was not populated with the new email added.

tegon:
I'm closing this pull request - since it doesn't fully solve the issue - to keep the discussion in the related issue: https://github.com/plataformatec/devise/issues/4542.
When we finish to gather the information needed to solve this and find a solution, we'll open a new pull request.

PR #4911, opened 2018-07-30
update rubydoc.info links in README to the url `www.rubydoc.info`

guillaumecabanel:
since the ssl certificate is only valid for www.rubydoc.info (not rubydoc.info)


tegon:
Thanks, @guillaumecabanel!

PR #4916, opened 2018-08-03
[#4841] Add `after_database_authentication` callback after sign_in immediately after password update

### LINKED ISSUE #4841: Method #after_database_authentication not called after reset_password
## Precheck

Callback #after_database_authentication is not called after reset_password but user is authenticated and in DB `sign_in_count` is 1.

## Environment

- Ruby 2.4.2
- Rails 5.1.4
- Devise 4.3.0

## Current behavior

1. Invite user with reset password link;
2. Follow that link;
3. User is logged in and password updated.
4. `after_database_authentication` callback is not triggered.

## Expected behavior

1. Invite user with reset password link;
2. Follow that link;
3. User is logged in and password updated.
4. `after_database_authentication` callback should be triggered.

kanmaniselvan:
Fixes #4841

### DIFF
--- a/test/controllers/passwords_controller_test.rb
+++ b/test/controllers/passwords_controller_test.rb
@@ -31,4 +31,9 @@ class PasswordsControllerTest < Devise::ControllerTestCase
     put_update_with_params
     assert_redirected_to custom_path
   end
+
+  test 'calls after_database_authentication callback after sign_in immediately after password update' do
+    User.any_instance.expects :after_database_authentication
+    put_update_with_params
+  end
 end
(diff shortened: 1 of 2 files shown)


kanmaniselvan:
@tegon For your review, please. Thank you.

test/controllers/passwords_controller_test.rb
```diff
@@ -31,4 +31,9 @@ def put_update_with_params
     put_update_with_params
     assert_redirected_to custom_path
   end
+
+  test 'calls after_database_authentication callback after sign_in immediately after password update' do
+    User.any_instance.expects :after_database_authentication
```
tegon: Can we use `@user` here? I'd like to avoid `any_instance` when it's possible.

PR #4925, opened 2018-08-16
Change priority of SECRET_KEY_BASE configuration locations

ttilberg:
This PR changes the priority of search locations:

- If `ENV['SECRET_KEY_BASE'` is set, use that: https://github.com/plataformatec/devise/pull/4869#issuecomment-413586348
- Check common Rails configuration locations:
  - Default, unified location in Rails 5 (includes autogenerated value for dev/test in newer Rails): https://github.com/plataformatec/devise/issues/4864, https://github.com/plataformatec/devise/pull/4869#issuecomment-387874576
  - Other common locations for backwards compatibility

I found the `SecretKeyFinder` to be hard to understand since there were a lot of safe-access checks, and a lot of duplication of effort. This is the result of my attempt to refactor this, making more quick to understand how and where we are checking.

The previous version was essentially a list of "if we get a value from safely navigating here, then return it". Now there is a simple list of method chains to attempt. It should be compatible with older rubies, as it avoids using `&.`.

`Enumerable#find` is convenient for this, since it will return `nil` if nothing matched, and we can `break` early to return a result (instead of the enumerated item). I learned this trick at Rubyconf last year, and thought it was clever: [link](https://youtu.be/nOa6FhMKsZ0?t=1282).

### DIFF
--- a/lib/devise/secret_key_finder.rb
+++ b/lib/devise/secret_key_finder.rb
@@ -2,26 +2,36 @@
 
 module Devise
   class SecretKeyFinder
+    COMMON_CONFIGURATION_LOCATIONS = [
+      'application.secret_key_base',
+      'application.credentials.secret_key_base',
+      'application.secrets.secret_key_base',
+      'application.config.secret_key_base'
+    ].freeze
+
     def initialize(application)
       @application = application
     end
 
     def find
-      if @application.respond_to?(:credentials) && key_exists?(@application.credentials)
-        @application.credentials.secret_key_base
-      elsif @application.respond_to?(:secrets) && key_exists?(@application.secrets)
-        @application.secrets.secret_key_base
-      elsif @application.config.respond_to?(:secret_key_base) && key_exists?(@application.config)
-        @application.config.secret_key_base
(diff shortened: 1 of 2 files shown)


lib/devise/secret_key_finder.rb
```diff
@@ -2,26 +2,36 @@
...
     private
 
-    def key_exists?(object)
-      object.secret_key_base.present?
+    attr_reader :application
+
+    def reduce_methods(chain)
+      begin
+        chain.split('.').reduce(self){|obj, msg| obj.send(msg)}
+      rescue NoMethodError
```
rubyrider: Why you are so explicitly handling NoMethodError here? Why not just releasing it to capture any class of error?

ttilberg: That is the only error I anticipate, and it is part of the code logic (return `nil`, wasn't successful) 

I think anything else should raise. This is my attempt at replacing the previous pattern in this file of `obj.call(:method) if obj.responds_to?(:method)` without using the newer safe navigator: `&.`. I didn't want to overstep my exception capturing.

PR #4926, opened 2018-08-17
rememberable and timeoutable should work together

### LINKED ISSUE #4778: timeout and rememberable
Why was the timeout and rememberable compatibility removed?

```ruby
def timedout?(last_access)
    return false if remember_exists_and_not_expired?
```

https://github.com/plataformatec/devise/blob/master/lib/devise/models/timeoutable.rb#L30
(removed in commit https://github.com/plataformatec/devise/commit/4ec7dc0f2777bde10125dcf6f72c536232086665)

## Current behavior

Timeout and Rememberable do not work together. The shorter one overrules the longer one.

## Expected behavior

I would expect rememberable to control sessions via cookie, while timeoutable can still control session time without cookies. (This is both relevant for http, where cookies are disabled on `secure: true`, as well as for users who explicitly choose not to select remember me).

At the moment, when I want to use rememberable, I have not control what so ever on session time, just on the session cookie expiration time.

prithvi16:
Fixes #4778

Now if rememberable is enabled and has not expired then `timedout?` returns false. 
Added a test for this behavior.

### DIFF
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -101,7 +101,7 @@ module Devise
       end
 
       def remember_expired?
-        remember_created_at.nil? || (remember_expires_at <= Time.now.utc)
+        remember_created_at.nil? || (remember_expires_at <= Time.zone.now.utc)
       end
 
       def remember_me?(token, generated_at)
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -100,6 +100,10 @@ module Devise
       def after_remembered
       end
 
+      def remember_expired?
+        remember_created_at.nil? || (remember_expires_at <= Time.now.utc)
+      end
+
       def remember_me?(token, generated_at)
         # TODO: Normalize the JSON type coercion along with the Timeoutable hook
         # in a single place https://github.com/plataformatec/devise/blob/ffe9d6d406e79108cf32a2c6a1d0b3828849c40b/lib/devise/hooks/timeoutable.rb#L14-L18
--- a/lib/devise/models/timeoutable.rb
+++ b/lib/devise/models/timeoutable.rb
@@ -27,10 +27,16 @@ module Devise
       end
 
       # Checks whether the user session has expired based on configured time.
+      # Returns false if Rememberable has been enabled and has not expired
       def timedout?(last_access)
+        return false if remember_exists_and_not_expired?
         !timeout_in.nil? && last_access && last_access <= timeout_in.ago
       end
 
+      def remember_exists_and_not_expired?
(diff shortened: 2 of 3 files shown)


lib/devise/models/rememberable.rb
```diff
@@ -100,6 +100,10 @@ def rememberable_options
       def after_remembered
       end
 
+      def remember_expired?
+        remember_created_at.nil? || (remember_expires_at <= Time.now.utc)
```
duderamos: Shouldn't we consider app timezone?

```ruby
Time.zone.now
```

wmene:
Any idea when this will get merged?

lib/devise/models/rememberable.rb
prithvi16: @duderamos yes, makes sense. Making this change.

lib/devise/models/timeoutable.rb
```diff
@@ -27,10 +27,16 @@ def self.required_fields(klass)
...
+      # Returns false if Rememberable has been enabled and has not expired
       def timedout?(last_access)
+        return false if remember_exists_and_not_expired?
         !timeout_in.nil? && last_access && last_access <= timeout_in.ago
       end
 
+      def remember_exists_and_not_expired?
+        return false unless respond_to?(:remember_created_at) && respond_to?(:remember_expired?)
+        remember_created_at && !remember_expired?
+      end
```
feliperenan: Let's leave a line blank between these two methods 🎨

test/models/timeoutable_test.rb
```diff
@@ -50,4 +50,10 @@ class TimeoutableTest < ActiveSupport::TestCase
     user = create_admin(remember_created_at: Time.current)
     assert user.timedout?(31.minutes.ago)
   end
+
+  test 'should return false if rememberable is enabled and has not expired' do
+    user = create_user
+    user.remember_me!
+    refute user.timedout?(31.minutes.ago)
+  end
 end
```
feliperenan: We could add one more test to cover the scenario that the rememberable is enabled and is expired.

lib/devise/models/rememberable.rb
```diff
@@ -100,6 +100,10 @@ def rememberable_options
       def after_remembered
       end
 
+      def remember_expired?
+        remember_created_at.nil? || (remember_expires_at <= Time.zone.now.utc)
```
duderamos: This is not exactly what I meant here: https://github.com/plataformatec/devise/pull/4926#pullrequestreview-155670776. Your change will end up returning the same value as before.

Consider the following:
```suggestion
        remember_created_at.nil? || (remember_expires_at <= Time.zone.now)
```

So the application will check reminders accordingly to local timezone.

PR #4933, opened 2018-08-28
Update migration.rb

gdurelle:
Use a Hash index algorithm on token comparison fields for better performance on large datasets

### DIFF
--- a/lib/generators/active_record/templates/migration.rb
+++ b/lib/generators/active_record/templates/migration.rb
@@ -13,8 +13,8 @@ class DeviseCreate<%= table_name.camelize %> < ActiveRecord::Migration<%= migrat
     end
 
     add_index :<%= table_name %>, :email,                unique: true
-    add_index :<%= table_name %>, :reset_password_token, unique: true
-    # add_index :<%= table_name %>, :confirmation_token,   unique: true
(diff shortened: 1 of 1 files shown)


tegon:
Hello @gdurelle, thanks for your pull request.

I'm not sure about this change though, it seems database and application specific. What I mean is that it might be better for your use case but not necessarily a good default for most cases.
I know that in Postgres, for example, hash index usage is [discouraged](https://www.postgresql.org/docs/9.5/static/indexes-types.html):

> Hash index operations are not presently WAL-logged, so hash indexes might need to be rebuilt with REINDEX after a database crash if there were unwritten changes. Also, changes to hash indexes are not replicated over streaming or file-based replication after the initial base backup, so they give wrong answers to queries that subsequently use them. For these reasons, hash index use is presently discouraged.

That's why I'm closing this PR. You can use the current generator and change it later to better suit your application needs.

Thank you!

gdurelle:
Hi, 

Yes I understand, it was an PG issue for a long time, BUT the last version is no longer victim of this problem.
https://www.postgresql.org/docs/10/static/indexes-types.html

That's why I considered submiting this PR.
But all things considered, for the sake of compatibility with old versions of PG, this is effectively a bad idea.

Thanks anyway for taking the time.

++

PR #4935, opened 2018-09-07
Fix can use options argument at DatabaseAuthenticatable#update_with_password

ihatov08:


tegon:
Hello @ihatov08, thanks for the pull request.

Would you mind to explain what issue this is solving?

ihatov08:
@tegon 
This method can specify the options variable as the second argument, but it fixed that an error occurs even if it is specified.

before
```ruby
[2] pry(main)> User.first.update_with_password({current_password: 'password'}, profile_name: 'hoge')
ArgumentError: wrong number of arguments (given 2, expected 1)
from /usr/local/bundle/gems/activerecord-5.2.1/lib/active_record/persistence.rb:423:in `update'
```

after
```ruby
[2] pry(main)> User.first.update_with_password({current_password: 'password'}, profile_name: 'hoge') 
=> true
```

lib/devise/models/database_authenticatable.rb
```diff
@@ -64,7 +64,7 @@ def clean_up_passwords
       # users to change relevant information like the e-mail without changing
       # their password). In case the password field is rejected, the confirmation
       # is also rejected as long as it is also blank.
-      def update_with_password(params, *options)
+      def update_with_password(params, **options)
```
feliperenan: Can we add a test for this change? I could add it closest to this one https://github.com/plataformatec/devise/blob/master/test/models/database_authenticatable_test.rb#L155

feliperenan:
Could you add a test for this change? We could add it closest to this one https://github.com/plataformatec/devise/blob/master/test/models/database_authenticatable_test.rb#L155

tegon:
@ihatov08 Sorry for taking so long to reply to you.
I've taken a look at your response, and it seems like we don't need the `options` argument at all. I say this because in your changes you added the following line:
```ruby
params.merge(options)
```
Which for the code example you sent
```ruby
User.first.update_with_password({current_password: 'password'}, password: 'new_password')
```
It will became a single hash: 
```ruby
params.merge(options)
=> {current_password: 'password', password: 'new_password'}
```
So, in the end, it's the same as if we pass one argument only:
```ruby
User.first.update_with_password(current_password: 'password', password: 'new_password')
```

I thought this was weird so I took a look at the code history, and the `options` argument was added to [support a feature present in Rails 3.1](https://github.com/plataformatec/devise/pull/1458), where it was possible to pass a [`as` option to specify a mass assignment scope](https://stackoverflow.com/questions/8796827/using-rails-3-1-as-admin-for-updating-attributes-protected-by-attr-accessib).
That option was then [deprecated in Rails 4](https://github.com/rails/rails/pull/7251) in favor of strong parameters.

Since we're supporting only Rails 4 and up, I think we can remove the `options` argument completely. But because `#update_password` is public API and people could be using in other ways, I think we should do the following:
1 - Change this PR to include a deprecation warning when the `options` atttribute is present, to warn people that this argument will be removed in v5;
2 - Create another PR that removes the attribute completely and merge into the v5 branch.

WDYT?

ihatov08:
@tegon 
Thank you for your reply.
The example I showed was not very good.

```
User.first.update_with_password({current_password: 'password'}, password: 'new_password') 
```

I think it is better to remove the options argument as well as you.

I think the second plan is good.
> 2 - Create another PR that removes the attribute completely and merge into the v5 branch.

ihatov08:
@tegon OK!
I will change this PR to add the deprication warning.

ihatov08:
@tegon Done!

tegon:
@ihatov08 Thanks!

I just noticed that there's another method doing a similar thing below: `#update_without_password`. Would you mind to add the deprecation warning there too?

ihatov08:
@tegon Fixed!

tegon:
Thanks!

Issue #4941, opened 2018-09-25
Cannot bootup rails (github HEAD version) with devise (github HEAD version)


oystersauce8:
After installing devise on a rails edge app using the steps mentioned in [this SO question](https://stackoverflow.com/questions/52487094/proper-way-of-installing-rails-head-version-with-devise#52487603), users get the error `Please specify the Rails release the migration was written for`

Steps to repro.

```
$ cd /tmp
$ ruby --version
ruby 2.3.3p222 (2016-11-21 revision 56859) [universal.x86_64-darwin17]
$ rbenv shell 2.5.1
$ ruby --version
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin17]
$ echo "source 'https://rubygems.org'" >> Gemfile
$ echo "gem 'rails', git: 'https://github.com/rails/rails.git'" >> Gemfile
$ bundle install
Fetching https://github.com/rails/rails.git
Fetching gem metadata from https://rubygems.org/.............
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/..........
Resolving dependencies...
Using rake 12.3.1
Using concurrent-ruby 1.0.5
Using i18n 1.1.0
Using minitest 5.11.3
Using thread_safe 0.3.6
Using tzinfo 1.2.5
Using activesupport 6.0.0.alpha from https://github.com/rails/rails.git (at master@0627442)
Using builder 3.2.3
Using erubi 1.7.1
Using mini_portile2 2.3.0
Using nokogiri 1.8.4
Using rails-dom-testing 2.0.3
Using crass 1.0.4
Using loofah 2.2.2
Using rails-html-sanitizer 1.0.4
Using actionview 6.0.0.alpha from https://github.com/rails/rails.git (at master@0627442)
Using rack 2.0.5
Using rack-test 1.1.0
Using actionpack 6.0.0.alpha from https://github.com/rails/rails.git (at master@0627442)
Using nio4r 2.3.1
Using websocket-extensions 0.1.3
Using websocket-driver 0.7.0
Using actioncable 6.0.0.alpha from https://github.com/rails/rails.git (at master@0627442)
Using globalid 0.4.1
Using activejob 6.0.0.alpha from https://github.com/rails/rails.git (at master@0627442)
Using mini_mime 1.0.1
Using mail 2.7.0
Using actionmailer 6.0.0.alpha from https://github.com/rails/rails.git (at master@0627442)
[...]

PR #4942, opened 2018-09-25
Issue #4941 , handle error 'Please specify the Rails release the migr…

oystersauce8:
Issue #4941 , handle error 'Please specify the Rails release the migration was written for'

### DIFF
--- a/lib/generators/active_record/devise_generator.rb
+++ b/lib/generators/active_record/devise_generator.rb
@@ -82,8 +82,8 @@ RUBY
         postgresql?
       end
 
-      def rails5?
-        Rails.version.start_with? '5'
+      def rails5_and_up?
+        Rails::VERSION::MAJOR >= 5
       end
 
       def postgresql?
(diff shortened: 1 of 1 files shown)


lib/generators/active_record/devise_generator.rb
```diff
@@ -82,8 +82,8 @@ def inet?
         postgresql?
       end
 
-      def rails5?
-        Rails.version.start_with? '5'
+      def rails5_and_up?
+        Rails::VERSION::MAJOR >= 5
```
rafaelfranca: This constant is private. Maybe `Rails.gem_version >= Gem::Version('5.0.0')`?

rafaelfranca: Ah, it is public, so to is fine.

rafaelfranca:
Can you also test devise with Rails 6. Right now it is not tested so I don't want to recommend people to use it.

oystersauce8:
My [new app](https://github.com/oystersauce8/untitled) worked ok I guess, basic [tests](https://github.com/oystersauce8/untitled/tree/master/test/integration) passed and running on heroku at https://untitled-d68f47d.herokuapp.com/

The only caveat was the responders gem still didn't "recognize" rails 6 - I sent a [PR](https://github.com/oystersauce8/responders/pull/1/files)

Issue #4939, opened 2018-09-21
idea: make deletion more explicit in method name #stored_location_for 


murb:
## Precheck

- Do not use the issues tracker for help or support, try Stack Overflow.
- For bugs, do a quick search and make sure the bug has not yet been reported
- Finally, be nice and have fun!

## Environment

- Ruby/Rails independent
- Devise 4.5.0

## Current behavior

Per documentation calling `#stored_location_for` deletes ('if it's navigational format') the url stored in the session.

## Expected behavior

IMHO calling a method with such name shouldn't remove the value without notice. Idea: create a method called `#pop_stored_location_for` (or something similar), and let `#stored_location_for` simply retrieve the current value. 

## Use case

Debugging a problem, I didn't expect `stored_location_for` to delete the value from the session. I assumed `stored_location_for` would simply be a getter where `store_location_for` is the setter. Turned out that `logger.debug stored_location_for(:user)` made things even harder to debug...

(happy to make a pull request if this change can be agreed upon)

tegon:
Hello @murb, thanks for noticing this.

I agree with you, this name can be misleading and cause unexpected issues. Feel free to create the pull request, I'd be happy to accept it.
Just one thing to notice is that since this method is public API, we'll have to include this changes in the next major version.

tegon:
We decided to not go forward with the refactor. More details in 
https://github.com/plataformatec/devise/pull/4946#issuecomment-449481906

PR #4946, opened 2018-09-27
Issue #4939: don’t delete by default when retrieving stored_location_for

murb:
First attempt :)

As described in issue #4939; I found the current behaviour unpredictable. This pull request introduces a new method `#pop_stored_location_for` that returns and deletes the location stored (on all cases, my idea was that by making this explicit it is no longer required to check `#is_navigational_format?`)

### DIFF
--- a/lib/devise/controllers/helpers.rb
+++ b/lib/devise/controllers/helpers.rb
@@ -213,7 +213,7 @@ module Devise
       #   end
       #
       def after_sign_in_path_for(resource_or_scope)
-        stored_location_for(resource_or_scope) || signed_in_root_path(resource_or_scope)
+        (is_navigational_format? ? pop_stored_location_for(resource_or_scope) : stored_location_for(resource_or_scope)) || signed_in_root_path(resource_or_scope)
       end
 
       # Method used by sessions controller to sign out a user. You can overwrite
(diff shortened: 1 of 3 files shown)


murb:
As for discussion: I don't know why the check for navigational_format? was introduced here. I don't think it is something such a method should be dealing with at all, and I'd argue that this can be perfectly decided upon per controller action.

murb:
not entirely happy with what I ended up with here... although it is a slight improvement by making popping the data more explicit by moving it up the call stack; the same issue still exists with `after_sign_in_path` (one would not assume from the method name that this adjusts state). 

To fix this it would require a larger effort affecting even more code; I'd flag all checks for `if_navigational_format?` in `*_path(resource)`-methods as smelly, and I'd create new methods to deal with redirects. Not sure whether that's the right path to move forward ...

tegon:
@murb I just finished to read your comments and the code. It turns out the rabbit hole was deeper than we initially though 😞 

I agree with your point, there are a lot of mixed concerns on those `_path_for` methods - not only `after_sign_in_path_for`. 
Sadly, it's going to be a big refactor to fix this, touching areas of the code where I don't know the reasons why they are there (like the `if_navigational_format?` check) and introducing a lot of breaking changes, since those methods are part of the public API.

That's why I will close this pull request. I'm sorry it didn't work out, but thank you for taking the time to tackling this.

PR #4955, opened 2018-10-16
Ignore useless files into the gem file.

huacnlee:
RubyGem only needs the`lib, app, config`, special it for avoid the useless file packing into the gem file.

### DIFF
--- a/devise.gemspec
+++ b/devise.gemspec
@@ -15,8 +15,7 @@ Gem::Specification.new do |s|
   s.description = "Flexible authentication solution for Rails with Warden"
   s.authors     = ['José Valim', 'Carlos Antônio']
 
-  s.files         = `git ls-files`.split("\n")
-  s.test_files    = `git ls-files -- test/*`.split("\n")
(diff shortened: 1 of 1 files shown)


huacnlee:
I have gem builded and install in local to test result:

```bash
$ gem build devise.gemspec
$ gem install devise-4.5.0.gem
$ gem open devise
```

devise.gemspec
```diff
@@ -15,8 +15,7 @@ Gem::Specification.new do |s|
   s.description = "Flexible authentication solution for Rails with Warden"
   s.authors     = ['José Valim', 'Carlos Antônio']
 
-  s.files         = `git ls-files`.split("\n")
-  s.test_files    = `git ls-files -- test/*`.split("\n")
```
huacnlee: unless remove this line, then test will still including gem file

tegon:
@rafaelfranca Any thoughts on this?

rafaelfranca:
👍 for me.

PR #4957, opened 2018-10-18
add in a pretty_print method for better pry output

dalexj:
Currently the output for any devise model is awful in pry. This adds in a copy of activerecord's pretty print but modified to use the same logic as the overriding `inspect` in devise to make sure it doesn't print out sensitive data.

most of this method is taken from https://apidock.com/rails/ActiveRecord/Core/pretty_print.

https://github.com/rweng/pry-rails/issues/83

### DIFF
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -125,6 +125,27 @@ module Devise
         "#<#{self.class} #{inspection.join(", ")}>"
       end
 
+      def pretty_print(pp)
+        pp.object_address_group(self) do
+          if defined?(@attributes) && @attributes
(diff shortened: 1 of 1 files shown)


tegon:
@dalexj thanks for the pull request. 

I don't really use pry, but it seems strange to me that Devise should implement a `#pretty_print` method - it seems like this should be a concern of the debugging libraries. Isn't this something that you can include in a local `.pryrc` file.

dalexj:
Activerecord creates this method, so it seems fitting that something sitting on top of activerecord like this would override it to create a more correct version. devise also overrides `inspect` so this is just adding to it

tegon:

Thank you!

PR #4975, opened 2018-11-21
fixed description for Devise::Generators::ControllersGenerator

kopylovvlad:
Hello. I found one little mistake in controllers generator's description. When I run command ```rails generate devise:controllers```  I see
```
For example:

  rails generate devise:controllers users -c=sessions

This will create a controller class at app/controllers/users/sessions_controller.rb like this:

  class Users::ConfirmationsController < Devise::ConfirmationsController
    content...
  end
```
You show how to overwrite sessions controller, but use the code of ConfirmationsController 😅


tegon:
Wow, nice catch!

Thanks, @kopylovvlad!

PR #4982, opened 2018-11-28
Do not keep the old email when replacing the email of an unconfirmed user

terryyin:
If the user hasn't  confirmed yet, it's not useful to keep both the previous email and the new email. The previously unconfirmed email will not be confirmed any way since the token is already changed. Leaving it there will only cause accidental troubles. The change allows the following: If the user is not confirmed yet, just replace the email. New token will still be generated and reconfirmation email will still be sent.

This change won't break existing data, where both email and unconfirmed email exist for an unconfirmed user. When confirmed, the unconfirmed email will still replace the email. So no worries.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -258,16 +258,20 @@ module Devise
         if Devise.activerecord51?
           def postpone_email_change_until_confirmation_and_regenerate_confirmation_token
             @reconfirmation_required = true
-            self.unconfirmed_email = self.email
-            self.email = self.email_in_database
+            if confirmed?
+              self.unconfirmed_email = self.email
+              self.email = self.email_in_database
+            end
             self.confirmation_token = nil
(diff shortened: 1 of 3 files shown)


tegon:
Hi @terryyin, thanks for the pull request.

I want to understand more of the reasons behind this change. When you said

> Leaving it there will only cause accidental troubles

Do you have any examples of the issues caused by this?

terryyin:
@tegon On our website the admin can see the unconfirmed account and try to fix their emails. Leaving a never confirmed email as the current email has confused us several times.

tegon:
@terryyin Got it. Well, if that's the case I think it's better to close this pull request and if in the future someone else find a functional issue because of this we can reopen and continue with the discussion.

Thanks!

terryyin:
Fair enough. I wish I have better test next time to document the problem I had next time. Thanks.

PR #4984, opened 2018-11-29
Skipping location storing if request xhr

r4do:
Hi guys!
I'm pretty sure that many of devise users use frontend frameworks, like vue or react with their own routers on frontend.
I've got the problem:
User is opening two tabs in same browser while logged in, and logs out in one of them (or session expires - will be same result). Then he forgets about this tab and tries to click on any link in other tab. Since there is no requests to backend needed to open other URL via frontend router, page is being opened and tries to get some data from backend (via json xhr request). After getting infinite loading because xhr request gets 302 redirect, user refreshes the page and being redirected to login page, where after successful login he sees page retrieved by xhr request, because it was saved by devise.
I think that the right solution for such requests is do not store location.

### DIFF
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -232,7 +232,7 @@ module Devise
     # authenticated yet, but we still need to store the URI based on scope, so
     # different scopes would never use the same URI to redirect.
     def store_location!
-      store_location_for(scope, attempted_path) if request.get? && !http_auth?
+      store_location_for(scope, attempted_path) if request.get? && !request.xhr? && !http_auth?
     end
 
     def is_navigational_format?


tegon:
Hi @r4do, thanks for the pull request.

First, I should say that devise does not support API-only applications out of the box. We've accepted some patches that could help with this - like the `is_flashing_format?` helper - but we don't guarantee it's compatible.

Back to the PR, your changes seems a lit bit hacky for me. When designing applications that use an API as a backend there's usually no session involved - so it shouldn't matter whether the URL is being stored or not.

Also, here:
>  After getting infinite loading because xhr request gets 302 redirect

Are you saying here that the frontend router is redirecting the user based on the response of the XHR request?

If you can provide a sample application reproducing the issue I think I would be able to help more. Right now I'm not sure how your application is designed and if we should add a patch in Devise or try to solve it in the app itself.

Thanks!

tegon:
> this is not about API-only applications. If you have some data which is being loaded via AJAX request, you'll get same result

Sorry, that's what I meant: non-HTML requests in general.

> No, devise is redirecting xhr request to login form

I'm afraid I don't understand exactly how the app is working. If you can send a sample application that would help me to find what's happening.

r4do:
hell, I've found the problem in my project: 
`config.http_authenticatable_on_xhr = false`
that's why I get redirect on xhr requests

PR #4987, opened 2018-12-13
Update trackable fields only in a database sign in

### LINKED ISSUE #4584: Trackable should be invoked on sign in, not on authentication
Hi,

I'm the developer of [devise-jwt](https://github.com/waiting-for-dev/devise-jwt). This gem provides with a warden strategy which authenticates a user if a valid JWT token is present in request headers.

Recently, a user came in with a pull request so that when devise-jwt strategy is used to authenticate a user trackable information is not updated (https://github.com/waiting-for-dev/devise-jwt/pull/23).

After studying it, I came to the conclusion that it is something that should be fixed from devise end.

Right now, [trackable behaviour is invoked from a warden `after_set_user` hook](https://github.com/plataformatec/devise/blob/88724e10adaf9ffd1d8dbfbaadda2b9d40de756a/lib/devise/hooks/trackable.rb). I think it is not correct. `update_tracked_fields!` should be invoked from the warden strategies that are meant to be used during a sign in, that is, from [database_authenticatable](https://github.com/plataformatec/devise/blob/master/lib/devise/strategies/database_authenticatable.rb). Warden strategies which are not used to sign in, like providing a token obtained during an actual sign in, should not be affected.

I think changing this should be very easy. We should just add one line of code to the validated user block in database_authenticatable strategy. I could prepare a pull request, but I would like to know beforehand whether it will be accepted or whether you think this could have some backward incompatibility issue I can't imagine right now.

Thanks!

waiting-for-dev:
Closes #4584 and waiting-for-dev/devise-jwt#23

### Bug
Users that use `devise-jwt`, will not have the correct behavior of trackable feature. As a request for APIs always requires authentication since there is no session in APIs world, Devise counts +1 on every request since  it contains authentication info.

It happens because Devise has a trackable hook that updates the trackable info everytime that the user is signed in by Warden.

### Fix
We are moving `update_trackable_fields!` from trackable hook (which was removed) to `sign_in_out` and `database_authenticatable`. This way, `update_trackable_fields!` is going to run only when the user signed in by Devise (only one time).


test/controllers/helpers_test.rb
```diff
@@ -270,6 +273,7 @@ def setup
     @controller.session[:user_return_to] = "/foo.bar"
     @mock_warden.expects(:user).with(:user).returns(nil)
     @mock_warden.expects(:set_user).with(user, scope: :user).returns(true)
+    @mock_warden.expects(:env).returns({})
```
tegon: Can we add a new test that checks that the `#update_tracked_fields!` method was called on `resource`?

lib/devise/controllers/sign_in_out.rb
```diff
@@ -51,6 +51,7 @@ def sign_in(resource_or_scope, *args)
           true
         else
           warden.set_user(resource, options.merge!(scope: scope))
+          resource.update_tracked_fields!(warden.request) if update_tracked_fields?(resource, warden)
```
tegon: The `#update_tracked_fields?` logic is being duplicated here and inside the `database_authenticatable` strategy. My concern is that in the future if there's a need to change this logic one of the places could be forgotten, possibly causing issues.
I'd say to move this conditional to `Devise::Models::Trackable#update_tracked_fields!`. We'll get the benefit of keeping the logic in the same place and also the caller wouldn't need to know about that conditional - i.e. "tell don't ask".
WDYT?

lib/devise/controllers/sign_in_out.rb
```diff
@@ -114,6 +115,12 @@ def expire_data_after_sign_in!
       end
 
       alias :expire_data_after_sign_out! :expire_data_after_sign_in!
+
+      def update_tracked_fields?(resource, warden)
+        resource.respond_to?(:update_tracked_fields!) &&
+          !warden.env['devise.skip_trackable'] &&
+          (!resource.respond_to?(:active_for_authentication?) || resource.active_for_authentication?)
```
tegon: The older hook was not checking this conditional, was it?
I thought that the `activatable` [hook](https://github.com/plataformatec/devise/blob/master/lib/devise/hooks/activatable.rb#L7) would solve this already.

tegon (commented):
@waiting-for-dev Thanks for tackling this!

I've made some comments, let me know what you think.

lib/devise/controllers/sign_in_out.rb
feliperenan: We need this check otherwise this test will fail:

https://github.com/plataformatec/devise/blob/76525939f659232b15f99ae356fb6262ae9bab2b/test/integration/trackable_test.rb#L72

I'm not sure, but I think it was being checked by `warden.authenticated?(options[:scope])` on the trackable hook.

waiting-for-dev: Yep, I have to look at it again, but I added it because some test was failing. But if @tegon says that this was already checked by `activatable` hook, maybe the problem is related with something not being stubbed in the test. I'll try to retake it as soon as I find the time, but if you or anybody can do it now please go ahead.

feliperenan: Could you give me permission to commit in this branch? I can continue the work that you've done here if you don't mind :)

waiting-for-dev: Done @feliperenan ! I have added you as a collaborator to the fork, so you now have push access to this branch. Thanks! :)

tegon: > I'm not sure, but I think it was being checked by warden.authenticated?(options[:scope]) on the trackable hook.

@feliperenan That makes sense. Is it possible to still do that? I'm not sure if we have access to the Warden instance or the resource's scope now, but I guess it would be safer to check whether the user is authenticated or not - e.g. in a speculative case where the `#sucess` method fails - but I might be overthinking 🤔 
Thoughts?

feliperenan: Looking into the old code, I think we don't need check for `warden.authenticated?` since the user should be already authenticated in the trackable context. Maybe this check was there because we were just being too defensive. Also, I removed it from trackable in the master branch and all tests still pass (I think we can rely on Devise test suite 😅).

tegon: Yeah, that makes sense. I guess we can continue with this implementation then. 👍

lib/devise/models/trackable.rb
```diff
@@ -46,6 +46,15 @@ def extract_ip_from(request)
...
       end
 
+      private
+
+      def skip_trackable_and_not_active_for_authentication?(request)
+        request.env['devise.skip_trackable'] || not_active_for_authentication?
+      end
+
+      def not_active_for_authentication?
+        self.respond_to?(:active_for_authentication?) && !self.active_for_authentication?
```
tegon: Can we ✂️ `self` here?
```suggestion
        respond_to?(:active_for_authentication?) && !active_for_authentication?
```

tegon (approved):

lib/devise/models/trackable.rb
feliperenan: Also, we don't need to ask if the resource `respond_to?(:active_for_authentication?)` since we already ask it in the caller.

waiting-for-dev:
Hey, great job there! Thanks for tackling it :smile:

BnayaZil:
Hi,
I'm not sure how to perform a migration for this modification, can I get any advice?
Should I just replace the `:trackable` hook with `:sign_in_out` and `:database_authenticatable`?

waiting-for-dev:
@BnayaZil usually you wouldn't need to do anything, as it is and implementation detail for the most common use case.

BnayaZil:
> @BnayaZil usually you wouldn't need to do anything, as it is and implementation detail for the most common use case.

What do u mean? I currently using `:trackable` as I want to collect the relevant data, should I change something? please find the following example:
```
class Admin
	include Mongoid::Document
	
	devise :database_authenticatable, :trackable, :timeoutable,
				:validatable, :rememberable

	## Trackable
	field :sign_in_count, type: Integer, default: 0
	field :current_sign_in_at, type: Time
	field :last_sign_in_at, type: Time
	field :current_sign_in_ip, type: String
	field :last_sign_in_ip, type: String

end
```
P.S- in the release notes written something about GDPR which have nothing to do with this PR (from what I read in the above description).

waiting-for-dev:
If you just want trackable fields to be updated when you sign in through database authentication, then you don't need to change anything. Check if expected behavior is still working in your case.

tegon:
@BnayaZil The release notes about GDPR are because from now on, the `trackable` module won't be included by default when a model is generated using Devise. Like you mentioned, this has nothing to do with this pull request.

The reason why this change is not in the release notes is that it's not released yet - it will be on the next major version (`5.0`). Before we cut that release we'll include some upgrade guides explaining how this change may affect your existing code. Like @waiting-for-dev mentioned, for most cases, this won't change anything, except if your application is relying on Warden callbacks.

BnayaZil:
I guess I just missed the part "module won't be included by default when a model is generated".
Thank you guys for being so responsive.

tegon:
@BnayaZil You're welcome! If you need anything else, feel free to ask. We're here to help 😄

rahulmr:
> Hi, I'm not sure how to perform a migration for this modification, can I get any advice? Should I just replace the `:trackable` hook with `:sign_in_out` and `:database_authenticatable`?

Hello, I am looking for an implementation for getting the last_sign_out for user, could you please let me know how this can be implemented? Is there any example? I want to show the last_sign_in_at and last_sign_out_at for an user in an user management app created using ruby on rails 6

Bilka2:
As a datapoint, we've been running this patch in production [since March 2026](https://github.com/otwcode/otwarchive/pull/5610) without problems.

carlosantoniodasilva:
@Bilka2 thanks. I don't know if what's in here is the right fix for the problem, but I haven't had time to stop and analyze if fully... the source issue is being tracked in the milestone as reopened though, hope to get to it at some point.

PR #4988, opened 2018-12-16
Removed extra characters

pestovpvl:
I noticed two extra characters, in my opinion they should not be there.


pestovpvl:
Hi!

I noticed two extra characters, in my opinion they should not be there. Maybe I'm wrong

tegon:
@pestovpvl Thanks! Looks like there are some of those in https://github.com/plataformatec/devise/blob/master/app/views/devise/shared/_links.html.erb too, can you also remove them?

PR #4996, opened 2018-12-27
 Make `#increment_failed_attempts` concurrency safe

### LINKED ISSUE #4981: Small security issue with :lockable
## Environment

- Ruby **any**
- Rails **any** running a multithreaded server (like puma)
- Devise **4.5.0**

## Current behavior
Pentesters found an issue where our users did not lock (using :lockable), despite running many many attempts to brute force the password.
To reproduce, try to login many times at exactly the same time (where your model is :lockable).
100 attempts within 1 milliseconds of each other will not increment the failed_attempts attribute on your user 100 times on a busy or slow database. They will most likely be incremented by approximately `100 % num_threads` where num_threads is the number of threads your server is configured to.

This is a test that describes the behaviour that I would reasonably expect:

```ruby
 describe '#increment_failed_attempts' do
   let(:user) { create :user }
   let(:same_user) { User.find(user.id) }
   it 'increments the count independently across instances' do
     expect {
       same_user.increment_failed_attempts
       same_user.save
       user.increment_failed_attempts
     }.to change { user.reload.failed_attempts }.by(2)
   end
 end
```
This test will fail in devise 4.5.0

## Expected behavior
The issue arises because we read and set failed_attempts in two steps, instead of in one database transaction.

Below is an excerpt of the method from `lockable.rb`, my comments added:
```ruby
def increment_failed_attempts
  self.failed_attempts ||= 0 # This line triggers a read in ActiveRecord if not already cached
  # some other thread now increments the counter to 1
  self.failed_attempts += 1 # We have already read the value 0, and will increment that erroneously to 1 again
end
```

Our workaround we use is something along the lines of:

```ruby
def increment_failed_attempts
  self.failed_attempts = self.class.connection.execute("UPDATE SET (failed_attempts = failed_attempts + 1) WHERE id = #{self.id} RETURNING failed_attempts").getvalue(0,0)
end
```

which passes the test above. This solution is postgres specific (and also assumes integer ids) so may not be suitable for devise, but the idea stands. Reading and writing from this attribute needs to happen on at least a row level transaction in the database.

tegon:
As reported in #4981, the method `#increment_failed_attempts` of `Devise::Models::Lockable` was
not concurrency safe. The increment operation was being done in two steps: first the value was read from the database, and then incremented by 1. This may result in wrong values if two requests try to update the value concurrently. For example:

```
Browser1 -------> Read `failed_attempts` from DB (1) -------> Increment `failed_attempts` to 2
    Browser2 -------> Read `failed_attempts` from DB (1) -------> Increment `failed_attempts` to 2
```

In the example above, `failed_attempts` should have been set to 3, but it will be set to 2. 

This pull request handles this case by calling `ActiveRecord::CounterCache.increment_counter` method, which will do both steps at once, reading the value straight from the database.

This pull request also adds a `ActiveRecord::AttributeMethods::Dirty#reload` call to ensure that the application gets the updated value - i.e. that other request might have updated. 
Although this does not ensure that the value is in fact the most recent one - other request could've updated it after the `reload` call - it seems good enough for this implementation. 
Even if a request does not locks the account because it has a stale value, the next one - that updated that value - will do it. That's why we decided not to use a pessimistic lock here.

Closes #4981.

### DIFF
--- a/lib/devise/models/lockable.rb
+++ b/lib/devise/models/lockable.rb
@@ -112,8 +112,7 @@ module Devise
       end
       
       def increment_failed_attempts
-        self.failed_attempts ||= 0
-        self.failed_attempts += 1
+        increment!(:failed_attempts)
       end
 
       def unauthenticated_message
--- a/lib/devise/models/lockable.rb
+++ b/lib/devise/models/lockable.rb
@@ -112,7 +112,8 @@ module Devise
       end
       
       def increment_failed_attempts
-        increment!(:failed_attempts)
+        self.class.increment_counter(:failed_attempts, id)
+        reload
       end
 
       def unauthenticated_message
--- a/test/models/lockable_test.rb
+++ b/test/models/lockable_test.rb
@@ -39,6 +39,19 @@ class LockableTest < ActiveSupport::TestCase
     end
   end
 
+  test "should read failed_attempts from database when incrementing" do
+    user = create_user
+    initial_failed_attempts = user.failed_attempts
+    same_user = User.find(user.id)
+
+    user.increment_failed_attempts
+    user.save
+    same_user.increment_failed_attempts
+    same_user.save
+
+    assert_equal initial_failed_attempts + 2, user.reload.failed_attempts
+  end
+
   test 'should be valid for authentication with a unlocked user' do
     user = create_user
     user.lock_access!
--- a/test/models/lockable_test.rb
+++ b/test/models/lockable_test.rb
@@ -45,9 +45,7 @@ class LockableTest < ActiveSupport::TestCase
     same_user = User.find(user.id)
 
     user.increment_failed_attempts
-    user.save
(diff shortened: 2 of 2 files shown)


glebsts:
Hello, @tegon 
Please excuse me for necroposting, but can this code in master be affected by same concurrency problem? 
https://github.com/plataformatec/devise/blob/master/lib/devise/hooks/lockable.rb
We are investigating similar behavior reported for latest version, so just in case..
Thank you.

tegon:
Hi @glebsts, the code you mentioned is for the other case: resetting the failed attempts account - i.e. unlocking a user.
I guess there's no problem in this case because if there are two concurrent requests they're both going to set the value to zero, and not increment like the original issue.
If you found a problem with this part of the code, please open a new issue so that we can discuss it better.

Thanks!

PR #5011, opened 2019-01-25
Fix SQLite3 warning

feliperenan:
Before setting this option, our test suite was giving the following warning:

```
DEPRECATION WARNING: Leaving `ActiveRecord::ConnectionAdapters::SQLite3Adapter.represent_boolean_as_integer` set to false is deprecated. SQLite databases have used 't' and 'f' to serialize boolean values and must have old data converted to 1 and 0 (its native boolean serialization) before setting this flag to true. Conversion can be accomplished by setting up a rake task which runs

  ExampleModel.where("boolean_column = 't'").update_all(boolean_column: 1)
  ExampleModel.where("boolean_column = 'f'").update_all(boolean_column: 0)

for all models and all boolean columns, after which the flag must be set to true by adding the following to your application.rb file:

  Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true
 (called from <top (required)> at $PATH/devise/test/rails_app/app/active_record/user.rb:5)
```

After configuring `represent_boolean_as_integer = true` as specified above, we don't have this warning anymore.

More info:
https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SQLite3Adapter.html#method-c-represent_boolean_as_integer

image


test/rails_app/config/application.rb
```diff
@@ -33,16 +33,15 @@ class Application < Rails::Application
     # config.assets.enabled = false
 
     config.action_mailer.default_url_options = { host: "localhost", port: 3000 }
-    rails_version = Gem::Version.new(Rails.version)
```
tegon: Is this removal related to the boolean warning mentioned? If they aren't I think it's better to keep those changes separated in another PR. Then we can take advantage of the warning-free output in the `master` branch too. 😄 
WDYT?

feliperenan: yeah, we could have it on master already without this change. As it isn't a big deal and affects only Rails 5.2+, I changed it and pointed to `5-rc` to not change things related to Rails 4.

I'm gonna remove this commit from this branch and commit only the last commit on `5-rc`.

PR #5028, opened 2019-02-21
 rememberable and timeoutable should work together

### LINKED ISSUE #4778: timeout and rememberable

fladson:
I took the liberty to fix the comments at #4926 and send as a new PR since that is inactive for days.
I hope that's ok with you @prithvi16

Fixes #4778

### DIFF
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -101,7 +101,7 @@ module Devise
       end
 
       def remember_expired?
-        remember_created_at.nil? || (remember_expires_at <= Time.zone.now.utc)
+        remember_created_at.nil? || (remember_expires_at <= Time.zone.now)
       end
 
       def remember_me?(token, generated_at)
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -101,7 +101,7 @@ module Devise
       end
 
       def remember_expired?
-        remember_created_at.nil? || (remember_expires_at <= Time.now.utc)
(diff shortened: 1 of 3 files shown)


fladson:
@tegon Could you review that again and retry the build please?

feliperenan:
Hi @fladson, thanks for the pull-request.

Are you able to reproduce this issue? because we aren't, see: https://github.com/plataformatec/devise/issues/4778#issuecomment-465718626.

prithvi16:
Thanks,  @fladson for the PR.  Sorry, I have been occupied with work and could not push any changes.

fladson:
@feliperenan I haven't tried to reproduce it actually but that's ok, you can close both PRs if you feel that this isn't needed anymore.

PR #5029, opened 2019-02-21
Adds feature require_password_to_destroy

### LINKED ISSUE #4816: Require password for account deletion by default

hayssac:
I took the liberty to fix conflicts and some comments at #4840.
Fixes #4816

cc @fladson 
@tegon Could you please review those changes again?

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -63,11 +63,23 @@ class Devise::RegistrationsController < DeviseController
 
   # DELETE /resource
   def destroy
-    resource.destroy
-    Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
-    set_flash_message! :notice, :destroyed
-    yield resource if block_given?
-    respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+    if resource.class.require_password_to_destroy
+      resource_destroyed = destroy_resource(resource, account_destroy_params)
+      if resource_destroyed
+        Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+        set_flash_message! :notice, :destroyed
+        yield resource if block_given?
+        respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+      else
+        render :edit
+      end
+    else
+      resource.destroy
+      Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+      set_flash_message! :notice, :destroyed
+      yield resource if block_given?
+      respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+    end
   end
 
   # GET /resource/cancel
@@ -92,6 +104,10 @@ class Devise::RegistrationsController < DeviseController
   # You can overwrite this method in your own RegistrationsController.
   def update_resource(resource, params)
     resource.update_with_password(params)
+  end
+
+  def destroy_resource(resource, params)
+    resource.destroy_with_password(params)
   end
 
   # Build a devise resource passing in the session. Useful to move
@@ -139,6 +155,10 @@ class Devise::RegistrationsController < DeviseController
 
   def account_update_params
     devise_parameter_sanitizer.sanitize(:account_update)
+  end
+
+  def account_destroy_params
+    devise_parameter_sanitizer.sanitize(:account_destroy)[:current_password]
   end
 
   def translation_scope
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -64,21 +64,14 @@ class Devise::RegistrationsController < DeviseController
   # DELETE /resource
   def destroy
     if resource.class.require_password_to_destroy
-      resource_destroyed = destroy_resource(resource, account_destroy_params)
-      if resource_destroyed
-        Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
-        set_flash_message! :notice, :destroyed
-        yield resource if block_given?
-        respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+      if destroy_resource(resource, account_destroy_params)
+        sign_out_resource(resource_name, resource)
       else
         render :edit
       end
     else
       resource.destroy
-      Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
-      set_flash_message! :notice, :destroyed
-      yield resource if block_given?
-      respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+      sign_out_resource(resource_name, resource)
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/controllers/devise/registrations_controller.rb
```diff
@@ -63,11 +63,16 @@ def update
...
+    if resource.class.require_password_to_destroy
+      if destroy_resource(resource, account_destroy_params)
+        sign_out_resource(resource_name, resource)
+      else
+        render :edit
+      end
+    else
+      resource.destroy
+      sign_out_resource(resource_name, resource)
+    end
```
sobrinho: Maybe:

```ruby
if destroy_resource(resource, account_destroy_params)
  sign_out_resource(resource_name, resource)
else
  render :edit
end
```

And then on destroy resource:

```ruby
def destroy_resource(resource, params)
  if resource.class.require_password_to_destroy
    resource.destroy_with_password(params)
  else
    resource.destroy
  end
end
```

lib/generators/templates/devise.rb
```diff
@@ -122,6 +122,8 @@
   # Send a notification email when the user's password is changed.
   # config.send_password_change_notification = false
 
+  #
+
```
sobrinho: Not needed.

test/integration/registerable_test.rb
```diff
@@ -271,6 +271,37 @@ def user_sign_up
...
+      get edit_user_registration_path
+
+      within(".destroy-password") do
+        fill_in 'current password', with: '87654321'
+      end
+
+      click_button "Cancel my account"
+      assert_contain "Current password is invalid"
+
+      assert_current_url '/users'
```
sobrinho: Maybe assert that the resource wasn't deleted, the opposite of line 286.

Not sure about the project code style but something like:

```ruby
refute User.to_adapter.find_all.empty?
```

app/controllers/devise/registrations_controller.rb
hayssac: Yeah I agree, that's way better. I will do these changes.

app/controllers/devise/registrations_controller.rb
```diff
@@ -165,4 +177,11 @@ def sign_in_after_change_password?
 
     Devise.sign_in_after_change_password
   end
+
+  def sign_out_resource(resource_name, resource)
+    Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
+    set_flash_message! :notice, :destroyed
+    yield resource if block_given?
```
sobrinho: A spec must have failed for that.

If not, it is missing.

We need to pass the block from destroy to this method.

test/integration/registerable_test.rb
feliperenan: here, I would only assert things that the User is seeing since it's seems a feature test and include (if there is no one) a unit test to ensure that user is deleted or not. So, I would suggest to keep it as it is and changing the assert above to follow the same approach.

sobrinho: In that case, the spec above would have to try to log in again and see that the account doesn't exist anymore.

app/controllers/devise/registrations_controller.rb
hayssac: We didn't have any tests for the `destroy action` but I'm having trouble to check the success case for this implementation :thinking:

test/integration/registerable_test.rb
hayssac: > here, I would only assert things that the User is seeing since it's seems a feature test and include (if there is no one) a unit test to ensure that the user is deleted or not. So, I would suggest to keep it as it is and changing the assert above to follow the same approach.

I inserted some tests for the destroy action, previously there were none, but I'm having a problem for the success case, working on it to see what can I do more.

hayssac: > In that case, the spec above would have to try to log in again and see that the account doesn't exist anymore.

Definitely a interesting thing to check, I'll improve this. But the `refute` works well too.

feliperenan: I'm ok if we use `refute` here to ensure that the user isn't created in this scenario.

I just point this out because, recently, we tried to move from webrat to capybara but we aborted because it requires a lot of changes. One of the reasons is that our test suite is mixing integration tests with features tests and webrat works a little different from capybara in this aspect. But reading the test again, I don't think it would be a problem at all.

app/controllers/devise/registrations_controller.rb
```diff
@@ -63,11 +63,11 @@ def update
 
   # DELETE /resource
   def destroy
-    resource.destroy
-    Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
-    set_flash_message! :notice, :destroyed
-    yield resource if block_given?
-    respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name) }
+    if destroy_resource(resource, account_destroy_params)
+      sign_out_resource(resource_name, resource)
```
sobrinho: The spec is failing because you aren't passing the block argument down, probably doing this will fix the spec:

```ruby
def destroy(&block)
  ...
  sign_out_resource(resource_name, resource, &block) # << here you pass the block down
  ...
end
```

app/views/devise/registrations/edit.html.erb
```diff
@@ -38,6 +38,21 @@
 
 <h3>Cancel my account</h3>
 
-<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?" }, method: :delete %></p>
+<% if resource.class.require_password_to_destroy %>
+  <%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { name: 'require_password_to_destroy', method: :delete }) do |f| %>
```
feliperenan: We still have problems with `non-unique id` pointed out by @tegon [here](https://github.com/plataformatec/devise/pull/4840/files#r238831376) . I think we don't need to change the `name` attribute, we should change the `id` instead: `html: { id: 'require_password_to_destroy', method: :delete }`, it should fix this problem :).

test/controllers/custom_registrations_controller_test.rb
```diff
@@ -35,6 +35,20 @@ class CustomRegistrationsControllerTest < Devise::ControllerTestCase
...
   end
 
+  test "yield resource to block on destroy failure" do
+    sign_in @user
+    delete :destroy, params: { user: { } }
+    assert @controller.destroy_block_called?, "destroy failed to yield resource to provided block"
+  end
+
+  test "yield resource to block on destroy success" do
+    swap Devise, require_password_to_destroy: true  do
```
feliperenan: This spec is falling on Rails 4.1 and 4.2. I ran this test without this `swap` and it worked on Rails 4.2 - I don't know why 🙈. Maybe we don't need this here.

mracos:
@hayssac are you still going to work on this?
I'm asking to know if I should assign it to someone else. 😄

PR #5031, opened 2019-02-22
Add config authentication_failure_route to customize redirect route

tinyfive:


test/failure_app_test.rb
```diff
@@ -232,6 +232,33 @@ def call_failure(env_params={})
...
+        assert_equal 'http://test.host/', @response.second['Location']
+      end
+    end
+
+    test 'returns to configured authentication_failure_route with lambda' do
+      swap Devise, authentication_failure_route: ->(_warden_options) { :new_user_registration_url } do
+        call_failure
+        assert_equal 302, @response.first
+        assert_equal 'You need to sign in or sign up before continuing.', @request.flash[:alert]
+        assert_equal 'http://test.host/users/sign_up', @response.second['Location']
```
awcodify: is it should be `/sign_in`?

tinyfive: @awcodify The `authentication_failure_route` set to a lambda return `:new_user_registration_url`, so `/sign_up` is correct.

colinross:
Currently, failures are handled by the failure_app which could be a full rack app, or just a lama defined in the routes (`devise_for`). This would allow you do whatever you need to do in case of a failure, including redirect.  I guess my point is I don't see why we would need a niche config setting for a more specialized case of what the generic failure app already handles.

test/failure_app_test.rb
```diff
@@ -232,6 +232,33 @@ def call_failure(env_params={})
...
+    test 'returns to configured authentication_failure_route with symbol' do
+      swap Devise, authentication_failure_route: :root_path do
+        call_failure
+        assert_equal 302, @response.first
+        assert_equal 'You need to sign in or sign up before continuing.', @request.flash[:alert]
+        assert_equal 'http://test.host/', @response.second['Location']
+      end
+    end
+
+    test 'returns to root if configured authentication_failure_route not exists' do
```
willianveiga, line 245: Isn't it better to raise a 404? Users can have a chance to fix it and we don't hide unexpected behavior. What do you think?

PR #5032, opened 2019-02-24
Use warden’s `authenticated?` for query methods to avoid side effects

### LINKED ISSUE #5013: allow_params_authentication! authenticates the user too
I'm facing the same issue described here https://github.com/plataformatec/devise/issues/4285.

I'm running the 4.5.0 of Devise.

Basically, the user is authenticated even when overriding the create method. I don't know if it's the intended behavior, but allow_params_authentication! prevents any authentication customizations.

I can skip this action, but I do believe that there is a deeper issue. The allow_params_authentication! does more than what its name describes.

JanBussieck:
Previously checks whether a certain scope is signed in were performed using warden’s
`authenticate?` or `authenticate` methods which would run the strategies and sign in the
scope if valid params were given. We want to remove this side effect from query methods.

Fixes #5013

### DIFF
--- a/lib/devise/controllers/helpers.rb
+++ b/lib/devise/controllers/helpers.rb
@@ -53,7 +53,7 @@ module Devise
 
             def #{group_name}_signed_in?
               #{mappings}.any? do |mapping|
-                warden.authenticate?(scope: mapping)
+                warden.authenticated?(scope: mapping)
               end
             end
 
@@ -119,7 +119,7 @@ module Devise
           end
 
           def #{mapping}_signed_in?
-            !!current_#{mapping}
+            !!(@current_#{mapping} || warden.authenticated?(scope: :#{mapping}))
           end
 
           def current_#{mapping}
--- a/lib/devise/controllers/sign_in_out.rb
+++ b/lib/devise/controllers/sign_in_out.rb
@@ -12,7 +12,7 @@ module Devise
       # authentication hooks, you can directly call `warden.authenticated?(scope: scope)`
       def signed_in?(scope=nil)
         [scope || Devise.mappings.keys].flatten.any? do |_scope|
-          warden.authenticate?(scope: _scope)
+          warden.authenticated?(scope: _scope)
         end
       end
 
--- a/test/controllers/helpers_test.rb
+++ b/test/controllers/helpers_test.rb
@@ -16,20 +16,20 @@ class ControllerAuthenticatableTest < Devise::ControllerTestCase
   end
 
   test 'proxy signed_in?(scope) to authenticate?' do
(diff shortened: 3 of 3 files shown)


lib/devise/controllers/helpers.rb
```diff
@@ -119,7 +119,7 @@ def authenticate_#{mapping}!(opts={})
           end
 
           def #{mapping}_signed_in?
-            !!current_#{mapping}
+            !!(@current_#{mapping} || warden.authenticated?(scope: :#{mapping}))
```
ferrl: The `current_#{mapping}` method still needs to call the `warden.authenticate` method, so if we keep calling it here we will still sign in the user if there are valid params. 

Should this method only call `warden.authenticated?(scope: :#{mapping})` or am I missing something?

tegon: I think that's correct, we should probably change `#current_#{mapping}` too.

Now I think it'd be good to have a test for that - ensure we're not signing the user in when these methods are called - too.

JanBussieck: We are not actually calling `#current_#{mapping}`, but simply check whether the ivar `@current_#{mapping}` is set, so `warden.authenticate` won't be invoked.
I think that the decision of whether or not `#current_#{mapping}` should run authentication as well, falls outside the scope of this PR, since it explicitly concerns query methods.
Thanks for your reviews, I will take care of the tests!

mracos:
@JanBussieck are you still going to work on this?
I'm asking to know if I should assign it to someone else. 😄

JanBussieck:
@mracos Sorry for the long absence, I'll finish this up today! Thanks for the ping.

JanBussieck:
I took another look at the test today and after some consideration, I am not sure what at test that checks that a resource is not being authenticated when `resource_signed_in?` is called would look like or whether, indeed, such a test would make very much sense. After all if there were any other method called on warden that could lead to authenticating a resource we would get an error since we only mock `authenticated?` on the warden mock.
Am I missing something?

tegon:
> After all if there were any other method called on warden that could lead to authenticating a resource we would get an error since we only mock authenticated? on the warden mock.

@JanBussieck that makes sense. I guess we are good with the current tests then.

mracos:
@JanBussieck I just rebased `5-rc` branch with `master`, can you rebase your branch with `5-rc`, please? 😄

JanBussieck:
@mracos Done :)

mracos:
Thanks for your contribution @JanBussieck. 💙

JanBussieck:
Thank you for your hard work on this gem!

Marcos Ferreira <notifications@github.com> schrieb am Do. 19. Sep. 2019 um
19:33:

PR #5036, opened 2019-02-26
Forget me should not clear the remember_token when expire_all_remembe…

JanBussieck:
…r_me_on_sign_out is enabled

### DIFF
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -57,7 +57,7 @@ module Devise
       # it exists), and save the record without validations.
       def forget_me!
         return unless persisted?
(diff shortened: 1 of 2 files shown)


JanBussieck:
I don't really understand for what reason the CI is failing here. Maybe someone from core could help me out.

feliperenan:
The CI broke because it couldn't install a higher nokogiri version because it requires Ruby 2.3+. I was able to fix that just re-running tests on CI but I also did this commit https://github.com/plataformatec/devise/commit/11026007206226c1189f6050ab05d2284f47a669 to avoid it in future PRs.

JanBussieck:
Is there anything missing from this PR? What can I do to help you move this along?

PR #5039, opened 2019-03-10
use created_at for checking active_for_authentication

nhattan:


nhattan:
Maybe it will solve this issue: https://github.com/plataformatec/devise/issues/4536

feliperenan:
Hi @nhattan, thanks for the pull request.

Could you explain the motivation behind this PR? This is going to help us understand what this patch is solving and make the properly code review.

Thanks :)

nhattan:
@feliperenan I updated the description. Please help to take a look. Thanks!

rubyrider:
I think created at is ambiguous to consider for this scenario. So what will be the case for the 'Shadow' User who are created from leads or other source and they become real user later ? I think the it should be done considering the confirmation_sent_at date instead of created_at timestamp.

nhattan:
@rubyrider Unfortunately, `confirmation_sent_at` gets updated to `Time.now.utc` every time user triggers a `resend_confirmation_instructions` which causes the issue, using `created_at` would help to solve it. In case `created_at ` is ambiguous, WDYT if we add a separate column called `first_confirmation_sent_at`? @feliperenan

rubyrider:
> @rubyrider Unfortunately, `confirmation_sent_at` gets updated to `Time.now.utc` every time user triggers a `resend_confirmation_instructions` which causes the issue, using `created_at` would help to solve it. In case `created_at ` is ambiguous, WDYT if we add a separate column called `first_confirmation_sent_at`? @feliperenan

I would suggest to create a field `confirmation_resend_at` and `confirmation_sent_count` instead of `first_confirmation_sent_at` ?

Issue #4696, opened 2017-11-07
sign_in IntegrationHelper produces errors when trying to load a session


RocketPuppy:
Rails 5 API mode, devise cookie auth works just fine in the actual application, meaning all middleware and initializers are set up properly.

I have an ApiController that looks like this:
```ruby
class ApiController < ActionController::API
end
```
And all my controllers inherit from it. I have multiple user scopes. The integration test helpers are set up according to the documentation:

```ruby
# test/test_helper.rb

class ActionDispatch::IntegrationTest
  include Devise::Test::IntegrationHelpers
end
```

In a test:

```ruby
class FooControllerTest < ActionDispatch::IntegrationTest
  test "should be successful" do
    sign_in CustomUser.first
    get foo_url, xhr: true, as: :json

    assert(@response.successful?) # fail
  end
end
```

In the response body I get this error message, implying things are failing even before they get to my controllers.

```
> undefined method `[]=' for nil:NilClass

actionpack (5.1.4) lib/action_dispatch/request/session.rb, line 217
-------------------------------------------------------------------


  212             load! unless loaded?
  213           end
  214
  215           def load!
  216             id, session = @by.load_session @req
> 217             options[:id] = id
  218             @delegate.replace(stringify_keys(session))
  219             @loaded = true
  220           end
  221
  222           def stringify_keys(other)


App backtrace
-------------

 - actionpack (5.1.4) lib/action_dispatch/request/session.rb:217:in `load!'
 - actionpack (5.1.4) lib/action_dispatch/request/session.rb:212:in `load_for_write!'
 - actionpack (5.1.4) lib/action_dispatch/request/session.rb:197:in `merge!'
 - actionpack (5.1.4) lib/action_dispatch/request/session.rb:17:in `create'
 - actionpack (5.1.4) lib/action_dispatch/middleware/session/abstract_store.rb:71:in `prepare_session'
 - rack (2.0.3) lib/rack/session/abstract/id.rb:231:in `context'
 - rack (2.0.3) lib/rack/session/abstract/id.rb:226:in `call'
 - actionpack (5.1.4) lib/action_dispatch/middleware/cookies.rb:613:in `call'
 - warden (1.2.7) lib/warden/manager.rb:36:in `block in call'
 - warden (1.2.7) lib/warden/manager.rb:35:in `catch'
 - warden (1.2.7) lib/warden/manager.rb:35:in `call'
 - rack (2.0.3) lib/rack/etag.rb:25:in `call'
[...]

PR #5041, opened 2019-03-14
Added mention of API mode complications to README

emersonthis:
Related to #4696

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ It's composed of 10 modules:
 	- [ActiveJob Integration](#activejob-integration)
 	- [Password reset tokens and Rails logs](#password-reset-tokens-and-rails-logs)
 	- [Other ORMs](#other-orms)
+	- [Rails API mode](#rails-api-mode)
 - [Additional information](#additional-information)
 	- [Heroku](#heroku)
 	- [Warden](#warden)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -694,6 +695,17 @@ config.log_level = :warn
 
 Devise supports ActiveRecord (default) and Mongoid. To select another ORM, simply require it in the initializer file.
 
+### Rails API Mode
+
+Rails 5+ has a built-in "[API Mode](https://edgeguides.rubyonrails.org/api_app.html)" which optimizes Rails for use as an API (only). One of the side effects is that it changes the order of the middleware stack, and this can cause problems for `Devise::Test::IntegrationHelpers`. This problem usually surfaces as an ```undefined method `[]=' for nil:NilClass``` error when using integration test helpers, such as `sign_in()`. The solution is simply to reorder the middlewares by adding the following to test.rb:
```
tegon:
```suggestion
Rails 5+ has a built-in [API Mode](https://edgeguides.rubyonrails.org/api_app.html) which optimizes Rails for use as an API (only). One of the side effects is that it changes the order of the middleware stack, and this can cause problems for `Devise::Test::IntegrationHelpers`. This problem usually surfaces as an ```undefined method `[]=' for nil:NilClass``` error when using integration test helpers, such as `#sign_in`. The solution is simply to reorder the middlewares by adding the following to test.rb:
```

README.md
```diff
@@ -694,6 +695,17 @@ config.log_level = :warn
...
+
+```
```
tegon:
```suggestion
```ruby
```

tegon:
@emersonthis Thank you!

PR #5043, opened 2019-03-14
Backport CVE-2019-5421 fix to 3.x

ouranos:
See #4996

### DIFF
--- a/test/models/lockable_test.rb
+++ b/test/models/lockable_test.rb
@@ -37,6 +37,17 @@ class LockableTest < ActiveSupport::TestCase
     end
   end
 
+  test "should read failed_attempts from database when incrementing" do
+    user = create_user
+    initial_failed_attempts = user.failed_attempts
+    same_user = User.find(user.id)
+
+    user.valid_for_authentication?{ false }
+    same_user.valid_for_authentication?{ false }
+
+    assert_equal initial_failed_attempts + 2, user.reload.failed_attempts
+  end
+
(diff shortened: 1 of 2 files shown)


test/models/lockable_test.rb
```diff
@@ -37,6 +37,17 @@ def setup
     end
   end
 
+  test "should read failed_attempts from database when incrementing" do
+    user = create_user
+    initial_failed_attempts = user.failed_attempts
+    same_user = User.find(user.id)
+
+    user.valid_for_authentication?{ false }
```
tegon: Just out of curiosity, what's the benefit of calling `#valid_for_authentication?` instead of `#user.increment_failed_attempts`?

ouranos: Afaik `increment_failed_attempts` is not available in the `3-stable` branch.
The incrementation is actually done in the [`valid_for_authentication?` method](https://github.com/plataformatec/devise/blob/3-stable/lib/devise/models/lockable.rb#L103)

tegon: Oh, that's correct. That's my bad, sorry 😞

dy-cs:
is there a timeline for when this fix will be released?

tegon:
@dy-cs There are no plans to release a version since 3.x version is not officially maintained anymore. We do accept those backport pull requests so that people can grab the Gem directly from GitHub pointing to the `3-stable` branch.

dy-cs:
@tegon gotcha. thanks for the reply!

Issue #5038, opened 2019-03-07
`#valid_password?` returns false when updating resource with empty password fields


codeminator:
## Environment

- Ruby **2.4.4**
- Rails **5.2.2**
- Devise **4.6.1**

## Description

In prior versions of devise (I tested with `v.4.4.2` since, it's the earliest version compatible with Rails `~> 5.2`), it was possible to call `update` statement on a resource with some fields, while submitting `password` and `password_confirmation` fields as empty strings, and still old password is valid:
```
> pw = 'thisIsPassword'
> user.update(password: pw, password_confirmation: pw)
User Update (7.5ms)  UPDATE `users` SET `updated_at` = '2019-03-07 10:18:45', `encrypted_password` = '$2a$10$mh8fvYDEujHZrDIAPao8k.rj.EweEwT/uf4i/Ax71d0oSMYYtvUHe' WHERE `users`.`id` = 5
   (1.4ms)  COMMIT
=> true
> user.valid_password?(pw)
=> true
user.update(email: "another_email@mail.com", password: "", password_confirmation: "")
   (0.4ms)  BEGIN
  User Update (0.7ms)  UPDATE `users` SET `email` = 'another_email@mail.com', `updated_at` = '2019-03-07 10:19:40' WHERE `users`.`id` = 5
   (3.9ms)  COMMIT
=> true
> user.valid_password?(pw)
=> true
```

While in `v.4.6.1` (the latest), `#valid_password?` returns `false` when doing the same steps (Basically, it sets `encrypted-password` column with `NULL`):
```
> pw = 'thisIsPasswordTest'
=> "thisIsPasswordTest"
> user.update(password: pw, password_confirmation: pw)
   (0.2ms)  BEGIN
  User Update (0.5ms)  UPDATE `users` SET `updated_at` = '2019-03-07 10:21:57', `encrypted_password` = '$2a$10$uRBpZWVLF2epl3J.apIiZ.KdQ9hFM0Qmn53hMpmNNgvUccCyjmPnq' WHERE `users`.`id` = 5
   (2.0ms)  COMMIT
=> true
> user.valid_password?(pw)
=> true
> user.update(email: "another_email_updated@test.com", password: "", password_confirmation: "")
   (0.2ms)  BEGIN
  User Update (0.4ms)  UPDATE `users` SET `email` = 'another_email_updated@test.com', `updated_at` = '2019-03-07 10:23:13', `encrypted_password` = NULL WHERE `users`.`id` = 5
[...]

Issue #5033, opened 2019-02-25
v4.6.0 breaks activeadmin forms that have `password` field


gsar:
## Environment

- Ruby 2.6.1
- Rails 5.2.2
- Devise 4.6.1

## Current behavior

It appears that PR #4261 introduced a change where `encrypted_password` could be set to `nil` even when no attempt was made to change `password`.  The relevant comment is here https://github.com/plataformatec/devise/pull/4261/files#r77820245

This results in user model forms that have a password field in activeadmin to break with a validation failure when the password field remains empty on submit:

```
ActiveRecord::NotNullViolation (PG::NotNullViolation: ERROR:  null value in column "encrypted_password" violates not-null constraint
```

## Expected behavior

No attempt should be made to set `encrypted_password` to `nil` if the `password` field doesn't have a value.  This was the case in devise versions prior to v4.6.0.

gsar:
In case it helps others, I am using this workaround in the activeadmin pages that have an editable devise password field:

```
  before_save do |user|
    if user.will_save_change_to_encrypted_password?
      user.restore_encrypted_password! unless user.encrypted_password.present?
    end
  end
```

zucler:
This is occurring in our application as well. Please fix.

chiperific:
Same issue here.
RailsAdmin form for user with `password` and `password_confirmation` fields won't submit, throws this error.

@gsar's monkey patch works.

phlegx:
I have the same problem on my application. I want to update a user without changing the password but some other fields. I get a validation error: `encrypted_password can't be blank`

nosretep:
Broke for me too. Just add `encrypted_password: ""` and that should satisfy the not null, and is also the default value of "".

`t.string :encrypted_password, null: false, default: ""` from the migration

phlegx:
@nosretep your solution breaks logins! If I submit an account form to update the account and leave `password` and `password_confirmation` empty, then the account should be updated without set `encrypted_password` to blank! If it get set to blank, then the account can no more login with his password!

nosretep:
@phlegx works for me on my specs, login seems to have not broken for me. I'll keep on the lookout though.

Alexey-Lukin:
> In case it helps others, I am using this workaround in the activeadmin pages that have an editable devise password field:
> 
> ```
>   before_save do |user|
>     if user.will_save_change_to_encrypted_password?
>       user.restore_encrypted_password! unless user.encrypted_password.present?
>     end
>   end
[...]

PR #5044, opened 2019-03-15
Do not set encrypted_password to nil if password is blank

timkrins:
If the password attribute was set to a blank or `nil`, this was also setting the `encrypted_password` to `nil` (password_digest returns `nil` for a blank)

This means that calling `update` on a model with a blank password always sets `encrypted_password` to `nil`, which usually errors as a database constraint issue.

Introduced by https://github.com/plataformatec/devise/pull/4261/files#diff-19b4dd928714c72f1338874351e8ff2dR40

Related to https://github.com/plataformatec/devise/issues/5033 and https://github.com/plataformatec/devise/issues/5038

### DIFF
--- a/lib/devise/models/database_authenticatable.rb
+++ b/lib/devise/models/database_authenticatable.rb
@@ -60,7 +60,7 @@ module Devise
       # the hashed password.
       def password=(new_password)
         @password = new_password
-        self.encrypted_password = password_digest(@password) 
+        self.encrypted_password = password_digest(@password) if @password.present?
       end
 
       # Verifies whether a password (ie from sign in) is the user password.
--- a/test/models/database_authenticatable_test.rb
+++ b/test/models/database_authenticatable_test.rb
@@ -122,6 +122,12 @@ class DatabaseAuthenticatableTest < ActiveSupport::TestCase
     assert_nil new_user(password: '').encrypted_password
   end
 
+  test 'should not set encrypted password to nil if updating with blank password' do
+    user = create_user
+    user.password = ''
+    refute_nil user.encrypted_password
+  end
+
   test 'should hash password again if password has changed' do
     user = create_user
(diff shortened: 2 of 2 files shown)


timkrins:
Hmm, need to determine thought behind original change. (reasons for keeping 'password' in sync with encrypted_password)

JulienItard:
Hello, indeed now i have database constraint issue `PG::NotNullViolation: ERROR:  null value in column "encrypted_password" violates not-null constraint`

hayesr:
Also having the problem @JulienItard is having in my specs. I'm not setting a blank password, so I'm not sure if this is really the issue, but others are pointing to it: https://github.com/plataformatec/devise/issues/4981#issuecomment-473419369

hayesr:
It seems that this issue may also have relevant info: https://github.com/plataformatec/devise/issues/5033

hayesr:
@timkrins My bet is that the original thought was to allow for changing other attributes on user records without changing the password.

NikoRoberts:
Yeah everyone has to put additional logic in to get the same behaviour now. It was a very common pattern to rely on Devise for the logic to ignore a blank password update.

It looks based on `allow_blank: true` that this was an intentional feature.

I’m not sure if this behaviour was taken into account when the PR was made.

NikoRoberts:
@timkrins regarding the impetus for the change. Why would password need to be in sync with encrypted password if it isn’t persisted?
The point of the accessor is that it is only used as a stepping stone for validation and should never represent what is in encrypted password.

tegon:
We're going to revert the change since there were no real exploits based on the attributes being unsynced. We thought this would be a simple change but since it's causing so much trouble for existing applications, we're going to revert it. 
See https://github.com/plataformatec/devise/issues/5033#issuecomment-476386275

Thanks.

PR #5046, opened 2019-03-18
Add SimpleCov for code coverage analysis

debbbbie:
## Testing Results
image

## To-Dos
- [ ] Display test coverage result to `codeclimate` (I can do this)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ group :test do
   gem "timecop"
   gem "webrat", "0.7.3", require: false
   gem "mocha", "~> 1.1", require: false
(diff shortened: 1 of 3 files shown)


feliperenan:
Thanks for the pull-request.

We already have the culture to ask for tests when a new change is proposed without one, and a report doesn't ensure that things are being tested correctly - in fact, it points out the lines that don't have coverage and it could be useful. But Devise is a well-tested framework and we've achieved that without a report like simplecov, so I don't think we need this tool in Devise today.

Thanks again for the proposal though.

Issue #4245, opened 2016-08-10
:database_authenticatable issue with clean_passwords


kaelumania:
Within the :database_authenticatable there is an issue at [Line 40](https://github.com/plataformatec/devise/blob/master/lib/devise/models/database_authenticatable.rb#L40)

Whenever the passwords are cleaned (set to `nil`), e.g. in the [RegistrationsController](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/registrations_controller.rb#L30) the encrypted_password remains dirty, which can have heavy security affecting side-effects.

I propose to set the encrypted_password also to `nil` or to revert to its original value. Here is a little test scenario:

``` ruby
model.password = 'abc'
model.password = nil

model.valid_password? 'abc' => true
```

lucasmazza:
@kaelumania sounds reasonable, can you please send a Pull Request for that? Thanks! 😄

sivagollapalli:
@lucasmazza Should we need reset `encrypted_password` to `nil` if the password is `nil` ? If yes, I think we need to remove `NOT NULL constraint` for `encrypted_password` from migration. I just want to confirm prior implementation.

lucasmazza:
@sivagollapalli no, we don't need to remove the constraint - from @kaelumania's report the goal is to ensure that both attributes are properly in sync, and not let a null password be persisted by the default.

sivagollapalli:
@lucascolucci So, the above test case should return `false` but currently, it returns `true`. Am I right?

kaelumania:
@sivagollapalli yes, I think you are right. We use devise `database_authenticatable` while allowing `nil` passwords. Because not every entity in our system has a password or supports access by password. Therefore our barrier to reject access when NO (`nil`) password is set, is line https://github.com/plataformatec/devise/blob/master/lib/devise/encryptor.rb#L13 that states if the password is `nil` it is not valid!

tegon:
Closed via #4261

PR #5048, opened 2019-03-20
Remove Valid Password Constraints from DatabaseAuthenticatable Concern

tagCincy:
# Overview
PR #4261 and subsequently #4970 were added to address issue #4245.  In doing so, a hard requirement for having a non-nil, non-blank password was injected into `DatabaseAuthenticatable`.  As many of us are upgrading from version < 4.6 to 4.6.1 to address https://github.com/plataformatec/devise/issues/4981 we are unable to do so because we have a pre-existing implementation that allows the creation of a Devise model with no password.  A common use case of this is pre-creating an account when a user submits an email for marketing purposes.

However, in my opinion, a valid use case is irrelevant to what this change did.  It broke the nice separation of concerns that existed in Devise prior to version 4.6.0.  `DatabaseAuthenticatable` should not be deciding what constitutes validity for a password, that is what the `Validatable` concern is for.  Given that `Validatable#password_required?` exists, which allows implementations to override default behavior and explicitly not require a password, is further evidence why this change was a mistake.

~Furthermore, the original issue that sparked this change, #4245, was either inaccurate to begin with, or had been properly fixed in 2+ years since being reported and the change being merged.  I have run the same "test" on an implementation using this branch:~

UPDATE: A colleague of mine pointed out that my change was different than the pre-4.6 behavior and that is why the test case from #4245 was passing.  I updated my change to set `encrypted_password` to a blank string if the passed password argument is blank as well.  This should mimic the pre-4.6 behavior as well as keeping the "default" `encrypted_password` value the same as defined in the schema. 

# Changelog
## Changes
- removed returning `nil` from `DatabaseAuthenticatable#password_digest` if the passed password argument is `nil` or `blank`
- returns `''` if a blank password argument is passed
- adds a simple unit test to match the test case from the original issue (#4245)
## Removed
- unit tests referencing the prior behavior

### DIFF
--- a/lib/devise/models/database_authenticatable.rb
+++ b/lib/devise/models/database_authenticatable.rb
@@ -198,7 +198,6 @@ module Devise
       # See https://github.com/plataformatec/devise-encryptable for examples
       # of other hashing engines.
       def password_digest(password)
-        return if password.blank?
         Devise::Encryptor.digest(self.class, password)
       end
 
--- a/lib/devise/models/database_authenticatable.rb
+++ b/lib/devise/models/database_authenticatable.rb
@@ -38,7 +38,7 @@ module Devise
       def initialize(*args, &block)
         @skip_email_changed_notification = false
         @skip_password_change_notification = false
-        super 
+        super
       end
 
       # Skips sending the email changed notification after_update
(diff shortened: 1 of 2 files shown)


djfpaagman:
+1 here, running into the exact same issue.

JulienItard:
Any idea when this will be merged ? I got same issue ! Thanks a lot 😄

tegon:
Hi @tagCincy, thanks for submitting this pull request but we decided to just revert original commit in a new pull request (https://github.com/plataformatec/devise/pull/5051) in order to keep the code the same as it was before.

Thanks again!

PR #5049, opened 2019-03-20
Regenerate confirmation token

jcgrethe:
Very simple functionality. I want to resend the confirmation instruction with a new token. Same code as resend but it put the `confirmation_token = nil` so i force the `send_confirmation_token` to generate a new one. The reason why i propose this new feature is because i need to deactivate the old confirmation token for security.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -130,6 +130,15 @@ module Devise
       end
 
       # Resend confirmation token.
+      # Regenerates the token.
+      def regenerate_confirmation_instructions
+        pending_any_confirmation do
+          self.confirmation_token = nil
+          send_confirmation_instructions
+        end
+      end
+
+      # Resend confirmation token.
       # Regenerates the token if the period is expired.
       def resend_confirmation_instructions
         pending_any_confirmation do
--- a/test/models/confirmable_test.rb
+++ b/test/models/confirmable_test.rb
@@ -343,6 +343,14 @@ class ConfirmableTest < ActiveSupport::TestCase
     end
   end
 
+  test 'generate a new token on regenerate' do
+    user = create_user
+    old  = user.confirmation_token
+    user = User.find(user.id)
+    user.regenerate_confirmation_instructions
+    assert_not_equal user.confirmation_token, old
+  end
+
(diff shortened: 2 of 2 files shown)


tegon:
Hi, @jcgrethe thanks for your proposal.

Can you explain when do you need to deactivate the old token?
We already have a configuration option to automatically expire the token after a time (`confirm_within`) and every time the user requests a new one the old will be deactivated. These options are not enough for your case?

jcgrethe:
Hi @tegon, thanks for answering.
In the documentation it said that:

>   Confirmable will not generate a new token if a repeat confirmation is requested
  during this time frame, unless the user's email changed too.
(https://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Confirmable) 

This is why i think that although it is within time, it is better to send a new one. The user will have more time to confirm and the old mail with the old confirmation token its expired.

tegon:
Hmm, that's correct, I totally confused `confirmation_token` with `reset_password_token`, my bad.

I'm a little confused now because you added the `#regenerate_confirmation_instructions` method but I can't find it being used anywhere in the code. At which point you think this will be used?

Also, can you explain your security concerns in more details - i.e. what's the possible exploit and how your solution solves it?

jcgrethe:
Hi again. I did this because if the user update the email i want to send a new confirmation token.  At that time, i didn't know that Confirmable will not generate a new token unless the user's email changed too.  But this was the first thing i tried and i think its not working. For example:

```
user=User.create(email:"first@email.com", password:"mypass")
irb(main):011:0> user.confirmation_token
=> "jxFTqHmKDp5qBt_QxpFD"
irb(main):012:0> user.update_attribute(:email, "second@email.com")
irb(main):013:0> user.reload
irb(main):014:0> user.send_confirmation_instructions
irb(main):015:0> user.confirmation_token
=> "jxFTqHmKDp5qBt_QxpFD"
 
```

I used this to force devise to change the token if the user's email changed, but if its a bug i can close this pull request and open an issue.

Also other feature is that for example the time for the token to be invalidated its very short, I can manually send a new mail with a new token. The user will have more time to confirm and I prevent the user from entering an email with an invalid confirmation token.

PR #5050, opened 2019-03-24
Fix dead link

softwaregravy:
The link that's in there goes to a non-existent page. I'm guessing it was moved. I put the link in that I think it was moved to.


tegon:
Thanks!

PR #5051, opened 2019-03-25
Reverts both "[#4245] Allow password to nil (#4261)" and "Add more tests (#4970)"

### LINKED ISSUE #5033: v4.6.0 breaks activeadmin forms that have `password` field

mracos:
Closes #5033 

More info in https://github.com/plataformatec/devise/issues/5033#issuecomment-476386275.

### DIFF
--- a/lib/devise/models/database_authenticatable.rb
+++ b/lib/devise/models/database_authenticatable.rb
@@ -65,6 +65,7 @@ module Devise
 
       # Verifies whether a password (ie from sign in) is the user password.
       def valid_password?(password)
+        return false if password.blank?
         Devise::Encryptor.compare(self.class, encrypted_password, password)
       end
 
--- a/lib/devise/models/database_authenticatable.rb
+++ b/lib/devise/models/database_authenticatable.rb
@@ -60,18 +60,17 @@ module Devise
       # the hashed password.
       def password=(new_password)
         @password = new_password
-        self.encrypted_password = password_digest(@password) 
+        self.encrypted_password = password_digest(@password) if @password.present?
       end
 
       # Verifies whether a password (ie from sign in) is the user password.
       def valid_password?(password)
-        return false if password.blank?
         Devise::Encryptor.compare(self.class, encrypted_password, password)
       end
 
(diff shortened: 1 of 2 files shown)


lib/devise/models/database_authenticatable.rb
```diff
@@ -60,17 +60,18 @@ def self.required_fields(klass)
       # the hashed password.
       def password=(new_password)
         @password = new_password
-        self.encrypted_password = password_digest(@password) 
+        self.encrypted_password = password_digest(@password) if @password.present?
       end
 
       # Verifies whether a password (ie from sign in) is the user password.
       def valid_password?(password)
+        return false if password.blank?
```
tegon: I guess this line doesn't need to be here, right? It was added in https://github.com/plataformatec/devise/pull/4261/files#diff-19b4dd928714c72f1338874351e8ff2dR45

mracos: Fixed by doing the reverts on chronological order

jbeker:
Thanks! Will there be a release shortly that includes this?

mracos:
@jbeker We already released 4.6.2 including these changes 😄

https://github.com/plataformatec/devise/releases/tag/v4.6.2
https://rubygems.org/gems/devise/versions/4.6.2

jbeker:
@mracos brilliant! thanks!

PR #5055, opened 2019-04-03
refactor method name to be more consistent

saiqulhaq:
fix https://github.com/plataformatec/devise/commit/1aab4499330c8a7641a3dea7e2ec3b96acdf46ae


lib/devise/failure_app.rb
```diff
@@ -278,14 +278,14 @@ def root_path_defined?(context)
       defined?(context.routes) && context.routes.url_helpers.respond_to?(:root_path)
     end
 
-    def rails_5_and_down?
-      return false if rails_5_up?
+    def rails_4_to_50?
```
feliperenan: Wdyt of `between_rails_4_and_5?`

saiqulhaq: I removed this method
because current Devise doesn't support rails 3
so I think no need to check is it rails 4 or 3

PR #5059, opened 2019-04-16
Setup Appraisal to help with maintenance

olbrich:
This PR is a work in progress.

I need Travis to run the build so I can debug / fix issues with the build.

Basic idea here is to setup Appraisal so that we can consistently check all supported ruby / rails versions.

A couple of changes:

1. removed Gemfile.lock because it should not be checked in for Gems
2. Added coverage for Rails 5.1 because it was missing

❓ Gemfile references jruby, but there is no indication of supported jruby versions.


rafaelfranca:
Thank you for the pull request but we don't need Appraisal to test with multiple versions of dependencies. We can use BUNDLE_GEMFILE and the gemfiles found here https://github.com/plataformatec/devise/tree/master/gemfiles.

Also Gemfile.lock should be checked in for Gems. See https://bundler.io/v2.0/guides/faq.html#committing-lockfiles-in-libraries.

Can you change your PR to just add coverage to Rails 5.1? Thank you.

olbrich:
@rafaelfranca Thanks for the link to the updated guidance.  I think there is a nuance here that needs to be considered.  The lock files for the various gemfiles should probably not be checked in because it obscures possible errors caused by updated dependencies.  If a new version of a dependency causes a failure, I would think that the maintainers would want to know that ASAP.  Travis has an option to run builds on a cron schedule, so leaving the lockfiles out of the gemfiles path and doing a weekly run should help identify broken dependencies faster.

If you want to include the Gemfile.lock in the repo then I suggest a couple of additional changes.  That lock file should point to the oldest supported version of rails instead of a new-ish one so that developers do not inadvertently break old versions by using new api methods.  The maintainers should also utilize the oldest supported ruby version (which looks to be 2.1 from the .travis.yml file), for the same reason.  

Do you have any guidance on the versions of jruby that are officially supported?  They don't seem to be documented anywhere.

rafaelfranca:
> The lock files for the various gemfiles should probably not be checked in because it obscures possible errors caused by updated dependencies. If a new version of a dependency causes a failure, I would think that the maintainers would want to know that ASAP. 

[The CI doesn't use the `Gemfile.lock` in the repository](https://github.com/plataformatec/devise/blob/master/.travis.yml#L79), only local development does. So what you expectations are already happening.

> If you want to include the Gemfile.lock in the repo then I suggest a couple of additional changes. That lock file should point to the oldest supported version of rails instead of a new-ish one so that developers do not inadvertently break old versions by using new api methods. The maintainers should also utilize the oldest supported ruby version (which looks to be 2.1 from the .travis.yml file), for the same reason.

Independent if Gemfile.lock is in the repository or not CI already test against all version we support so there is no way developers inadvertently break old versions using new APIs because CI will not pass.

> Do you have any guidance on the versions of jruby that are officially supported? They don't seem to be documented anywhere.

I don't think jruby is officially supported.

olbrich:
> [The CI doesn't use the `Gemfile.lock` in the repository](https://github.com/plataformatec/devise/blob/master/.travis.yml#L79), only local development does. So what you expectations are already happening.

Ah, perfect.  I missed that, thanks.

> Independent if Gemfile.lock is in the repository or not CI already test against all version we support so there is no way developers inadvertently break old versions using new APIs because CI will not pass.

While it's true that the build will fail if contributors use newer language features, they won't discover their mistake until they have kicked off a CI build.
One alternative approach here would be to symlink the `Gemfile` to the oldest supported rails version one (`gemfiles/Gemfile.rails-4.1-stable`) and set a `.ruby-version` file at `2.1.10`.

> I don't think jruby is officially supported.

Then it would make sense to remove all the jruby-specific stuff from the Gemfiles.
It would also make sense to have a clearly defined support policy for older ruby/rails versions.

I'll open some new PRs along those lines.

PR #5062, opened 2019-04-19
Cleanup gemfiles

olbrich:
* Simplify gemfiles for rails version support.
* add a gemfile for Rails 5.1

❓ Why are we skipping the builds for mongoid with Rails >= 5.0?


olbrich:
@rafaelfranca Any idea why the builds for mongoid with Rails >= 5.0 are being excluded in master?

mracos:
Thanks for your contribution @olbrich! 💙

I don't quite understand the motivation to "simplify the Gemfiles" and since there's a lot of conflicts I'm closing this PR. But I still think that adding rails 5.1 to CI is relevant so if you still want to send us a PR we will happily review it.

About:
> ❓ Why are we skipping the builds for mongoid with Rails >= 5.0?

I'm not sure about it since that was a decision made long ago but I would guess that it was because of some incompatibilities, maybe @tegon or @rafaelfranca can fill us in about it. 

If you want to try out enabling mongoid for rails > 5 and sending us a patch we would be happy to discuss it.

PR #5065, opened 2019-04-26
Fix rails 6.0.rc1 email uniqueness validation deprecation error

### LINKED ISSUE #5064: Deprecation warning in Rails 6.0.0.rc1 for email  uniqueness validation
## Environment

- Ruby 2.5.3
- Rails 6.0.0.rc1
- Devise 4.6.2

## Current behavior

```
DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the :email attribute in User model, pass `case_sensitive: true` option explicitly to the uniqueness validator.
```

## Expected behavior
No deprecation warning

## Possible solution

Since case_sensitive seems to be an option since rails 5 this could be added without breaking backwards compatibility: 
```ruby
# devise/models/validatable.rb
if Devise.activerecord51?
  validates_uniqueness_of :email, allow_blank: true, if: :will_save_change_to_email?, case_sensitive: true
[...]
```

Vasfed:
Rails 6 rc1 introduced a deprecation in uniqueness-validation:

> Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the case insensitive column, pass case_sensitive: true option explicitly to the uniqueness validator.

this PR fixes produced error.

Fixes #5064

### DIFF
--- a/lib/devise/models/validatable.rb
+++ b/lib/devise/models/validatable.rb
@@ -30,7 +30,7 @@ module Devise
         base.class_eval do
           validates_presence_of   :email, if: :email_required?
           if Devise.activerecord51?
-            validates_uniqueness_of :email, allow_blank: true, if: :will_save_change_to_email?
+            validates_uniqueness_of :email, allow_blank: true, case_sensitive: true, if: :will_save_change_to_email?
             validates_format_of     :email, with: email_regexp, allow_blank: true, if: :will_save_change_to_email?
           else
             validates_uniqueness_of :email, allow_blank: true, if: :email_changed?


mracos:
Thanks! 💙

rpbaptist:
This solves the deprecation warning, but raises a concern.

Maybe I'm understanding this wrong, but I think this setting will also have the effect that `bob@example.com` will be seen as a different address than `BOB@example.com` while email is not case sensitive and it should be treated as the same address.

Vasfed:
@rpbaptist you're correct, but that's what previous behaviour was, and that will be a breaking change (= need a major version release, someone might be relying on this or just have such data in their DB)

PR #5072, opened 2019-05-04
generate a confirmation_token over an empty string

### LINKED ISSUE #5071: Confirmable should generate a token if confirmation_token is currently an empty string
## Environment

- Ruby 2.5.5p157
- Rails 5.2.3
- Devise 4.6.1

## Current behavior

Currently, when using the confirmable module, devise only checks if the `confirmation_token` field not nil. If it's nil, it will generate a confirmation token for you; if it's not, it won't. However, it can be set to an empty string, which should not be allowed, as it presents a security issue, allowing someone passing an empty string as the confirmation token to sign in as the person with the blank string confirmation token. 

This is the offending code: https://github.com/plataformatec/devise/blob/e91b8ee0bac0503af119bf7333499d7729166a4d/lib/devise/models/confirmable.rb#L249

## Expected behavior

Devise should do something like `if self.confirmation_token.present?`, which checks for nil and blank.

ajsharp:
Fixes #5071

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -246,7 +246,7 @@ module Devise
         # Generates a new random token for confirmation, and stores
         # the time this token is being generated in confirmation_sent_at
         def generate_confirmation_token
-          if self.confirmation_token && !confirmation_period_expired?
+          if self.confirmation_token.present? && !confirmation_period_expired?
             @raw_confirmation_token = self.confirmation_token
           else
             self.confirmation_token = @raw_confirmation_token = Devise.friendly_token
--- a/test/models/confirmable_test.rb
+++ b/test/models/confirmable_test.rb
@@ -33,6 +33,12 @@ class ConfirmableTest < ActiveSupport::TestCase
     end
   end
 
+  test 'should generate a token over an empty string' do
+    user = create_user(confirmation_token: '')
+    assert_not_nil user.confirmation_token
+    assert_not_equal user.confirmation_token, ''
+  end
+
   test 'should confirm a user by updating confirmed at' do
     user = create_user
     assert_nil user.confirmed_at


ajsharp:
@mracos Any thoughts here? This should probably be considered a security issue, as it allows anyone to access the system with an empty token.

mracos:
Sorry @ajsharp, I've been really busy with work lately, so I didn't have much time to take a deeper look at this but will try to give you some feedback in the next days.
Thanks!

amangano-privy:
Would it be possible to add a validation to prevent the model from being saved when confirmation_token is an empty string?

tegon:
@ajsharp Thanks for your contribution!

I was digging into this, and a better approach for us is to ensure it's not possible to confirm a user with a blank `confirmation_token` parameter - i.e. to return a validation error inside the `ConfirmationsController#show` [action](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/confirmations_controller.rb#L22).
That should be enough to ensure no "invalid" records will be confirmed by mistake.

As for the `confirmation_token` being set to blank, that doesn't seem to be possible with the current code we have. I'm assuming something else could've happened in the system to let the records in that state. Unless we find a real case where Devise is setting the token to a blank string, I think it's not worth it to include more code to handle this.

@amangano-privy I thought about adding the validation, but we can't always require this field, so we would need a context validation. That would only protect the model against the known use cases. Another thing to take into account is that a bunch of places inside Devise uses `validate: false`.
So I think the validation wouldn't add that much value. The current code is already generating the token in a `before_create` callback, so I'm guessing it would require some external action to leave the field with a blank string - instead of `nil`.

WDYT you think? Thoughts?

tegon:
Closed in favor of https://github.com/plataformatec/devise/pull/5132

PR #5085, opened 2019-06-06
Add support for Rails 6.0.0.rc1

jtapia:


devise.gemspec
```diff
@@ -22,6 +22,6 @@ Gem::Specification.new do |s|
   s.add_dependency("warden", "~> 1.2.3")
   s.add_dependency("orm_adapter", "~> 0.1")
   s.add_dependency("bcrypt", "~> 3.0")
-  s.add_dependency("railties", ">= 4.1.0", "< 6.0")
```
rafaelfranca: `< 6.0` already support Rails 6.0.0.rc1. Why did you have to change this?

jasnow: I'm getting this when I try edge devise:
```
Bundler could not find compatible versions for gem "railties":
  In Gemfile:
    devise was resolved to 4.6.2, which depends on
      railties (>= 4.1.0, < 6.0)

    rails was resolved to 6.1.0.alpha, which depends on
      railties (= 6.1.0.alpha)
```

jtapia: I did that change because of an issue that I'm seeing using Rails 6.0.0.rc1

rafaelfranca: Well @jasnow, you are trying to using devise with 6.1.0.alpha, but the current code works with 6.0.0.rc1.

rafaelfranca: > I did that change because of an issue that I'm seeing using Rails 6.0.0.rc1

Can you post here the error you are getting?

jasnow: I see that I have expanded the scope of this issue (to Rails 6.1.x) - sorry - verified that Rails 6.0.0.rc1 works with devise for both of my "problem" repos. I will watch for Rails 6.1.x work on devise to comment further. Thanks for all of ya'll work.

Issue #5037, opened 2019-03-05
Support for Rails 6+ ?


MatthiasRMS:
Hi,

I'm trying to update my app to Rails 6 but I get this error:

```
devise was resolved to 4.6.1, which depends on
responders was resolved to 2.4.1, which depends on
actionpack (< 6.0, >= 4.2.0)
```

Is it planned to update it for Rails 6 ?
I'd be happy to contribute to the upgrade and submit a PR if you already know what there is to do to maintain the compatibility.

Matthias

georgedrummond:
I don't think this is a devise issue. I managed to get a `rails 5.2.2` and `devise 4.6.1` project to update just fine.

Try checking out a clean branch, pinning rails to `6.0.0.beta2` and running `bundle update`.

MatthiasRMS:
Yes 5.2.2 works fine. 
But since Devise relies on responders, which relies on actionpack < 6, and Rails 6 on actionpack 6+, there is a conflict.

feliperenan:
Hi @MatthiasRMS, thanks for the issue.

Yes, it is planned. See https://github.com/plataformatec/devise/pull/5009, but we don't have a date to release it yet. Probably when we have the final Rails 6 release.

It seems that responders already supports Rails 6, probably you are getting this error because it's not the final release of Rails. 

https://github.com/plataformatec/responders/blob/master/responders.gemspec#L21

chemic:
I was migrating my project from rails 4.X to 6-beta2 and that was my roadblock. Went currently with Rails 5.2.2 but hopefully could hop on the 6 train soon when Devise supports it.

saiqulhaq:
Devise .travis.yml setting for rails 6 is

```
source "https://rubygems.org"

gemspec path: ".."

gem "rails", '6.0.0.beta1'
gem "omniauth"
gem "omniauth-oauth2"
gem "rdoc"

gem "activemodel-serializers-xml", github: "rails/activemodel-serializers-xml"

gem "rails-controller-testing"

gem "responders", "~> 2.4"

group :test do
  gem "omniauth-facebook"
  gem "omniauth-openid"
  gem "timecop"
  gem "webrat", "0.7.3", require: false
  gem "mocha", "~> 1.1", require: false
  gem 'test_after_commit', require: false
end

platforms :ruby do
  gem "sqlite3", "~> 1.3.6"
end
```

so it should be OK with responders

but the test doesn't pass
[...]

PR #5086, opened 2019-06-12
Officially support Rails 6.0

rafaelfranca:
Also remove upper bound on railties so people can try devise with new versions without having to wait us to change the gem and report bugs.

Closes https://github.com/plataformatec/devise/issues/5037
Closes https://github.com/plataformatec/devise/pull/5085


rafaelfranca:
Yeah. There were in the repository with the "execution flag" in the filesystem (100755 → 100644 in the diff). I fixed it.

PR #5091, opened 2019-06-28
Fix sign_out_via default method(s) documentation

kennyadsl:
👋 

The default for `Devise.sign_out_via` configuration has been [changed from `:get` to `:delete`](https://github.com/plataformatec/devise/pull/4061) but the documentation has not been updated accordingly. 

Here's where the default is set in the codebase:

- https://github.com/plataformatec/devise/blob/81cb5b00f46cc24e6f41d6f1841ecaf442013898/lib/devise.rb#L226-L228
- https://github.com/plataformatec/devise/blob/81cb5b00f46cc24e6f41d6f1841ecaf442013898/lib/generators/templates/devise.rb#L256-L257

### DIFF
--- a/lib/devise/rails/routes.rb
+++ b/lib/devise/rails/routes.rb
@@ -135,7 +135,7 @@ module ActionDispatch::Routing
     #  * failure_app: a rack app which is invoked whenever there is a failure. Strings representing a given
     #    are also allowed as parameter.
     #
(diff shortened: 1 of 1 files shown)


tegon:
@kennyadsl Thank you for fixing this but it seems like it was already fixed in https://github.com/plataformatec/devise/pull/5099 so I'm going to close this pull request.

Issue #4767, opened 2018-01-28
Validatable should allow email uniqueness to be scoped


kylefox:
A common scenario is to scope email to a parent model (i.e. `Account`), but the Validatable module does not easily allow this.

Judging by the posts on StackOverflow, many people struggle with this issue. It would be a Good Thing if Devise could support this common functionality out of the box.

The [How to: Scope login to subdomain](https://github.com/plataformatec/devise/wiki/How-to:-Scope-login-to-subdomain) article recommends [completely removing `:validatable`](https://github.com/plataformatec/devise/wiki/How-to:-Scope-login-to-subdomain#check-that-you-do-not-have-validatable-in-the-devise-call-on-the-model) from the model. This results in _all_ email and password validations being removed — seems like throwing the baby out with the bathwater.

The wiki goes on to recommend that if you want to keep any of the validations, you should cherry-pick the ones you want — presumably by copy/pasting portions of [validatable.rb](https://github.com/plataformatec/devise/blob/master/lib/devise/models/validatable.rb). This approach seems error-prone, unmaintainable, and goes against the "flexible" philosophy in Devise's tagline 🙂

I propose validatable support a new option called `email_scope` that allows an `:email` uniqueness scope to be specified. For example:

```ruby
class User < ApplicationRecord
  belongs_to :account

  devise :database_authenticatable, :validatable, email_scope: :account
end
```

(The [caveats about indexes](https://github.com/plataformatec/devise/wiki/How-to:-Scope-login-to-subdomain#modify-migration) would still apply. I don't think it's necessary to change how Devise's migrations work — Rails already makes these schema changes trivial, after all).

I'm happy to help out with this feature if there's general agreement that this functionality should be included, and totally open to discussing other ways this scenario could be better addressed by Devise. Thanks!

argent-smith:
👍 … makes real sense in case of «soft-deleted» authenticatables as well.

one-more-alex:
+1 to having scope option.

While have no scope I am doing some terrible recreation of validator:
<pre>
# redefine uniqueness validator to confirm soft deletion
  self._validators[:email].reject!{|v| v.class == ActiveRecord::Validations::UniquenessValidator}
  self._validate_callbacks.delete self._validate_callbacks.find{|v| v.filter.class == ActiveRecord::Validations::UniquenessValidator && v.filter.attributes == [:email]}
  validates_uniqueness_of :email, allow_blank: true, if: :will_save_change_to_email?, scope: :deleted_at
[...]

PR #5094, opened 2019-07-09
add email_scope option for email uniqueness validator

flynfish:
This is a reboot of the closed PR #4793 with some tests. Issue #4767 has a lot of information about why this would be useful. 

Description from #4793 

- per subdomain auth;
- soft deletion and paranoid like support;
- example from https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-sign-in-using-their-username-or-email-address.

### DIFF
--- a/test/integration/registerable_test.rb
+++ b/test/integration/registerable_test.rb
@@ -124,7 +124,9 @@ class RegistrationTest < Devise::IntegrationTest
     #    https://github.com/mongoid/mongoid/issues/756
     (pending "Fails on Mongoid < 2.1"; break) if defined?(Mongoid) && Mongoid::VERSION.to_f < 2.1
 
-    create_user
+    user = create_user
+    user.update_attribute(:username, nil)
+
     get new_user_registration_path
 
     fill_in 'email', with: 'user@test.com'
@@ -254,7 +256,7 @@ class RegistrationTest < Devise::IntegrationTest
     assert_contain "Password confirmation doesn't match Password"
     refute User.to_adapter.find_first.valid_password?('pas123')
   end
-  
+
   test 'a signed in user should see a warning about minimum password length' do
     sign_in_as_user
     get edit_user_registration_path
--- a/test/models/validatable_test.rb
+++ b/test/models/validatable_test.rb
@@ -26,22 +26,13 @@ class ValidatableTest < ActiveSupport::TestCase
     assert user.valid?
   end
 
-  test 'should require uniqueness of email when email_scope is used' do
+  test 'should allow duplicate email when email_scope attribute does not match' do
     existing_user = create_user
 
-    user = new_user(email: '')
-    assert user.invalid?
-    assert_no_match(/taken/, user.errors[:email].join)
-
-    user.email = existing_user.email
+    user = new_user(email: existing_user.email)
     user.username = nil
     assert user.valid?
     assert_no_match(/taken/, user.errors[:email].join)
-
-    user.email = existing_user.email
-    user.username = existing_user.username
-    assert user.invalid?
-    assert_match(/taken/, user.errors[:email].join)
 
     user.save(validate: false)
     assert user.valid?
--- a/test/models/validatable_test.rb
+++ b/test/models/validatable_test.rb
@@ -26,6 +26,27 @@ class ValidatableTest < ActiveSupport::TestCase
     assert user.valid?
   end
 
+  test 'should require uniqueness of email when email_scope is used' do
+    existing_user = create_user
+
+    user = new_user(email: '')
+    assert user.invalid?
+    assert_no_match(/taken/, user.errors[:email].join)
+
+    user.email = existing_user.email
+    user.username = nil
+    assert user.valid?
+    assert_no_match(/taken/, user.errors[:email].join)
+
+    user.email = existing_user.email
+    user.username = existing_user.username
+    assert user.invalid?
+    assert_match(/taken/, user.errors[:email].join)
+
+    user.save(validate: false)
+    assert user.valid?
+  end
+
   test 'should require correct email format if email has changed, allowing blank' do
     user = new_user(email: '')
     assert user.invalid?
--- a/test/rails_app/lib/shared_user.rb
+++ b/test/rails_app/lib/shared_user.rb
@@ -7,7 +7,7 @@ module SharedUser
     devise :database_authenticatable, :confirmable, :lockable, :recoverable,
            :registerable, :rememberable, :timeoutable,
            :trackable, :validatable, :omniauthable, password_length: 7..72,
-           reconfirmable: false
+           reconfirmable: false, email_scope: [:username]
 
     attr_accessor :other_key
 
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -115,6 +115,10 @@ module Devise
   mattr_accessor :email_regexp
   @@email_regexp = /\A[^@\s]+@[^@\s]+\z/
 
+   # Option to scope the email uniqueness validator
+   mattr_accessor :email_scope
+   @@email_scope = nil
+
   # Range validation for password length
   mattr_accessor :password_length
   @@password_length = 6..128
--- a/lib/devise/models/validatable.rb
+++ b/lib/devise/models/validatable.rb
@@ -12,6 +12,7 @@ module Devise
     # Validatable adds the following options to devise_for:
     #
     #   * +email_regexp+: the regular expression used to validate e-mails;
+    #   * +email_scope+: option to scope the email uniqueness validator;
     #   * +password_length+: a range expressing password length. Defaults to 6..128.
     #
     module Validatable
@@ -30,10 +31,10 @@ module Devise
         base.class_eval do
           validates_presence_of   :email, if: :email_required?
           if Devise.activerecord51?
-            validates_uniqueness_of :email, allow_blank: true, case_sensitive: true, if: :will_save_change_to_email?
+            validates_uniqueness_of :email, allow_blank: true, case_sensitive: true, if: :will_save_change_to_email?, scope: email_scope
             validates_format_of     :email, with: email_regexp, allow_blank: true, if: :will_save_change_to_email?
           else
-            validates_uniqueness_of :email, allow_blank: true, if: :email_changed?
+            validates_uniqueness_of :email, allow_blank: true, if: :email_changed?, scope: email_scope
             validates_format_of     :email, with: email_regexp, allow_blank: true, if: :email_changed?
           end
 
@@ -66,7 +67,7 @@ module Devise
       end
 
       module ClassMethods
-        Devise::Models.config(self, :email_regexp, :password_length)
+        Devise::Models.config(self, :email_regexp, :password_length, :email_scope)
       end
     end
   end
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -173,6 +173,9 @@ Devise.setup do |config|
   # to give user feedback and not to assert the e-mail validity.
   config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
 
+  # Option to scope the email uniqueness validator. Default is nil;
+  # config.email_scope = [:username]
+
   # ==> Configuration for :timeoutable
   # The time you want to timeout the user session without activity. After this
   # time the user will be asked for credentials again. Default is 30 minutes.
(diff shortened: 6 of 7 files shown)


ACPK:
@flynfish @tegon - Do you know the status of this? I'd like to implement Device for a project that uses subdomains. At the moment, it uses the Apartment gem.

flynfish:
@ACPK I haven't heard anything else from @tegon yet. We are using my fork at the moment until this can hopefully be merged.

tegon:
Sorry guys, I wasn't able to look at this yet, but I'll definitely look at it as soon as I can.

christianrolle:
@tegon scoping the email makes totally sense in a multi tenant scenario.
We're gonna fork it with that important enhancement as well. Hope we're able to come back to trunk soon.

bananatron:
Scoping email validations is HUGE - thanks so much for this!

test/rails_app/lib/shared_user.rb
```diff
@@ -7,7 +7,7 @@ module SharedUser
     devise :database_authenticatable, :confirmable, :lockable, :recoverable,
            :registerable, :rememberable, :timeoutable,
            :trackable, :validatable, :omniauthable, password_length: 7..72,
-           reconfirmable: false
+           reconfirmable: false, email_scope: [:username]
```
tegon: I would rather create another test model with this option so that we can keep both behaviors being tested. Something similar to https://github.com/plataformatec/devise/blob/master/test/rails_app/lib/shared_user_without_email.rb

test/integration/registerable_test.rb
```diff
@@ -124,7 +124,9 @@ def user_sign_up
     #    https://github.com/mongoid/mongoid/issues/756
     (pending "Fails on Mongoid < 2.1"; break) if defined?(Mongoid) && Mongoid::VERSION.to_f < 2.1
 
-    create_user
+    user = create_user
+    user.update_attribute(:username, nil)
```
tegon: Would mind to explain why this change was needed?

test/models/validatable_test.rb
```diff
@@ -26,6 +26,18 @@ class ValidatableTest < ActiveSupport::TestCase
     assert user.valid?
   end
 
+  test 'should allow duplicate email when email_scope attribute does not match' do
```
tegon: I believe something like this should be enough:

```ruby
existing_user = create_user
user = new_user(email: existing_user.email, username: "New username")

assert user.valid?
```

test/integration/registerable_test.rb
flynfish: I don't remember now...I'll pick this back up in the next few days along with the other comments you made.

christianrolle:
@flynfish Did you have the chance to push this change forward? We certainly should get it done.

flynfish:
> @flynfish Did you have the chance to push this change forward? We certainly should get it done.

I didn't have a chance, but thanks for the reminder. I'm gonna set a goal to get the comments addressed in the next week.

sdonahuebk:
@flynfish we're looking to enable this in our project. Would you recommend forking for now?

Unnumbered:
any news?

xhocquet:
Hey @tegon , I'm stepping in to try and push this over the finish line for my colleague @flynfish 

I've split the test case out and reverted some unnecessary changes caused by conflating the two test users. Could you take another look and let me know what you think? Thanks! 🙏

xhocquet:
Hey @carlosantoniodasilva , I noticed you were actively merging PRs, could you help us get this one across the finish line? I think the community would appreciate this feature!

carlosantoniodasilva:
@xhocquet hey, yes, appreciate the ping. I've recently started to catch up on some of the issues/PRs here, just want to wrap up a few things on the other projects and then Devise is next. I'm putting this one high in the list to follow-up on first, please stay put for a little longer, I should get to it soon.

danzanzini:
Hello @carlosantoniodasilva, I was taking a look at this PR and I haven't found where these changes impact the migration generated to create users.

As far as I know, the migration generated by Devise adds a database index to ensure email uniqueness. Without changing this index, this feature will not work since the new user creation will be blocked at the database layer.

I'm not so sure about all of this, but can you take a look when reviewing this PR?

Thank you

xhocquet:
Thanks for the feedback @danzanzini , some thoughts - 

By following the referenced [wiki post](https://github.com/heartcombo/devise/wiki/How-To%3A-Allow-users-to-sign-in-with-something-other-than-their-email-address), the migration created by Devise generators would already be insufficient since they would not contain the 'username' for example. One option could be to add a commented out row with the scoped index in the migration to guide developers to the right usage?

Given the migration gotcha and the incompatibility with the recoverable module, maybe it makes sense to create a new wiki for scoped emails outlining the caveats?

danzanzini:
@xhocquet creating a wiki article is a good idea and should solve all these problems, I think it will also remove the need for the commented line in the migration since it would be noise for most users.

I'll work on a first version and then bring it here for discussion.

Thank you :)

danzanzini:
@xhocquet Here is the [first version of the wiki page](https://github.com/heartcombo/devise/wiki/How-To:-Scopes-email-validation), feel free to make any adjustments you find necessary.

Since this feature is not merged, I haven't added it to the wiki index.

xhocquet:
@danzanzini Small typo fix but otherwise the wiki looks great! I'm not sure if there's other considerations for this change, but the existing page seems to reflect the two you mentioned succinctly.

vickodin:
On the other hand, we have next 2 options in [config/initializers/devise.rb](https://github.com/heartcombo/devise/blob/master/lib/generators/templates/devise.rb#L41)

```ruby
  # ==> Configuration for any authentication mechanism
  # Configure which keys are used when authenticating a user. The default is
  # just :email. You can configure it to use [:username, :subdomain], so for
  # authenticating a user, both parameters are required. Remember that those
  # parameters are used only when authenticating and not when retrieving from
  # session. If you need permissions, you should implement that in a before filter.
  # You can also supply a hash where the value is a boolean determining whether
  # or not authentication should be aborted when the value is not present.
  # config.authentication_keys = [:email]

  # Configure parameters from the request object used for authentication. Each entry
  # given should be a request method and it will automatically be passed to the
  # find_for_authentication method and considered in your model lookup. For instance,
  # if you set :request_keys to [:subdomain], :subdomain will be used on authentication.
  # The same considerations mentioned for authentication_keys also apply to request_keys.
  # config.request_keys = []
```

I'm not yet understand why we need request_keys, but we can use:
config.authentication_keys = [:account_id, :email] instead of email_scope

What do you think?

xhocquet:
@vickodin I think your suggestion is not directly related to the problem this PR is trying to solve. The problem we are trying to solve is that multiple users with the same email cannot be created because of Devise validations. 

The configs you point to could be used to lookup these users after they are created. Personally we would not use this and would instead use domain-based scoping to load our user records.

vickodin:
> @vickodin I think your suggestion is not directly related to the problem this PR is trying to solve. The problem we are trying to solve is that multiple users with the same email cannot be created because of Devise validations.
> 
> The configs you point to could be used to lookup these users after they are created. Personally we would not use this and would instead use domain-based scoping to load our user records.

@xhocquet  Thank you for explanations!

Then when this PR will be merged?

Thanks in advance!

vickodin (approved):
Looks good!

xhocquet:
Hey @carlosantoniodasilva, happy new year! Hoping to bump this in your notifications 😄

carlosantoniodasilva:
@xhocquet happy new year! Consider it bumped 😉

There may be more to it than just the validation so I need to look into it a bit more carefully in the coming weeks.

xxz3:
+1, we need this please :)

poetry-like-code:
Although this request seems to be dragging for morethan 3 years now, i don't see anyone venting out any frustration, kudos to everyone for that. :)
And +1, we need this please :)

stephenpassero:
@carlosantoniodasilva I noticed @poetry-like-code brought up this PR recently, but didn't see any changes. Is there any update on getting this merged?

Thanks!

carlosantoniodasilva:
@stephenpassero appreciate the ping, unfortunately no updates at this time yet.

This is still where it stands:
> There may be more to it than just the validation so I need to look into it a bit more carefully in the coming weeks.

Unfortunately 3+ months and I haven't stopped to look into this carefully yet, but I'm bumping it again in my list here. First I have to get a release out there. 😅

akshaysmurthy:
Until this is merged, what is the recommendation for people like me who wants to implement this in our app today? 😄

danzanzini:
Hi @akshaysmurthy, I did these steps to implement it in my app:

1) Changed the unique index of the email column:
``` 
def up
  remove_index :users, :email
  add_index :users, [:email, :tenant_id], unique: true
end
```

2) Overwritted the `will_save_change_to_email?` method so devise's skips its built-in validations:
```
# user.rb
def will_save_change_to_email?
  false
end
```

3) Created my own email and password validations (Password validation is also skipped because of step 2):
```
# user.rb

validates :email, uniqueness: { scope: :tenant_id }
validates :email, format: /\A[^@\s]+@[^@\s]+\z/
```

Hope it helps

mrabetr:
when is this due to be merged please? thanks

flynfish:
@tegon @vickodin @carlosantoniodasilva it would be really helpful to the community if we could get this merged or if you could provide any insight into what we can do to help out further

carlosantoniodasilva:
@danzanzini thanks for the detailed steps there.

@akshaysmurthy the simplest thing you can do if you need any further customization: don't use validatable, write your own email/password validations that include the uniqueness scope.

@mrabetr there's no due date for merging PRs. ;)

@flynfish at this point I cannot provide any insights, without actually stopping to look into the problem and think about it more carefully, my apologies. Like I said above, multi tenancy might be more than just the validation so I want to consider the problem better, it's just not been high in my priority list, all things considered. In any case, validatable is composed of just a few validations that you can easily add to your model to customize things further for now.

I am unfortunately not in a position to provide any timeline for when I can work on this. Again, I apologize to everyone for not having given this proper attention yet, but I want to reassure that I'm keeping an eye on each and every comment.

❤️💜💛💚💙

bvogel:
Yeah, another year and a tad have passed, making this probably the longest outstanding MR in this project with the most interest. Any update? (i.e. bump!)

flamontagne:
> Hi @akshaysmurthy, I did these steps to implement it in my app:
> 
> 1. Changed the unique index of the email column:
> 
> ```
> def up
>   remove_index :users, :email
>   add_index :users, [:email, :tenant_id], unique: true
> end
> ```
> 
> 2. Overwrote the `will_save_change_to_email?` method so Devise skips its built-in validations. This change may impact other parts of your application if this method is used anywhere else.
> 
> ```
> # user.rb
> def will_save_change_to_email?
>   false
> end
> ```
> 
> 3. Created my own email validations:
> 
> ```
> # user.rb
> 
> validates :email, uniqueness: { scope: :tenant_id }
> validates :email, format: /\A[^@\s]+@[^@\s]+\z/
> ```
> 
> Hope it helps

This is a great workaround, however redifining will_save_change_to_email? [will also mess with the Confirmable plugin](https://github.com/heartcombo/devise/blob/ec0674523e7909579a5a008f16fb9fe0c3a71712/lib/devise/models/confirmable.rb#L272).

Instead of redefining will_save_change_to_email? I think it's better to disable the validatable plugin. Just make sure you add all the validators you need (for email and password columns) in the User model.

bvogel:
@carlosantoniodasilva if you aren't willing to implement this, please be honest enough to at least give an explanation and close this PR.

flamontagne:
A word of caution: I just found out that the workaround mentionned in this thread will also mess badly with the password reset function. If the same email address is associated with two or more tenants, the password reset link in the email may point to the WRONG tenant. (e.g. wrongsubdomain.myapp.com/reset-password?token=blablabla...)

bvogel:
As this just hit us _again_, @carlosantoniodasilva - any news and/or decisions on how to move forward with this?

justwiebe:
@flamontagne In my case, I just needed to support soft-deleted users, so I was able to fix the password reset link by hacking `send_reset_password_instructions`
```ruby
class User < ApplicationRecord
  def self.send_reset_password_instructions(attributes = {})
    recoverable = where(deleted_at: nil).find_or_initialize_with_errors(reset_password_keys, attributes, :not_found)
    recoverable.send_reset_password_instructions if recoverable.persisted?
    recoverable
  end
end
```

fa11enangel:
@justwiebe thank you for the hint. I've updated the wiki subdomain tutorial to add a password reset instruction:

see: https://github.com/heartcombo/devise/wiki/How-to:-Scope-login-to-subdomain

@flamontagne you can have a look at it. It should help you to find a solution.

PR #5098, opened 2019-07-11
Option to use password_hash instead of encrypted_password

ankane:
Hi, thanks for this awesome gem.

This PR adds an option to use the `password_hash` field instead of `encrypted_password` with the intent to make it the default going forward (since Devise doesn't encrypt passwords).

```ruby
Devise.setup do |config|
  # Use password_hash field instead of encrypted_password
  config.use_password_hash_field = true
end
```

A few notes:

- I wanted to use the name `password_digest` to match Rails `has_secure_password`, but it looks like `password_digest(password)` is the public interface for custom hashing. Could also go with `hashed_password`.
- I tried using `alias_attribute` so `encrypted_password` would continue to work even when enabled (for cleaner code/better backwards compatibility), but it didn't play nicely with Mongoid (Mongoid requires attributes to be defined before they can be aliased). We could still include this for Active Record.
- I didn't want to blow up the test suite, so just added two tests to the build matrix.
- 3rd party libraries would want to use `Devise.password_field` going forward.

Let me know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,13 @@ gemfile:
   - gemfiles/Gemfile.rails-4.1-stable
 
 matrix:
+  include:
+    - rvm: 2.6.0
+      gemfile: Gemfile
+      env: DEVISE_ORM=active_record DEVISE_PASSWORD_HASH=t
+    - rvm: 2.5.3
+      gemfile: gemfiles/Gemfile.rails-4.2-stable
+      env: DEVISE_ORM=mongoid DEVISE_PASSWORD_HASH=t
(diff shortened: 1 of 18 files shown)


ankane:
To give more context: Changing a field name should not be done lightly. Looking at this long-term, I hope Devise is still a top library 10-20 years from now and new developers aren't confused why the password field is named `encrypted_password` when they've heard you should never encrypt passwords. I think it makes sense to deal with the pain of this transition sooner rather than later (as more docs / tutorials / etc are written each week).

PR #5107, opened 2019-08-02
Allow remember_for option to be dynamic

nneufeld:
This will allow the `remember_for` option to be dynamic, by defining a `remember_for` method on the resource model in the same way as the `timeout_in` option supports.
https://github.com/plataformatec/devise/wiki/How-To:-Add-timeout_in-value-dynamically

### DIFF
--- a/lib/devise/models/rememberable.rb
+++ b/lib/devise/models/rememberable.rb
@@ -63,7 +63,7 @@ module Devise
       end
 
       def remember_expires_at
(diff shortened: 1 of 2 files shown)


mracos:
Hi, @nneufeld thanks for your contribution.

Can you explain better what's the use case for this feature? This is going to help us understand what this patch is solving and make a proper code review. 😄

PR #5109, opened 2019-08-06
Fix specs on Rails 6 RC2

tegon:
The `ActiveRecord::MigrationContext` class now has a `schema_migration` attribute.
Reference: https://github.com/rails/rails/pull/36439/files#diff-8d3c44120f7b67ff79e2fbe6a40d0ad6R1018

This pull request also includes some changes to avoid deprecation warnings: https://github.com/plataformatec/devise/commit/3273f9e08943deb1b413268392bb640512aee799, https://github.com/plataformatec/devise/commit/77cb3941ec3c198435309370c9e538ab1339a378 and https://github.com/plataformatec/devise/commit/69534c666c66a30c4e1c814c0c15846059f35dcc

### DIFF
--- a/test/rails_app/config/application.rb
+++ b/test/rails_app/config/application.rb
@@ -45,7 +45,7 @@ module RailsApp
       Devise::SessionsController.layout "application"
     end
 
-    # Remove this check once Rails 5.0 support is removed.
+    # Remove the first check once Rails 5.0 support is removed.
     if Devise::Test.rails52_and_up? && !Devise::Test.rails6?
       Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true
     end
--- a/test/rails_app/config/application.rb
+++ b/test/rails_app/config/application.rb
@@ -46,7 +46,7 @@ module RailsApp
     end
 
     # Remove this check once Rails 5.0 support is removed.
-    if Devise::Test.rails52_and_up?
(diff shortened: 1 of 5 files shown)


test/rails_app/config/application.rb
```diff
@@ -46,7 +46,7 @@ class Application < Rails::Application
     end
 
     # Remove this check once Rails 5.0 support is removed.
-    if Devise::Test.rails52_and_up?
+    if Devise::Test.rails52_and_up? && !Devise::Test.rails6?
```
Lucasnar: Probably something like `# Remove the first check once Rails 5.0 support is removed.` now? Do you know if Rails 6 is always going to require this? (It might make sense to update the docs to consider it too.)

test/rails_app/config/boot.rb
```diff
@@ -8,6 +8,10 @@ module Devise
   module Test
     # Detection for minor differences between Rails 4 and 5, 5.1, and 5.2 in tests.
     
+    def self.rails6?
```
Lucasnar: Should we update this module's docs as well?

test/rails_app/config/application.rb
tegon: We'll still need the `rails6?` check once we remove support for Rails 5 because here we need to set this config for Rails 5.2. When Rails 6 is the only supported version, then we can remove this code altogether.

tegon:
@Lucasnar Thanks!

PR #5112, opened 2019-08-10
Refactored `SecretKeyFinder#find` method

Gokul595:
Refactored `SecretKeyFinder#find` method and reduced Cyclomatic Complexity.


mracos:
Thanks for your contribution @Gokul595. 💙

Nice work on reducing cyclomatic complexity!

But we actually think that the current code is more straightforward in its current form since it's a method that **conditionally has to do a procedural check** for where the `:secret_key_base` is stored, so just by reading the code we can grasp the method objective with the sequential `ifs` and `elsifs`.
Hiding this explicit conditional with metaprogramming and the likes can actually increase indirection and make the method objective less clear on first glance.

And that's why I'm closing this PR.

But I agree that it could be improved so if you have other ideas on how to make this method better we will happily review. 😄

PR #5132, opened 2019-09-02
Always return an error when `confirmation_token` is blank

### LINKED ISSUE #5071: Confirmable should generate a token if confirmation_token is currently an empty string

tegon:
As reported in #5071, if for some reason, a user in the database had the `confirmation_token`
column as a blank string, Devise would confirm that user after receiving a request with a blank `confirmation_token` parameter.

After this commit, a request sending a blank `confirmation_token` parameter will always receive a validation error, even if there are users with a blank `confirmation_token` in the database.

For applications that have users with a blank `confirmation_token` in the database, it's recommended to manually regenerate or to nullify them.

Closes #5071

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -349,6 +349,12 @@ module Devise
         # Options must have the confirmation_token
         def confirm_by_token(confirmation_token)
           confirmable = find_first_by_auth_conditions(confirmation_token: confirmation_token)
+
+          if confirmable && confirmation_token.blank?
+            confirmable.errors.add(:confirmation_token, :blank)
+            return confirmable
+          end
+
           unless confirmable
             confirmation_digest = Devise.token_generator.digest(self, :confirmation_token, confirmation_token)
             confirmable = find_or_initialize_with_error_by(:confirmation_token, confirmation_digest)
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -348,17 +348,18 @@ module Devise
         # If the user is already confirmed, create an error for the user
         # Options must have the confirmation_token
         def confirm_by_token(confirmation_token)
-          confirmable = find_first_by_auth_conditions(confirmation_token: confirmation_token)
-
           # When the `confirmation_token` parameter is blank, if there are any users with a blank
           # `confirmation_token` in the database, the first one would be confirmed here.
           # The error is being manually added here to ensure no users are confirmed by mistake.
           # This was done in the model for convenience, since validation errors are automatically
           # displayed in the view.
-          if confirmable && confirmation_token.blank?
+          if confirmation_token.blank?
+            confirmable = new
             confirmable.errors.add(:confirmation_token, :blank)
             return confirmable
           end
+
+          confirmable = find_first_by_auth_conditions(confirmation_token: confirmation_token)
 
           unless confirmable
             confirmation_digest = Devise.token_generator.digest(self, :confirmation_token, confirmation_token)
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -350,6 +350,11 @@ module Devise
         def confirm_by_token(confirmation_token)
           confirmable = find_first_by_auth_conditions(confirmation_token: confirmation_token)
 
+          # When the `confirmation_token` parameter is blank, if there are any users with a blank
+          # `confirmation_token` in the database, the first one would be confirmed here.
+          # The error is being manually added here to ensure no users are confirmed by mistake.
+          # This was done in the model for convenience, since validation errors are automatically
+          # displayed in the view.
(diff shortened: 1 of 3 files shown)


lib/devise/models/confirmable.rb
```diff
@@ -349,6 +349,17 @@ def send_confirmation_instructions(attributes={})
         # Options must have the confirmation_token
         def confirm_by_token(confirmation_token)
           confirmable = find_first_by_auth_conditions(confirmation_token: confirmation_token)
+
+          # When the `confirmation_token` parameter is blank, if there are any users with a blank
+          # `confirmation_token` in the database, the first one would be confirmed here.
+          # The error is being manually added here to ensure no users are confirmed by mistake.
+          # This was done in the model for convenience, since validation errors are automatically
+          # displayed in the view.
+          if confirmable && confirmation_token.blank?
```
amangano-privy: As written, this code may leak information about the internal state of the database (that some records have blank tokens). Could we check for presence of confirmation_token before the call to `find_first_by_auth_conditions`?

tegon: > this code may leak information about the internal state of the database (that some records have blank tokens)

I'm not sure about that. It's the same response as when there are no records with blank tokens. Do you think this is still leaking information?

amangano-privy: If that's the case then I guess it's probably ok for now, but if the `find_or_initialize_with_error_by` or `token_generator` methods change, it could unintentionally create an information leakage issue here. I think it'd be safer to just return early (without making any queries) if we detect that the token is blank.

```ruby
if confirmation_token.blank?
  confirmable = new
  confirmable.errors.add(:confirmation_token, :blank)
  return confirmable
end
```

tegon: > but if the find_or_initialize_with_error_by or token_generator methods change, it could unintentionally create an information leakage issue here

Do you have examples of which kind of change could cause leaks?

amangano-privy: > > but if the find_or_initialize_with_error_by or token_generator methods change, it could unintentionally create an information leakage issue here
> 
> Do you have examples of which kind of changes could cause leaks?

Any change that would result in a different response being returned when the token is blank and there is no record in the database.

Examples:
- `TokenGenerator#digest` changes to return a digest even when `value` is blank. This would result in a record being initialized and returned with error set to `:invalid` instead of `:blank`.
- `find_or_initialize_with_error_by` changes to always use the provided `error` argument instead of special-casing blank errors.

tegon: Ok, get it. Both of these cases would be caught by our integration tests, so I guess this implementation is fine.
But since we already know there's no need to perform the query I guess it's better to return early as you mentioned. It also shows better the intention behind the code, which is to validate whether the parameter is present or not.

tegon: I've added the changes, thanks for the review!

test/models/confirmable_test.rb
```diff
@@ -77,6 +77,24 @@ def setup
     assert_equal "can't be blank", confirmed_user.errors[:confirmation_token].join
   end
 
+  test 'should return a new record with errors when a blank token is given and a record exists on the database' do
```
amangano-privy: I think this no longer needs 'and a record exists on the database'

test/models/confirmable_test.rb
```diff
@@ -77,6 +77,24 @@ def setup
...
+    user = create_user(confirmation_token: '')
+
+    confirmed_user = User.confirm_by_token('')
+
+    refute user.reload.confirmed?
+    assert_equal "can't be blank", confirmed_user.errors[:confirmation_token].join
+  end
+
+  test 'should return a new record with errors when a nil token is given and a record exists on the database' do
```
amangano-privy: I think this no longer needs 'and a record exists on the database'

test/models/confirmable_test.rb
tegon: I think it's better to keep since that's the behavior we're trying to ensure it doesn't happen anymore.
I know that since the current code validates the parameter before doing anything else, it's unlikely for this to happen, but this test ensures we don't add regressions in the future.

PR #5136, opened 2019-09-13
Add email validation scoping

christianrolle:
For multi tenancy the email validation scoping is
crucial. This enhancement is actually based on the
currently open PR:
https://github.com/plataformatec/devise/pull/5094
As soon as the feature is merged, one should use the
Devise trunk again.


danzanzini:
Isn't this a duplication of #5049?

PR #5142, opened 2019-09-25
Explain layout of default config initializer

### LINKED ISSUE #4555: [DOCUMENTATION] In the initializer, why are some settings not commented out?
I just started playing with Devise for the first time today. In going through `config/initializers/devise.rb`, I noticed that most configuration settings are the same as the defaults provided in Devise's `lib/devise.rb`.<sup>[1]</sup> Many of the settings are also commented out, but there appears to be no rhyme or reason as to which ones.

As a new user, I found this very confusing: I had to dig into Devise's source code to confirm that these default values were, in fact, mirrored in the `module Devise` definition, and that it would be safe to comment them out. At first glance, my assumption was that any uncommented setting _had_ to be manually configured, or else the gem would not work. 

Is there a reason for this arrangement that I've somehow missed? Perhaps this was intended as a convenience for experienced users, to remind them of which settings might be more likely to require attention? In any case, that was not my initial interpretation, and getting to the bottom of this question created an unnecessary obstacle in getting started. 

In the interest of facilitating adoption for new users, I'd like to suggest that, where reasonable, all default settings in the Devise initializer be commented out from the start. Thoughts?

---

**Note 1:** Some exceptions to this rule (like `@@secret_key` or `@@mailer_sender`) make sense — they should be unique for each application, so they default to `nil`. Others (like `@@paranoid` or `@@allow_unconfirmed_access_for`), are quite confusing — what reason is there to break the pattern established through the rest of the file?

Combined with the inconsistencies described above, these small and unpredictable exceptions make it difficult for a beginner to make reasoned inferences about how the configuration process is designed.

rlue:
Fixes #4555.

### DIFF
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
+# Assuming you have not yet modified this file, each configuration option below
+# is set to its default value. Note that some are commented out while others
+# are not: uncommented lines are intended to protect your configuration from
+# breaking changes in upgrades (i.e., in the event that future versions of
+# Devise change the default values for those options).
+#
 # Use this hook to configure devise mailer, warden hooks and so forth.
 # Many of these configuration options can be set straight in your model.
 Devise.setup do |config|
--- a/test/rails_app/config/initializers/devise.rb
+++ b/test/rails_app/config/initializers/devise.rb
@@ -3,6 +3,12 @@
 require "omniauth-facebook"
 require "omniauth-openid"
 
+# Assuming you have not yet modified this file, each configuration option below
+# is set to its default value. Note that some are commented out while others
+# are not: uncommented lines are intended to protect your configuration from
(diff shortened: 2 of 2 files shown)


rlue:
Questions for @tegon or @rafaelfranca:

1. Did I update the correct initializer file in this PR? (`lib/generators/templates/devise.rb`)
2. Should I also update the `test/rails_app/config/initializers/devise.rb` test data file?
3. Are you satisfied with the language of this PR, and the placement of the comment in the file?

tegon:
The changes look good to me. I'll wait for @rafaelfranca input before we merge it.

I don't have strong opinions about updating the `test/rails_app/config/initializers/devise.rb` file or not but I guess it might be good to keep things in sync.

There are some weird failures in Travis, which are probably not related to these changes since there are no code changes here. You can append `[ci skip]` to your commit in order to not run the CI.
I'll try to take a loot at them later.

Thanks!

rlue:
Done, thanks.

PR #5092, opened 2019-07-06
Set stretches to the minimum cost supported by the bcrypt algorithm in tests

sergey-alekseev:
During generating and setting `encrypted_password` by default `BCrypt::Password.create` is [used](https://github.com/plataformatec/devise/blob/f39c6fd92774cb66f96f546d8d5e8281542b4e78/lib/devise/encryptor.rb#L11). At the moment in the test environment the cost is set to 1. It is passed to `BCrypt::Password.create` and later set to 4 (`BCrypt::Engine::MIN_COST`) ([1](https://github.com/codahale/bcrypt-ruby/blob/05d81878a443b12ea39927ac0d84d8ff565e24bc/lib/bcrypt/password.rb#L46) -> [2](https://github.com/codahale/bcrypt-ruby/blob/05d81878a443b12ea39927ac0d84d8ff565e24bc/lib/bcrypt/engine.rb#L65-L66)). Observing `encrypted_password` being generated with the cost 4 while the cost 1 is specified via stretches may be confusing.

PR #5145, opened 2019-09-26
Regenerate password hash on a sign in after the stretches config changes

### LINKED ISSUE #5143: Regenerate password hash on a sign in after the stretches config changes
I want to remind about a feature which was requested in https://github.com/plataformatec/devise/pull/1862, https://github.com/plataformatec/devise/issues/3635 and https://github.com/plataformatec/devise/pull/3549#issuecomment-193235701. 

https://github.com/rails/rails/pull/35321/commits/87aa60ec367212e5ca49f39a9de29088f22ff469 handles this:  

* Rehash passwords with newer cost when they're authenticated.

  On successful authentication, rehash passwords that had originally
  been hashed with an outdated, lower cost. Ensures old passwords don't
  grow increasingly vulnerable to offline cracking.

For me it totally makes sense to **regenerate password hash on a sign in after the stretches config changes**. I can go ahead and submit a PR. Using a devise gem version would be better than using a devise patched version. What do you think?

_(copied from https://github.com/plataformatec/devise/pull/5074#issuecomment-492696360 as @mracos [suggested](https://github.com/plataformatec/devise/pull/5074#issuecomment-532298342))_

sergey-alekseev:
Closes #5143.  
Depends on #5092.


lib/devise/strategies/database_authenticatable.rb
```diff
@@ -13,6 +13,14 @@ def authenticate!
         if validate(resource){ hashed = true; resource.valid_password?(password) }
           remember_me(resource)
           resource.after_database_authentication
+
+          if BCrypt::Password.new(resource.encrypted_password).cost != resource.class.stretches
+            resource.update_column(
```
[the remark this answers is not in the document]
sergey-alekseev: I did it intentionally.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ### Unreleased
 * enhancements
   * Increase default stretches to 12 (by @sergey-alekseev)
+  * Set stretches to the minimum cost supported by the bcrypt algorithm in tests (by @sergey-alekseev)
+  * Regenerate password hash on a sign in after the stretches config changes (by @sergey-alekseev)
```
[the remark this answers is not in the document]
sergey-alekseev: It's consistent with the rest of the file.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ### Unreleased
 * enhancements
   * Increase default stretches to 12 (by @sergey-alekseev)
+  * Set stretches to the minimum cost supported by the bcrypt algorithm in tests (by @sergey-alekseev)
```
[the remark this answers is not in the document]
sergey-alekseev: It's consistent with the rest of the file.

lib/devise/strategies/database_authenticatable.rb
```diff
@@ -24,6 +25,17 @@ def authenticate!
           Devise.paranoid ? fail(:invalid) : fail(:not_found_in_database)
         end
       end
+
+      private
+
+      def regenerate_bcrypt_password_hash(resource, password)
+        if !Devise.respond_to?(:encryptor) &&
+          ::BCrypt::Password.new(resource.encrypted_password).cost != resource.class.stretches
```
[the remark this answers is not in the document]
sergey-alekseev: Will change during the next iteration when I get human feedback regarding the PR's code.

lib/devise/strategies/database_authenticatable.rb
```diff
@@ -24,6 +25,17 @@ def authenticate!
           Devise.paranoid ? fail(:invalid) : fail(:not_found_in_database)
         end
       end
+
+      private
+
+      def regenerate_bcrypt_password_hash(resource, password)
+        if !Devise.respond_to?(:encryptor) &&
```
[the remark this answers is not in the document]
sergey-alekseev: Looking forward for your recommendations how to better detect whether bcrypt or other hashing algorithm is used.

sergey-alekseev:
Some Travis CI jobs failed because of this bug – https://github.com/seattlerb/minitest/issues/803. The bug was introduced recently in the latest version. The fix is already in the minitest's master, just needs a gem version to be released.

I was able to reproduce it locally after installing Ruby 2.3 and doing `bundle update` (Travis [does this](https://github.com/plataformatec/devise/blob/f48b6f1651233a39c52dd70e9ac6c35f70086062/.travis.yml#L79) before running `bin/test`).

lib/devise/strategies/database_authenticatable.rb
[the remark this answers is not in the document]
sergey-alekseev: Done.

sergey-alekseev:
One Ruby 2.2 Travis CI test [failed](https://travis-ci.org/plataformatec/devise/jobs/590776409) because they released a gem version with both the match fix and Ruby 2.3+ requirement. So Ruby 2.2 is stuck with a buggy 5.12.0, not sure how to better fix it without additional Minitest gem updates. [Started](https://github.com/seattlerb/minitest/issues/803#issuecomment-536182589) a discussion in that issue.

sergey-alekseev:
✅ Travis CI is green now because a Minitest's maintainer released a new gem version.

lib/devise/strategies/database_authenticatable.rb
```diff
@@ -24,6 +25,17 @@ def authenticate!
...
+            ::BCrypt::Password.new(resource.encrypted_password).cost != resource.class.stretches
+          resource.password = password
+          resource.skip_password_change_notification!
+          resource.save
```
mainameiz: Please pay attention here.
`update_column` should be used here, but sourcelevel-bot consider this as a mistake, but it isn't - https://github.com/plataformatec/devise/pull/5145#discussion_r328703342.

sergey-alekseev: @mainameiz what do you mean by "Please pay attention here."?

Thanks for referencing my own comment regarding the usage of `update_column`. I used it initially and then during a refactoring iteration found the usage of `resource.skip_password_change_notification!` in the repo which (as far as I remember) in conjunction with `resource.save` seemed like a more proper implementation. Again, as far as I remember that's the only callback we need to avoid during the update. I haven't looked into the related repo code again now, but what I just wrote should be trustworthy :) I'd appreciate if you can review and verify.

mainameiz: I'm not really familiar with this codebase but I suppose that `resource` variable is a instance of `User` model of some application. So it's state in database may be not valid. And if it's not valid then `resource.save` will not update the password.
I suppose, Devise should regenerate the digest nevertheless, so `update_column` should be used here.

sergey-alekseev: Thanks for the feedback. Indeed, `resource.save` may fail if a `resource` is not valid.

I recalled why `#update_column` can't be used. See [the Travis CI build](https://travis-ci.org/plataformatec/devise/builds/590032290) for [the initial commit](https://github.com/plataformatec/devise/commit/496dfe9fe666707b190f4de42e1da19bc0d346ee) where `#update_column` was used. All jobs with `DEVISE_ORM=mongoid`  failed because it's an Active Record method.

msxavi: What about `resource.save(validate: false)`?

msxavi:
How are we tracking with this?

lib/devise/strategies/database_authenticatable.rb
sergey-alekseev: From https://guides.rubyonrails.org/v6.0/active_record_callbacks.html#running-callbacks:  
>The following methods **trigger** callbacks:
...
save(validate: false)
...

On the contrary we want to avoid callbacks like sending the password change notification.

sergey-alekseev:
> How are we tracking with this?

Is it a question to the devise owners? Can you please clarify your question? @msxavi

lib/devise/strategies/database_authenticatable.rb
msxavi: @sergey-alekseev 
> On the contrary we want to avoid callbacks like sending the password change notification

Invoking `skip_password_change_notification!` as the name tells, skips the callback you mentioned.

Your original code `resource.save` also triggers callbacks, so I didn't get your point.

msxavi:
> > How are we tracking with this?
> 
> Is it a question to the devise owners? Can you please clarify your question? @msxavi

@sergey-alekseev since you are responsible for this PR, the question was meant for you.

lib/devise/strategies/database_authenticatable.rb
msxavi: Maybe this could solve it:
```suggestion
          update_column(:encrypted_password, BCrypt::Password.create(password, cost: resource.class.stretches).to_s)
```

sergey-alekseev: Ah, I thought you want to use  
```ruby
resource.save(validate: false)
```  
instead of  
```ruby
resource.skip_password_change_notification!
resource.save
```
because of the previous context in this thread.  

Now I see your suggestion was to replace `resource.save` with `resource.save(validate: false)`. This totally makes sense. `resource.save(validate: false)` is extensively used in the codebase (e.g. look into https://github.com/plataformatec/devise/pull/4194). I made the change in https://github.com/plataformatec/devise/pull/5145/commits/31998a959d339b5352ad42861df2b2dbc1aef8f7#diff-d1ac35f7ea071d6c944aeb7ee4cd06cfR36 and covered this change in tests in https://github.com/plataformatec/devise/pull/5145/commits/31998a959d339b5352ad42861df2b2dbc1aef8f7#diff-0cd3b7071e3422d894ff8564d7f8edc9R128. Thanks for the suggestion! @msxavi

sergey-alekseev: Not sure about your suggestion to use `#update_column` as [just above](https://github.com/plataformatec/devise/pull/5145#discussion_r352292285) I explained why it can't be used. @msxavi

sergey-alekseev:
>How are we tracking with this?

To be clear – I don't understand this question. @msxavi

lib/devise/strategies/database_authenticatable.rb
msxavi: Oh sure, no problem. I ended up using `update_column` because `save(validate: false)` didn't work quite well for me due to database constraints. Just to be clear, I'm trying to solve the stretches issue in the product I work on.

That's my final code which seems to work:
```
    return unless self.encrypted_password.present?
    if BCrypt::Password.new(self.encrypted_password).cost != self.class.stretches && persisted?
      update_column(:encrypted_password, BCrypt::Password.create(password, cost: self.class.stretches).to_s)
    end
```

sergey-alekseev: Curious why `save(validate: false)` didn't work in your case.  

I use similar code with `#update_column` in a monkey-patched `Devise::Strategies::DatabaseAuthenticatable`, but this PR is to replace this monkey-patched solution.

aried3r:
Hey! We're interested in this feature as well, since we are currently using monkey-patches which we need to review again every time a new devise version is released and having this as part of the official code would be great!

Is there any way we can help this become a reality?

lib/generators/templates/devise.rb
```diff
@@ -103,15 +103,22 @@
...
-  # algorithm), the cost increases exponentially with the number of stretches (e.g.
-  # a value of 20 is already extremely slow: approx. 60 seconds for 1 calculation).
-  config.stretches = Rails.env.test? ? 1 : 12
+  # For bcrypt (the default algorithm), this is the cost for hashing the password.
+  # If using other algorithms, it sets how many times you want the password to be hashed.
+  config.stretches = if Rails.env.test?
+    # Limiting the stretches to the minimum cost supported by the algorithm
+    # will increase the performance of your test suite dramatically.
+    #
+    # For bcrypt, the minimum cost is 4. For other algorithms it defaults to 1.
```
ghiculescu, line 112: > For bcrypt, the minimum cost is 4.

This is news to me and probably many others, given the old default of 1 here. Is this documented anywhere? What happens if you use a value less than 4?

sergey-alekseev: Thanks for the questions. I thought the comment answers the question, but probably this should be clarified. Can you move your question to https://github.com/heartcombo/devise/pull/5092, so we proceed there? @ghiculescu

sergey-alekseev:
I rebased over the latest master to resolve minor merge conflicts and removed one commit as a dependency from this pull request. Originally I hoped for a quick merge, so I included a commit from #5092 to avoid merge conflicts.

sergey-alekseev:
Actually after excluding changes from https://github.com/heartcombo/devise/pull/5092 in this PR tests became inaccurate, so I had to apply [the changes above](https://github.com/heartcombo/devise/compare/6d18e1705406b8b0a73fe86bf0a941ec838d6749..80f65dda9ae4a5060c52b1248bb9b73a33de1734).

otherguy:
Would love this too!

PR #5148, opened 2019-10-03
Call set_flash_message helper instead of flash accessor

### LINKED ISSUE #5080: Unable to provide model-specific `already_authenticated` i18n key
```
en:
  devise:
    failure:
      already_authenticated: "You are already signed in."
      user:
        already_authenticated: ""
```

The latter with empty string is merely ignored. I wonder if it is a bug. If it's not, I find it useful to make it work in the same way as `sessions` key (https://github.com/plataformatec/devise#i18n).

gurgelrenan:
The flash accessor is being called directly instead of the #set_flash_message helper. The PR change this behavior calling `set_flash_message` helper.

fixes #5080

### DIFF
--- a/app/controllers/devise_controller.rb
+++ b/app/controllers/devise_controller.rb
@@ -112,7 +112,7 @@ MESSAGE
     end
 
     if authenticated && resource = warden.user(resource_name)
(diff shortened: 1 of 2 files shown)


tegon:
Thanks, @gurgelrenan!

PR #5149, opened 2019-10-07
Wrap shared links in div for easy styling

jonathanhefner:
I chose the class `devise-links` because it is neutral and explicit, but I can change it if there is a better name.

I also wanted to add a test to check for the presence of the class in the view, but I couldn't determine the best place to put such a test.  Any suggestions?

### DIFF
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -1,3 +1,4 @@
+<div class="devise-links">
 <%- if controller_name != 'sessions' %>
   <%= link_to "Log in", new_session_path(resource_name) %><br />
 <% end %>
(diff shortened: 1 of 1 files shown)


tegon:
Hi @jonathanhefner, thanks for your pull request.

This view is very simple and it's meant to be included as a default with Devise. As you can see, the texts are not using i18n and there are a bunch of `<br/>` tags to display the elements.

Usually, it's better to override this view so that you can better customize it in your application - this has more flexibility than the wrapper div. Would that work for you?

jonathanhefner:
@tegon I understand about overriding views, but this still seems like a convenient addition.  I always wrap the shared links after running `rails g devise:views`, because there is no way to style them otherwise.  In contrast, I only add i18n when necessary.

Do you feel like this change does too little compared to the editing that is typically required?  If so, are there additional changes you would like to see, e.g. remove `br` tags?

tegon:
This change is small, so it wouldn't be that much trouble to accept it. But it opens a precedent for the future when someone might need to include another tag or attribute in order to support their use case. 
Overriding seems like a more natural path for users to follow, and easier to maintain for us.

PR #5154, opened 2019-10-12
Remove unnecessary braces at token_generator

brunoocasali:
And include a new blank line to "separate" logic contexts when reading code.

Thanks for the awesome gem!

### DIFF
--- a/lib/devise/token_generator.rb
+++ b/lib/devise/token_generator.rb
@@ -4,7 +4,7 @@ require 'openssl'
 
 module Devise
   class TokenGenerator
-    def initialize(key_generator, digest = "SHA256")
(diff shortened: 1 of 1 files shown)


tegon:
Hi @brunoocasali, thanks for your contribution!

Since this is a purely aesthetic change, I'm going to close the pull request in favor of keeping the Git history tidy. Basically, when we're investigating a bug or the reason why something was added to the codebase, running a `git blame` here would hit this commit which would not be much helpful for those cases. We follow the same rationale as [Rails](https://github.com/rails/rails/pull/13771#issuecomment-32746700) does for those cases.

Thanks again!

PR #5166, opened 2019-11-26
Support dynamic password lengths for Rails > 5.2

manojmj92:
Hello,

Rails [now supports ](https://github.com/rails/rails/pull/30674) length validator to have dynamic min and max values.

I feel this would be a nice addition to have Devise support this feature. I have made the change such that no current installations of Devise will have trouble with this change. Everything should work as perfectly as before.

This change is useful for instances where applications have the need to support dynamic password lengths, without a system restart. Such applications will just have to override the `password_length` method in their own classes to support dynamic password lengths.

Real world use case: We have been trying to implement dynamic password length at GitLab. Implementation issue: https://gitlab.com/gitlab-org/gitlab/issues/36776
Right now, we have implemented the same using a [monkey patch for the password length validator](https://gitlab.com/gitlab-org/gitlab/-/blob/master/config/initializers/devise_dynamic_password_length_validation.rb#L37)


manojmj92:
Hello @tegon, could you kindly review? 🙂

manojmj92:
@carlosantoniodasilva could you please review? 🙂

carlosantoniodasilva:
@manojmj92 sorry about the delay here, things have been a bit crazy but this is on my radar to get back to you. Thanks for the work.

manojmj92:
Hello @carlosantoniodasilva, could you please review? 🙂

carlosantoniodasilva:
@manojmj92 yes, my apologies, this one slipped through the cracks, appreciate the ping. I've recently started to catch up on some of the issues/PRs here and this one is high on my radar, I'll get to it soon, just have a few other things to wrap up first.

ghiculescu:
> Such applications will just have to override the `password_length` method in their own classes to support dynamic password lengths.

It would be helpful to see an example, maybe in the CHANGELOG. I found https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/user.rb#L488 which gave me a better idea of what's being proposed.

Issue #5168, opened 2019-11-29
Use `exists?` instead of `find_first` in boolean checking


Dijkztra:
## Environment

- Ruby **2.4.4**
- Rails **5.2.0**
- Devise **3.5.6**

## Current behaviour
Some boolean checking query use `find_first` method. `find_first` will create a `select *` query when run, then the result will automatically converted into ActiveRecord object by Rails.

https://apidock.com/rails/v4.0.2/ActiveRecord/FinderMethods/find_first

## Expected behavior
Using `exists?` method make the resulting query is `select 1` and ActiveRecord doesn't need to convert the query result into a model object. Also, both Mongoid and Rails have `exists?` support.

https://apidock.com/rails/ActiveRecord/Base/exists%3F/class

PR #5169, opened 2019-11-29
User exists? instead of find_first

Dijkztra:
https://github.com/plataformatec/devise/issues/5168

### DIFF
--- a/lib/devise/token_generator.rb
+++ b/lib/devise/token_generator.rb
@@ -19,7 +19,7 @@ module Devise
       loop do
         raw = Devise.friendly_token
         enc = OpenSSL::HMAC.hexdigest(@digest, key, raw)
-        break [raw, enc] unless klass.to_adapter.find_first({ column => enc })
+        break [raw, enc] unless klass.to_adapter.exists?({ column => enc })
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/devise/token_generator.rb
```diff
@@ -19,7 +19,7 @@ def generate(klass, column)
       loop do
         raw = Devise.friendly_token
         enc = OpenSSL::HMAC.hexdigest(@digest, key, raw)
-        break [raw, enc] unless klass.to_adapter.find_first({ column => enc })
+        break [raw, enc] unless klass.exists?({ column => enc })
```
gabrielbaldao: In the another change you was using { key: value } and now { key => value }, what do you think about use one as default, I suggest { key: value }

carlosantoniodasilva:
@Dijkztra thanks for your contribution.

Devise uses https://github.com/ianwhite/orm_adapter in order to work with Active Record and Mongo without having to worry about specifics of each, which is why it uses `to_adapter.find_first`. I'm sure there's a slight gain here, as I understand `exists?` is meant to be faster, so if ORM Adapter provided us with a way to run it under the hood, we'd be all in in those spots. For now, however, I think we'll stick with `find_first`.

I am also unsure if mongo is working 100% in master right now, but we'll take a look at that eventually.

Issue #1047, opened 2011-05-04
Omniauthable for multiple models


levity:
Why is omniauthable restricted to a single model per app in devise 1.2? Is this limitation actively being worked on? If I were to try and hack up a workaround, what advice or warnings would you give me?

PR #5176, opened 2020-01-05
add support for multiple omniauthable models

john-denisov:
AFAIK `devise` doesn't support omniauthable for multiple models because long time ago `omniauth` did not support custom path_prefix for each strategy. But according to [this comment from @josevalim](https://github.com/plataformatec/devise/issues/1047#issuecomment-1101655) omniauth solved that issue at least 9 years ago.

I implemented a sample rails application that shows how omniauthable can be used for multiple devise models. 
main commit of that app: https://github.com/john-denisov/showcase-multiple-omniauthable-models/commit/7b1f2371fabbc6cc8b78215739c6309f7d7f788f
installation instructions: https://github.com/john-denisov/showcase-multiple-omniauthable-models/blob/master/README.md
I followed [this guide](https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview) to implement omniauth sign in for users and managers in my demo app

I will also update https://github.com/plataformatec/devise/wiki/OmniAuth-with-multiple-models if this PR gets approved 

Will appreciate any ideas how to improve this feature. Theoretically we can automatically duplicate all registered providers and set correct `path_prefix` for them if `devise :omniauthable` is specified in more than 1 model. That will reduce amount of configuration required but it won't allow to have separate sets of providers for different models 🤔

### DIFF
--- a/lib/devise/rails/routes.rb
+++ b/lib/devise/rails/routes.rb
@@ -441,7 +441,9 @@ ERROR
         end
         path_prefix = Devise.omniauth_path_prefix || "/#{mapping.fullpath}/auth".squeeze("/")
 
-        set_omniauth_path_prefix!(path_prefix)
+        unless mapping.to.omniauth_providers.all? { |provider| Devise.omniauth_configs[provider].options[:path_prefix].present? }
+          set_omniauth_path_prefix!(path_prefix)
+        end
 
         mapping.to.omniauth_providers.each do |provider|
           match "#{path_prefix}/#{provider}",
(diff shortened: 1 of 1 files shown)


tommyalvarez:
+1 for this one @ivan-denysov . Can I help in anyway to move this forward?

fractaledmind:
@ivan-denysov: This current setup doesn't quite cover the possible situations thoroughly.

Firstly, you might have more providers defined on the model than in the configuration (if you use the `:developer` provider in a test env, for example). If that is the case, then `mapping.to.omniauth_providers` will have an entry not present in `Devise.omniauth_configs.keys` and the code will throw an error since `Devise.omniauth_configs[provider]` will be `Nil`.

Secondly, you might have an additional provider defined in addition to the provider used on multiple models. In that case, you will likely have a configuration _without_ a `path_prefix` set and your code will return `false` forcing the setter and thus throwing the error we are looking to avoid.

Thirdly, you might have multiple providers used by multiple models.

Given these possibilities, I think the better articulation of the desired logic is:

>For every provider that has multiple configurations set, ensure that each configuration for each provider supplies a `path_prefix`; however, for any provider with only one configuration, set the path_prefix.

Here is my attempt to capture this logic:

```ruby
single_provider_with_multiple_configs = Devise.omniauth_configs.group_by { |_,config| config.provider }.select { |_,configs| configs.count > 1 }
all_configs_for_each_provider_with_multiple_configs_has_path_prefix_set = single_provider_with_multiple_configs.all? { |_,configs| configs.all? { |_, config| config.options[:path_prefix].present? } }

if single_provider_with_multiple_configs.empty? || !all_configs_for_each_provider_with_multiple_configs_has_path_prefix_set
  set_omniauth_path_prefix!(path_prefix)
end
```

PR #4824, opened 2018-03-30
issue 4526 adds skip email and password change notifications methods

iorme1:
I added methods to skip email change notifications and password change notifications on per edit basis like the issue was requesting.  I also added tests for these two methods.  All the tests are passing and the functionality worked when I tried it out on my test app.  This is my first PR so please let me know if I missed conforming to any guidelines or if I need to improve or adjust the code to fit your structure better. Thanks!

PR #5185, opened 2020-01-27
Respect @skip_email_changed_notification in overridden method.

smnscp:
The flag for skipping email change notification introduced in #4824 is ignored in this override of `send_email_changed_notification?`. When using both, DatabaseAuthenticatable and Confirmable, calling `skip_email_changed_notification!` won’t keep the notification from being sent.

### DIFF
--- a/lib/devise/models/confirmable.rb
+++ b/lib/devise/models/confirmable.rb
@@ -310,7 +310,7 @@ module Devise
         # requests the email change, instead of when the change is confirmed.
         def send_email_changed_notification?
           if self.class.reconfirmable
-            self.class.send_email_changed_notification && reconfirmation_required?
+            !@skip_email_changed_notification && self.class.send_email_changed_notification && reconfirmation_required?
           else
             super
           end


smnscp:
Hi @tegon, since you merged #4824 maybe you are the right one to have a look at this one.

tegon:
Hi @smnscp, thanks for your contribution.

Can you add a test so that we don't add any regressions to this behavior in the future?

smnscp:
Hi @tegon, sorry for forgetting tests. I added two cases now to also make sure those two `skip…!` methods do not interfere in some unexpected way. I’m not totally happy with testing behaviour of one Concern in the context of another. But I guess this is the price we pay for overriding protected methods.

smnscp:
Hi @tegon, I just wanted to check back if there is sth. more I can do since this PR still has the `Needs more work` label.

smnscp:
Hi @tegon, how’s this one?

carlosantoniodasilva:
@smnscp hey, @tegon has been away from the project for a while now, I've been planning on doing a sweep on issues/prs and I'm adding this one to the list to take a look first ok. Thanks!

smnscp:
Thanks, @carlosantoniodasilva!

PR #5187, opened 2020-01-31
Update README - add a short example to clarify.md

BKSpurgeon:
Add an example to clarify.


rafaelfranca:
Thank you for the pull request but this clarification is on the previous block of text.

PR #5193, opened 2020-02-19
Performance improvement for timeoutable

pickerflicker:
move timeout_skip check earlier to avoid unnecessary record lookup

### DIFF
--- a/lib/devise/hooks/timeoutable.rb
+++ b/lib/devise/hooks/timeoutable.rb
@@ -21,8 +21,8 @@ Warden::Manager.after_set_user do |record, warden, options|
 
     proxy = Devise::Hooks::Proxy.new(warden)
 
-    if record.timedout?(last_request_at) &&
-        !env['devise.skip_timeout'] &&
+    if !env['devise.skip_timeout'] &&
+        record.timedout?(last_request_at) &&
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
Thanks for your contribution. Would you mind clarifying what you mean by "unnecessary record lookup" here? Is that an actual DB lookup (which I'm not clearly seeing from the code), or just skipping the timestamp check that happen in `timedout?`:

https://github.com/heartcombo/devise/blob/29943a26e61d68198666c59e07457dba3c75f581/lib/devise/models/timeoutable.rb#L30-L32

pickerflicker:
@carlosantoniodasilva This is to skip the user defined `timeout_in` method on the user model. I'd imagine most of the time this would result in a db lookup, ie: if user admin or has some special role.

carlosantoniodasilva:
We can't really account for what people may implement in their model, but more often than not, that shouldn't result in a DB lookup. Even being admin or a special role, it would be dependent upon the client implementation -- and I'm inclined to think most people that use this module don't even override that method, very few would do.

That said, I see no harm in reversing those conditionals and bailing out early based on the "skip" flag set in the env, so I'll merge it. Thanks for the contribution.

Issue #4951, opened 2018-10-06
current_user should not perform sign in by default


vincentwoo:
## Environment

- Ruby **[2.5.1]**
- Rails **[5.2]**
- Devise **[4.5]**

## Current behavior

It is very hard to prevent sign in with Devise. If you subclass SessionsController to override create, even if you don't allow super to run, any template code that tries to check `current_user` will automatically run warden's authentication, which signs in the user.

## Expected behavior

Users should only be signed in after an invocation of `sign_in`.

lacostenycoder:
@vincentwoo can you provide a code example of how to reproduce the current behavior in this issue?

vincentwoo:
I can later but this snippet from the devise codebase explains the behavior: https://github.com/plataformatec/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/lib/devise/controllers/helpers.rb#L60

vincentwoo:
Okay, so let's say you have a SessionsController with the goal to prevent logins conditionally:

```ruby
class SessionsController < Devise::SessionsController
  def create
    user = User.find_by_email(sign_in_params[:email])
    if user.some_condition?
      return redirect_to some_other_url
    end
    super # otherwise login as usual
  end
end
```

If, say, the view for `some_other_url` calls `current_user`, the the user will be logged in anyway. This seems bad.

brendancarney:
I agree. This made it more challenging to implement two-factor auth. We want to redirect to the two-factor auth page after checking the user's password, but some other code (like session tracking) checks if there is a `current_user`, causing the user to be logged in.

lacostenycoder:
@brendancarney what did you do as a workaround?

lacostenycoder:
@vincentwoo This may seem obvious to you, but, why would you call `current_user` in a view for `some_other_url` if that user should not be authenticated in the first place?  If you want to insure user is not authenticated?  Why would you not just do this?

```ruby
class SessionsController < Devise::SessionsController
  def create
    user = User.find_by_email(sign_in_params[:email])
    if user.some_condition?
      sign_out_and_redirect current_user
    end
    super # otherwise login as usual
  end
end
```

and handle where to send them in `ApplicactionController#after_sign_out_path_for(user)` ?

vincentwoo:
Because you want to conditionally show a different view to logged in users who mistakenly end up on that page through an old link from email, slack, etc.

vincentwoo:
[...]

PR #5202, opened 2020-03-03
Add config setting to disable authenticating when calling current_user

joeyparis:
This addresses the discussion in #4951. I understand the desire to keep the behavior as is since it's been in the code base for over 10 years and is used in who knows how many different systems. That being said, I still see the value in not automatically authenticating the user when calling `current_user`. You can read the discussion in #4951 for both sides of the argument.

This pull request introduces an opt-in config setting to disable automatic authentication when calling `current_user`. This has no change for users that don't opt-in to this setting.

I will be honest, I am not happy with the testing that I added. It should probably be reviewed and improved by someone more familiar with the Devise internals than me.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -297,6 +297,10 @@ module Devise
   mattr_accessor :sign_in_after_change_password
   @@sign_in_after_change_password = true
 
+  # When set to true, calling current_user will attempt to authenticate the user
+  mattr_accessor :current_user_attempts_login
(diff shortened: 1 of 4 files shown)


joeyparis:
Nearly a year and a half later, but I was able to finally write some proper tests that I feel comfortable with. Hopefully, a maintainer can take a look at this now that it has actual testing. I'd really like to see this in production!

PR #5203, opened 2020-03-03
Backport CVE-2019-16109 fix to 3-stable

StephenCharles:
Original fix: https://github.com/heartcombo/devise/pull/5132


rafaelfranca:
Thank you for the pull request but the version 3.x of Devise is not maintained anymore, even for security fixes.

PR #5204, opened 2020-03-05
Clarify DatabaseAuthenticable's behaviour for password fields

unleashy:
`DatabaseAuthenticable` overrides the `password=` method in a model, bypassing setting that attribute, and this should be reflected in the documentation.

We got bitten by this at work because we're using a legacy database where the User model already has a `password` field and upon the introduction of Devise the password stopped being stored there; we had to override the method again to get our needed behaviour but having this in the documentation would've saved us a few hours of debugging.

Thanks!

### DIFF
--- a/lib/devise/models/database_authenticatable.rb
+++ b/lib/devise/models/database_authenticatable.rb
@@ -7,9 +7,9 @@ module Devise
     # Authenticatable Module, responsible for hashing the password and
     # validating the authenticity of a user while signing in.
     #
-    # This module overrides the `password=` method and skips setting the password
-    # when Rails is setting attributes, instead it hashes the plaintext password
-    # and stores it in `encrypted_password` for legacy reasons.
+    # This module defines a `password=` method. This method will hash the argument
+    # and store it in the `encrypted_password` column, bypassing any pre-existing
+    # `password` column if it exists.
     #
     # == Options
     #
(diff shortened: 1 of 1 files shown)


p8:
Passwords should never be stored plain text in the database because it's very insecure.
There are countless of companies making headlines because of passwords leaking.

Devise stores a hash of the password in the encrypted_password column.
You should migrate the passwords to the encrypted_password column and remove the old password column.

lib/devise/models/database_authenticatable.rb
```diff
@@ -7,6 +7,10 @@ module Models
     # Authenticatable Module, responsible for hashing the password and
     # validating the authenticity of a user while signing in.
     #
+    # This module overrides the `password=` method and skips setting the password
```
rafaelfranca: It is correct to say it defines a `password=` but not that it skips setting the password since for what this gem is concerned, the table should not have a password column.

So I think this would be something I'd accept:

```
# This module defines a `password=` method. This method will hash the argument and store it in the
# `encrypted_password` column.

unleashy:
@p8 We are **not** storing the password in plaintext! We are storing it hashed and salted, but in the `password` field instead of the `encrypted_password` field that Devise requires, because we're using  a legacy database.

lib/devise/models/database_authenticatable.rb
unleashy: Fair enough, changes made, I believe it's accurate now? sorry for the inaccuracy :)

rafaelfranca:
Thank you :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

p8:
@unleashy That's good to know 😄 👍

PR #5208, opened 2020-03-12
Explain how changing stretches affects existing password hashes

p8:
I could not find documentation about what happens if you change stretches.


sergey-alekseev:
@p8 [here](https://github.com/heartcombo/devise/pull/5074#issuecomment-492687076) you can find the detailed flow.

sergey-alekseev:
I may follow up to this update with an improvement later when I have time because it's not really precise.

PR #5214, opened 2020-03-26
Fix kwargs usage for Ruby 2.7

aubinlrx:
Fix one deprecation warning. This prevents us from using behavior that was deprecated in Ruby 2.7.

### DIFF
--- a/lib/devise/test/controller_helpers.rb
+++ b/lib/devise/test/controller_helpers.rb
@@ -31,8 +31,8 @@ module Devise
       end
 
       # Override process to consider warden.
-      def process(*)
-        _catch_warden { super }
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request but this beaks the gem with Ruby < 2.7. I think here we need to use `ruby2_keywords`, or probably some more complicated code.

Laykou:
@rafaelfranca Is this waiting for change from @aubinrlx?

I think more people will have problems with this since Devise is pretty popular and projects are being slowly upgraded to Ruby 2.7

carlosantoniodasilva:
Replaced by https://github.com/heartcombo/devise/pull/5235.

PR #5232, opened 2020-05-02
Updated README with additional video resources for learning Devise

kobaltz:
I'm preparing an episode on Drifting Ruby to demonstrate on how to contribute to a gem, so I made this small PR; adding a resource of information into the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,11 +166,12 @@ bin/test test/models/trackable_test.rb:16
 
 ## Starting with Rails?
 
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
@kobaltz appreciate the PR, unfortunately I had to revert it right after I merged: https://github.com/heartcombo/devise/commit/a451b98cf2311b6bb73b3762083294db131a4603.

I realized too late that it was included in the section that talks about not using Devise, so including a link for a Devise specific video there doesn't go along well, unfortunately. If you have any other resource that's authentication with Rails specific, but not Devise, we're happy to include. Thanks again!

PR #5250, opened 2020-06-09
remove useless rails51? method

hyuraku:
`self.rails51?` method is not using in `lib`, and the Only `test` uses the method.
I guess better to remove this method.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -297,10 +297,6 @@ module Devise
   mattr_accessor :sign_in_after_change_password
   @@sign_in_after_change_password = true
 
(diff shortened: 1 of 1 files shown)


pnghai:
this will crash devise-token-auth

carlosantoniodasilva:
@pnghai unfortunately they should not be using a `nodoc` method that happens to be exposed, it is meant to be a private Devise implementation/detail, and thus we should be free to change/remove it any time (in other words, it's not Devise's public interface to be consumed). We don't have control over that though, but they'll have to update their implementation. Feel free to send a PR there.

PR #5258, opened 2020-06-18
Fix warning from `thor`

deivid-rodriguez:
The `:orm` option can also have symbol values.

Fixes https://github.com/heartcombo/devise/issues/5252.


carlosantoniodasilva:
Thanks!

PR #5265, opened 2020-07-10
Increase the default minimum password length from 6 to 8

jpawlyn:
Increasing the minimum password length from 6 to 8 was done in 2012 (https://github.com/heartcombo/devise/commit/c179cef365f7188c91cbbc3db924a9f1f9563c3c#diff-ee89a7275db7a2af385d05e737aa11f8). This change was subsequently reverted but the default in the initializer template was uncommented (https://github.com/heartcombo/devise/commit/bdf0bc7b1e66d1a70e4ecda8507145658a4f1a96).

Since the value in the initializer has been uncommented for so long, there should be little chance of this update causing a breaking change.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -117,7 +117,7 @@ module Devise
 
   # Range validation for password length
   mattr_accessor :password_length
-  @@password_length = 6..128
+  @@password_length = 8..128
 
   # The time the user will be remembered without asking for credentials again.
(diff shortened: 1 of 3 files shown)


jpawlyn:
Just curious if you think it would make sense to increase the minimum password length from 6 to 10 instead of 8 🤔  ?

PR #5278, opened 2020-08-12
serializable_hash deeper dup

dpep:
I ran into an issue where options[:except] is a frozen array, which explodes when we try to concat values in `serializable_hash`.  This also means that we're generally modifying the passed in value, despite the `dup`, since it's shallow. To fix this, could we do a deep dup?

alternatively, we could do:
```
options[:except] = Array(options[:except]).dup
```

### DIFF
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -104,7 +104,7 @@ module Devise
       # given to :except will simply add names to exempt to Devise internal list.
       def serializable_hash(options = nil)
         options = options.try(:dup) || {}
-        options[:except] = Array(options[:except])
+        options[:except] = Array(options[:except]).dup
 
         if options[:force_except]
           options[:except].concat Array(options[:force_except])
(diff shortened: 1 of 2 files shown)


lib/devise/models/authenticatable.rb
```diff
@@ -103,7 +103,7 @@ def authenticatable_salt
       # and passing a new list of attributes you want to exempt. All attributes
       # given to :except will simply add names to exempt to Devise internal list.
       def serializable_hash(options = nil)
-        options = options.try(:dup) || {}
+        options = options.try(:deep_dup) || {}
```
carlosantoniodasilva, line 106: Could we use your alternative suggestion, `dup` on the line below? I think it makes it more explicit about what we're duping / touching, and it will prevent duping other objects that may be passed in that'd be duped with `deep_dup`.

You can also update this test: https://github.com/heartcombo/devise/blob/f26e05c20079c9acded3c0ee16da0df435a28997/test/models/serializable_test.rb#L45-L47, maybe something like this to catch this scenario as well:

```ruby
assert_key "username", @user.as_json({ only: :username, except: [:email].freeze }.freeze)["user"]
```

dpep: done and done - both solid suggestions, thanks!

carlosantoniodasilva:
Merged, thanks @dpep. :+1:

PR #5279, opened 2020-08-13
Adding Persian/Farsi translation 

prp-e:
I translate strings as much as I could. If there's anything more to be fixed, I'd be pleased to know.


carlosantoniodasilva:
Thanks for your contribution @prp-e. Unfortunately devise does not store custom locales in its own repo, but we keep a wiki with links to translation files in https://github.com/heartcombo/devise/wiki/I18n that anyone can contribute to, and there's a devise-i18n extension / gem that you could take a look and perhaps submit your translations to: https://github.com/tigrish/devise-i18n.

Issue #2332, opened 2013-03-15
ActionController::Live and authenticate_user! fails to catch :warden


jamilbk:
Calling authenticate_user! in a controller that has included ActionController::Live causes `ArgumentError (uncaught throw :warden)`

Can be reproduced simply with:

``` ruby
class FooController < ApplicationController
  include ActionController::Live
  before_action :authenticate_user!
  def foo; render text: ""; end
end
```

And then GET'ing foo from the rails console. Example backtrace:

```
ArgumentError (uncaught throw :warden):
  warden (1.2.1) lib/warden/proxy.rb:128:in `throw'
  warden (1.2.1) lib/warden/proxy.rb:128:in `authenticate!'
  /Users/jamil/.rvm/gems/ruby-2.0.0-p0@tmp/bundler/gems/devise-d29b744d92a3/lib/devise/controllers/helpers.rb:48:in `authenticate_user!'
  activesupport (4.0.0.beta1) lib/active_support/callbacks.rb:407:in `_run__624670321848815552__process_action__callbacks'
  activesupport (4.0.0.beta1) lib/active_support/callbacks.rb:78:in `run_callbacks'
  actionpack (4.0.0.beta1) lib/abstract_controller/callbacks.rb:17:in `process_action'
  actionpack (4.0.0.beta1) lib/action_controller/metal/rescue.rb:29:in `process_action'
  actionpack (4.0.0.beta1) lib/action_controller/metal/instrumentation.rb:31:in `block in process_action'
  activesupport (4.0.0.beta1) lib/active_support/notifications.rb:158:in `block in instrument'
  activesupport (4.0.0.beta1) lib/active_support/notifications/instrumenter.rb:20:in `instrument'
  activesupport (4.0.0.beta1) lib/active_support/notifications.rb:158:in `instrument'
  actionpack (4.0.0.beta1) lib/action_controller/metal/instrumentation.rb:30:in `process_action'
  actionpack (4.0.0.beta1) lib/action_controller/metal/params_wrapper.rb:245:in `process_action'
  activerecord (4.0.0.beta1) lib/active_record/railties/controller_runtime.rb:18:in `process_action'
  actionpack (4.0.0.beta1) lib/abstract_controller/base.rb:136:in `process'
  actionpack (4.0.0.beta1) lib/abstract_controller/rendering.rb:44:in `process'
  actionpack (4.0.0.beta1) lib/action_controller/metal/live.rb:118:in `block in process'


[...]

PR #5286, opened 2020-08-26
Fix hanging tests for streaming controllers using Devise

### LINKED ISSUE #5285: Devise tests hang when a controller using ActionController::Live calls authenticate_user!
## Current behavior

Given a streaming controller that uses `authenticate_user!`:

```ruby
class MyController < ApplicationController
  include ActionController::Live
  before_action :authenticate_user!
  ...
end
```

controller tests will hang when authentication fails. This is because `#_process_unauthenticated` is setting the response body with `@controller.response.body=` rather than `@controller.response_body=`. The subtle difference is that ActionController::Live [overrides](https://github.com/rails/rails/blob/ee6ee11b01dd8259ab2e8998d8d4984ddf39e340/actionpack/lib/action_controller/metal/live.rb#L280) `@controller.response_body=` to close the stream after setting the body; if the stream isn't closed, the thread will hang indefinitely.

## Expected behavior

Does not hang indefinitely. :)

I've got a PR coming in a few minutes.

afn:
Fixes #5285.

### DIFF
--- a/lib/devise/test/controller_helpers.rb
+++ b/lib/devise/test/controller_helpers.rb
@@ -143,7 +143,7 @@ module Devise
           @controller.response.headers.merge!(headers)
           @controller.response.content_type = headers["Content-Type"] unless Rails::VERSION::MAJOR >= 5
           @controller.status = status
-          @controller.response.body = response.body
+          @controller.response_body = response.body
           nil # causes process return @response
         end
 
--- a/test/rails_app/app/controllers/streaming_controller.rb
+++ b/test/rails_app/app/controllers/streaming_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class StreamingController < ApplicationController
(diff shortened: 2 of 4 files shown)


afn:
Whoops! In Rails 6 (and 5), #2332 doesn't impact controller tests, but apparently in Rails 4 it does. Should be fixed now.

carlosantoniodasilva:
:+1: thanks!

afn:
Any idea when this will be released (to rubygems, that is?)?

carlosantoniodasilva:
I don't have an ETA, but it looks like I'll have to cut something soon ;)

I had someone asking: https://github.com/heartcombo/devise/issues/5252#issuecomment-691989588, but I have to check on a couple other things first in regards to what's in master. Hopefully 🔜

carlosantoniodasilva:
@afn v4.7.3 is out including this change, thanks!

PR #5297, opened 2020-09-29
Allow overriding the source of the password_reset_token

artfuldodger:
Changes no behavior by default, but this allows for pulling the token
out of the session for example.

The motivation behind this is to stop leaking the token to third parties
when tracking JavaScript is present on the page (e.g. Google Analytics).

We're overriding `edit` to set the token in the session and then redirect,
taking it out of the URL, and this change would enable us to be slightly
less coupled to Devise's internals.

### DIFF
--- a/app/controllers/devise/passwords_controller.rb
+++ b/app/controllers/devise/passwords_controller.rb
@@ -26,7 +26,7 @@ class Devise::PasswordsController < DeviseController
   def edit
     self.resource = resource_class.new
     set_minimum_password_length
-    resource.reset_password_token = params[:reset_password_token]
+    resource.reset_password_token = reset_password_token_from_request
   end
 
   # PUT /resource/password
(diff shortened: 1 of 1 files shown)


crespire:
+1 on this, I had to do this today, and was one of the strategies I saw in https://thoughtbot.com/blog/is-your-site-leaking-password-reset-links

Polaris253:
what is the point of setting a password if someone without the password can overide a password?

Polaris253 (changes requested):
A password is meant to secure something, allowing an override defeats the purpose. @github/support What do you think about these updates?

artfuldodger:
@Polaris253 hi! So you know when you forget your password, and you submit a request to reset your password? It sends you an email containing a link with a password reset token in the url. This PR just makes it easy to override where that token is pulled from, as urls are often logged to places you may not want your tokens stored.

PR #5301, opened 2020-10-16
Allow mailer methods to receive a block

jfo84:
Since the internal method `devise_mail` takes a block and ActionMailer's `mail` takes a block, I thought that it made sense for the devise mailer methods to also receive a block to make format overrides a bit easier.


SalamanderHP:
ready

PR #5306, opened 2020-11-02
chore: Cleanup old rack/rails session bug fix

janz93:
Seven years ago rails `session.keys` could be empty if the session was not loaded yet.

To prevent an error the removed code was introduced
https://github.com/heartcombo/devise/issues/2660

Since then rails changed the behavior and makes sure that the session is loaded before someone wants to access any session keys
https://github.com/rails/rails/blob/3c99817394f634710a6280c5d5512c0ea663bce3/actionpack/lib/action_dispatch/request/session.rb#L120

Which means the `session.empty?` is not needed anymore

### DIFF
--- a/lib/devise/controllers/sign_in_out.rb
+++ b/lib/devise/controllers/sign_in_out.rb
@@ -106,10 +106,12 @@ module Devise
       private
 
       def expire_data_after_sign_in!
+        # TODO: remove once Rails 5.2+ and forward are only supported.
         # session.keys will return an empty array if the session is not yet loaded.
         # This is a bug in both Rack and Rails.
         # A call to #empty? forces the session to be loaded.
         session.empty?
+
         session.keys.grep(/^devise\./).each { |k| session.delete(k) }
       end
 


carlosantoniodasilva (changes requested):
Thanks for your contribution.

It looks like this change was added here in Rails: https://github.com/rails/rails/commit/3498aacbbebb41e529b6755f4ccfdfbb84c28830, but only exists on Rails 5.2+. Unfortunately we still support Rails versions prior to that so we can't remove that line yet.

Could you please change it to mention this has been fixed in Rails 5.2+ and we can remove it once we only support that version forward? Thanks.

lib/devise/controllers/sign_in_out.rb
```diff
@@ -106,10 +106,12 @@ def sign_out_all_scopes(lock = true)
       private
 
       def expire_data_after_sign_in!
+        # TODO: remove once Rails 5.2+ and forward are only supported.
```
janz93, line 109: I personally would be fine with having the answer to why the `TODO` was added only in the commit message. 

But I am not sure and would love to get someone else's opinion

carlosantoniodasilva: Having the reasoning in the commit message only is fine in this case, we can trace if down if needed from git. Thanks.

PR #5310, opened 2020-11-06
Allow wrapping of lockable Warden after sign-in hook

### LINKED ISSUE #5264: Proposal: a way to handle multi-DB ActiveRecord setups
Rails 6 provides connection switching to allow for applications with multiple databases. One of the supported behaviors is a primary/replica setup where the replica database is read-only. Devise users will run into some issues with the default Rails 6 connection switching, [as it restricts writes to non-`GET/HEAD/OPTIONS` requests](https://guides.rubyonrails.org/active_record_multiple_databases.html#activating-automatic-connection-switching). 

As a result, some requests will trigger `ActiveRecord::ReadOnlyError` as a result of hooks firing, etc. Some of the Devise methods can be worked around directly in models like this:

```ruby
# app/models/user.rb
def forget_me!
  ActiveRecord::Base.connected_to(role: :writing) do
    super
  end
end
```

Unfortunately some of the cases where writes inside a `GET` request don't seem quite as straightforward to work around. Some of the Warden `after_sign_in` hooks call `#save` directly on the model. Since Warden hooks are additive, it's not easy to replace the hook with a monkey-patched one. We're currently experiencing some failures for `lockable` that we (believe we) are not able to code around without forking Devise. Example:

https://github.com/heartcombo/devise/blob/f39c6fd92774cb66f96f546d8d5e8281542b4e78/lib/devise/hooks/lockable.rb#L5-L12

I'd like to propose a pull request to make the Rails 6 primary/replica functionality work without requiring any changes for users who are on older versions of Rails or who aren't leveraging ActiveRecord's `connected_to` functionality. There are two ways I can see to approach this:

1. Require users to wrap their model methods in a `connected_to` clause around `super` as described above. `hooks/lockable` would need to be updated to push [its call to `#save`](https://github.com/heartcombo/devise/blob/f39c6fd92774cb66f96f546d8d5e8281542b4e78/lib/devise/hooks/lockable.rb) to a model method which could then be wrapped. Then we would document that this is the recommended approach for users leveraging primary/replica setups. This would be the smallest change, but would be less convenient for multiple-database users in Rails 6.

2. Add a new configuration option, something named like `devise_database_role`. If that role is set, then all model saves in model modules or hooks would get wrapped in a `connected_to` block using that role. If the setting is not defined, then all saves happen directly as now. This approach would be more comprehensive and be a one-stop change for primary/replica setups vs. needing to patch each relevant method, but requires more invasive changes across several model modules.

I'm excited to tackle either of these changes, but looking for feedback on the preferred approach.

Related: #5133

geoffharcourt:
For applications using a primary/replica multiple database setup, Rails
6's default connection switching can lead to the `Lockable` after
sign-in hook firing on `GET` requests, getting the read-only database,
and then attempting to write while zeroing out the user's failed attempts.

This change adds a customizable wrapper proc that can allow applications
to customize this behavior. For a primary/replica setup using
ActiveRecord's default connection switching, it might look like this:

```ruby
Devise.setup do |config|
  # default is Proc.new { |hook| hook.call }
  config.warden_hook_save_wrapper = Proc.new do |hook|
    # ensure the writable connection is used to avoid read-only write errors
    ApplicationRecord.connected_to(role: :writing) do
      hook.call
    end
  end
end
```

This ensures that writes don't happen in a read-only connection if that's the 
current connection context.

Fix #5264

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -297,6 +297,9 @@ module Devise
   mattr_accessor :sign_in_after_change_password
   @@sign_in_after_change_password = true
 
+  mattr_accessor :warden_hook_save_wrapper
+  @@warden_hook_save_wrapper = Proc.new { |hook| hook.call }
+
   def self.activerecord51? # :nodoc:
     defined?(ActiveRecord) && ActiveRecord.gem_version >= Gem::Version.new("5.1.x")
   end
--- a/lib/devise/hooks/lockable.rb
+++ b/lib/devise/hooks/lockable.rb
@@ -5,8 +5,12 @@
 Warden::Manager.after_set_user except: :fetch do |record, warden, options|
   if record.respond_to?(:failed_attempts) && warden.authenticated?(options[:scope])
     unless record.failed_attempts.to_i.zero?
-      record.failed_attempts = 0
-      record.save(validate: false)
+      wipe_attempts = Proc.new do
+        record.failed_attempts = 0
+        record.save(validate: false)
+      end
+
+      Devise.warden_hook_save_wrapper.call(wipe_attempts)
     end
   end
 end


geoffharcourt:
Note: I'm not sure how to set up proper Warden hook tests here.

mhenrixon:
Any news on this? Has been a minute now :)

carlosantoniodasilva:
Geoff and I have chatted (yesterday actually) about it, and here's the outcome we're going to try: instead of a new configuration to wrap this particular warden hook, we'll expose a model method that will be called from the hook here, similar to [how we do for trackable](https://github.com/heartcombo/devise/blob/0cd72a56f984a7ff089246f87a8b259120545edd/lib/devise/hooks/trackable.rb#L9), which will give you control over that in your model code to wrap it and call `super`.

This way you can at least override all the pieces inside the app to switch to a writing connection in Devise methods. As a second step we can look into making Devise more aware of multi-DB support and have it do the work of switching connections, but that's a bigger change we need to look at more carefully and from a broader perspective.

carlosantoniodasilva:
@geoffharcourt @mhenrixon I'm giving this one a shot in https://github.com/heartcombo/devise/pull/5369, let me know about any feedback/concerns there please. Thanks!

geoffharcourt:
Closing in favor of #5369! Thank you @carlosantoniodasilva!

PR #5317, opened 2020-11-23
Fix kwargs usage for Ruby 2.7

stanhu:
This updates controller_helpers.rb so that the `process` method is
compatible with other controller helpers in Ruby 2.7.


stanhu:
Closing in favor of https://github.com/heartcombo/devise/pull/5235.

PR #5319, opened 2020-12-08
More consistent use of `#set_flash_message!`

afrase:
When using devise on an API only rails application some controllers still tried to add to the `request.flash` object.

- Use `#set_flash_message!` instead of `#set_flash_message` in `app/controllers/devise_controller.rb`, `app/controllers/devise/passwords_controller.rb`, and `app/controllers/devise/registrations_controller.rb`.
- Consistently use parenthesis when calling `#set_flash_message!`.

### DIFF
--- a/app/controllers/devise/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise/omniauth_callbacks_controller.rb
@@ -8,7 +8,7 @@ class Devise::OmniauthCallbacksController < DeviseController
   end
 
   def failure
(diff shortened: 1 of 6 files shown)


carsomyr:
Just ran into a `undefined method \`flash' for #<ActionDispatch::Request GET "http://localhost:3000/api/user_session/new" for ::1>` message because of this hasn't been merged yet.

We have `Rails.application.config.api_only = true`, which disables flash middleware, which causes the error.

PR #5323, opened 2020-12-31
Test on rails 6.1

tricknotes:
Rails 6.1 has been released.
https://edgeguides.rubyonrails.org/6_1_release_notes.html

This PR added CI matrix for Rails 6.1.


test/models/authenticatable_test.rb
```diff
@@ -30,12 +30,12 @@ class AuthenticatableTest < ActiveSupport::TestCase
 
   test 'find_or_initialize_with_errors adds blank error' do
     user_with_error = User.find_or_initialize_with_errors([:email], { email: "" })
-    assert_equal [:email, "can't be blank"], user_with_error.errors.first
+    assert_equal ["Email can't be blank"], user_with_error.errors.full_messages_for(:email)
```
a-baqtyan, line 33: assert_equal ['Email can't be blank'], user_with_error.errors.full_messages_for(:email)

Use single quotations

tricknotes: This string contains `'`, so we should escape it as back slash in the single quotations.
Therefore, using `"` makes it readable.

Why do you think so?

carlosantoniodasilva: Double quotes is fine here, thanks.

carlosantoniodasilva:
Thanks!

PR #5324, opened 2021-01-01
Clarify Turbo vs Turbolinks requirements

ghiculescu:
Follow up to https://github.com/heartcombo/devise/pull/4678

I spent a long time digging around the `Turbo` source code looking for a `Turbolinks::Controller` equivalent while trying to do an upgrade, then I realised is no longer necessary.

### DIFF
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -297,7 +297,8 @@ Devise.setup do |config|
   # config.omniauth_path_prefix = '/my_engine/users/auth'
 
   # ==> Turbolinks configuration
-  # If your app is using Turbolinks, Turbolinks::Controller needs to be included to make redirection work correctly:
+  # If your app is using Turbolinks (but *not* Turbo), Turbolinks::Controller needs
+  # to be included to make redirection work correctly:
   #
   # ActiveSupport.on_load(:devise_failure_app) do
   #   include Turbolinks::Controller


entretechno-jeremiah:
Looks like different configuration is necessary for Turbo: [_How to use Devise with Hotwire & Turbo.js (Example) | GoRails_](https://gorails.com/episodes/devise-hotwire-turbo)

lib/generators/templates/devise.rb
```diff
@@ -297,7 +297,8 @@
   # config.omniauth_path_prefix = '/my_engine/users/auth'
 
   # ==> Turbolinks configuration
-  # If your app is using Turbolinks, Turbolinks::Controller needs to be included to make redirection work correctly:
+  # If your app is using Turbolinks (but *not* Turbo), Turbolinks::Controller needs
```
rafaelfranca, line 300: Do we need to tell that Turbolinks is different of Turbo? I thought it was clear by the name that requirements for one are different of the other.

ghiculescu: Maybe it's just me, but I assumed that the requirements would be the same since Turbo (Drive) is being pitched an evolution of Turbolinks.

ghiculescu:
I'm gonna close this for now as it looks like it's not that simple, see also https://github.com/hotwired/turbo-rails/issues/34 (cc @bhaibel)

lib/generators/templates/devise.rb
miharekar: If I'm not mistaken this doesn't even work for current (now deprecated) Turbolinks, but only for Turbolinks Classic.

PR #5328, opened 2021-01-08
Extract kwargs to fix 2.7 warnings

igor-drozdov:
The ControllerHelpers#process method simply propagates the invokation to another caller, for example to https://github.com/rails/rails-controller-testing/blob/4b15c86e82ee380f2a7cc009e470368f7520560a/lib/rails/controller/testing/template_assertions.rb#L60

This commit extracts kwargs from args in order to handle them properly

### DIFF
--- a/lib/devise/test/controller_helpers.rb
+++ b/lib/devise/test/controller_helpers.rb
@@ -31,7 +31,7 @@ module Devise
       end
 
       # Override process to consider warden.
-      def process(*)
+      def process(*, **)
         _catch_warden { super }
 
         @response


igor-drozdov:
@carlosantoniodasilva wdyt about this change?  could you please have a look?

carlosantoniodasilva:
Thanks @igor-drozdov. We don't seem to have any more warnings on Ruby 2.7, I believe due to the usage of `ruby2_keywords` there... the change you propose in this branch though did introduce a new warning:

> /home/travis/build/heartcombo/devise/lib/devise/test/controller_helpers.rb:40: warning: Skipping set of ruby2_keywords flag for process (method accepts keywords or method does not accept argument splat)

Maybe we don't need ruby2_keywords there after all? I think we might for compatibility with older rubies since Devise still supports a lot of them. (not for long, hopefully)

igor-drozdov:
> I think we might for compatibility with older rubies since Devise still supports a lot of them

yea, I think this is a reasonable case for using ruby2_keywords 👍 

> We don't seem to have any more warnings on Ruby 2.7

I think you're right, ruby2_keywords should do the job, I've mistaken here. I'm sorry for the noise and thank you for the quick and great response!

PR #5331, opened 2021-01-17
Update omniauthable tests for OmniAuth 2.0

jkowens:
Hope this helps with https://github.com/heartcombo/devise/pull/5327. Updated omniauthable links to use `method: :post` and revised corresponding tests.  I opened a PR with omniauth-openid (https://github.com/omniauth/omniauth-openid/pull/44), but for now I updated the Gemfile to pull it from a branch on my fork.

### DIFF
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -20,6 +20,6 @@
 
 <%- if devise_mapping.omniauthable? %>
   <%- resource_class.omniauth_providers.each do |provider| %>
-    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider) %><br />
+    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), method: :post %><br />
   <% end %>
 <% end %>
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,7 @@ gem "responders", "~> 3.0"
 
 group :test do
   gem "omniauth-facebook"
-  gem "omniauth-openid"
+  gem "omniauth-openid", git: 'https://github.com/jkowens/omniauth-openid', branch: 'patch-1'
   gem "timecop"
   gem "webrat", "0.7.3", require: false
   gem "mocha", "~> 1.1", require: false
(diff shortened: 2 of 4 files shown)


carlosantoniodasilva:
Thanks @jkowens, I'll take a look and report back.

app/views/devise/shared/_links.html.erb
```diff
@@ -20,6 +20,6 @@
 
 <%- if devise_mapping.omniauthable? %>
   <%- resource_class.omniauth_providers.each do |provider| %>
-    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider) %><br />
+    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), method: :post %><br />
```
carlosantoniodasilva, line 23: @BobbyMcWho mind if I ask for some clarification here?

In the [CVE wiki](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284#mitigating-in-rails-applications) it mentions using link_to with method=post or a button_to should be enough, but in the release notes for [OmniAuth v2](https://github.com/omniauth/omniauth/releases/tag/v2.0.0) you mention specifically:

> If you are using hyperlinks or buttons styled to redirect to your login route, you should update these to be a submit input or a submit type button wrapped in a form.

I'm assuming you meant those as being GET requests, that should be updated to using POST instead? (since Rails button_to or link_to + method=post use forms under the hood?)

BobbyMcWho: We want to make sure that whatever way we send the POST, we're including a CSRF token to be validated in the request_validation_phase. I suggested wrapping inputs with a rails form helper since by default those automatically generate a per-form token that can be validated by Rails' RequestForgeryProtection class

jkowens: A link_to with method=post will submit a form with an authenticity token. I assume it is a per-form token, but not sure 🤔

carlosantoniodasilva: Right, thanks for the clarification :+1:. I believe link_to + method=post or a button_to should be fine then, like @jkowens mentioned they should also include the authenticity token by default in Rails. (whether they are per-form or not is controlled by the application via the individual form and/or global Rails config iirc.)

carlosantoniodasilva:
I'm merging this to my branch and will update the dependencies there now that [omniauth-openid has been released as v2.0.1](https://github.com/omniauth/omniauth-openid/pull/44#issuecomment-763027213). Thanks again!

PR #5333, opened 2021-01-18
Fix Rails default log level

ghiculescu:
This changed in https://github.com/rails/rails/commit/229fd2a02fc694b4b7756445b6647777aa94e25d - the advice about changing your log level still stands though.


carlosantoniodasilva:
Thanks!

PR #5327, opened 2021-01-07
Improve OmniAuth version check to allow anything from 1.0 forward

carlosantoniodasilva:
This should enable people to try OmniAuth 2 currently in pre-release.

Closes #5326, And I believe also closes #5236. (for now at least)

**TO-DO**

- [x] Update Wiki: https://github.com/heartcombo/devise/wiki/OmniAuth:-Overview to use POST

PR #5337, opened 2021-01-30
Update omniauth constraint

koenpunt:
The error for an old version of omniauth is raised, when in fact the version is too new.

This changes the error message, and adds a version constraint to the gem in the Gemfile for development.


koenpunt:
Nevermind, already proposed in #5327

PR #5341, opened 2021-02-02
Add Truffleruby head to CI

gogainda:


carlosantoniodasilva:
Thank you @gogainda, but like I mentioned in https://github.com/heartcombo/responders/pull/224, for now I'm not invested in maintaining support for other Ruby implementations, and this is across all of the heartcombo projects.

PR #5345, opened 2021-02-15
Fix cookie overflow errors with the return_to paths in session variables

ahukkanen:
When the Rails application is using the cookie store to store the session variables (which is the default), it can overflow in case Devise sets the `user_return_to` path with long URL parameters.

This fixes the issue by stripping out the URL parameters if the path with parameters exceeds 3 KB (cookie overflow happens at 4 KB and there can be other session variables too).

In case some other session store has been configured, this won't have any effect to current functionality.

### DIFF
--- a/lib/devise/controllers/store_location.rb
+++ b/lib/devise/controllers/store_location.rb
@@ -35,7 +35,7 @@ module Devise
       #
       def store_location_for(resource_or_scope, location)
         session_key = stored_location_key_for(resource_or_scope)
-        
+
         path = extract_path_from_location(location)
         session[session_key] = path if path
       end
@@ -56,7 +56,7 @@ module Devise
       def extract_path_from_location(location)
         uri = parse_uri(location)
 
-        if uri 
+        if uri
           path = remove_domain_from_uri(uri)
           path = add_fragment_back_to_path(uri, path)
 
@@ -65,7 +65,18 @@ module Devise
       end
 
       def remove_domain_from_uri(uri)
-        [uri.path.sub(/\A\/+/, '/'), uri.query].compact.join('?')
+        path = [uri.path.sub(/\A\/+/, '/'), uri.query].compact.join('?')
+        return path unless Rails.application.config.session_store == ActionDispatch::Session::CookieStore
+
+        # Allow 3 KB size for the path because there can be also some other
+        # session variables out there (leave 1 KB for the other variables).
+        return path if path.bytesize <= ActionDispatch::Cookies::MAX_COOKIE_SIZE - 1024
+
+        # For too long paths, remove the URL parameters to avoid
+        # ActionDispatch::Cookies::CookieOverflow exception. ActionDispatch
+        # allows a maximum total cookie size of 4 KB because this is the hard
+        # limit that can be handled by all the browsers.
+        path.split("?").first
       end
 
       def add_fragment_back_to_path(uri, path)
--- a/test/integration/authenticatable_test.rb
+++ b/test/integration/authenticatable_test.rb
@@ -313,6 +313,46 @@ class AuthenticationRedirectTest < Devise::IntegrationTest
     assert_nil session[:"user_return_to"]
   end
 
+  test 'short parameters preserved in the return_to session variable with cookie based session store' do
+    get users_path, params: { a: "123", b: "456" }
+    assert_redirected_to new_user_session_path
+    assert_equal "#{users_path}?a=123&b=456", session[:user_return_to]
+  end
+
+  test 'long parameters stripped from the return_to session variable with cookie based session store' do
+    # Generates 4 KB worth of parameters
+    long_params = 4.times.map do |i|
+      ["param#{i}", 'a' * 993] # 6 bytes + 1 byte ('=') + 993 bytes = 1000 bytes each
+    end.to_h
+
+    get users_path, params: long_params
+    assert_redirected_to new_user_session_path
+    assert_equal users_path, session[:user_return_to]
+  end
+
+  test 'long parameters not stripped from the return_to session variable with cache session store' do
+    # Disable the path stripping by stubbing the session store to something else
+    # than the cookie store.
+    Rails.application.config.stubs(:session_store).returns(ActionDispatch::Session::CacheStore)
+
+    # Prevent the cookie overflow error by increasing the MAX_COOKIE_SIZE value
+    orig_size = ActionDispatch::Cookies.send(:remove_const, :MAX_COOKIE_SIZE)
+    ActionDispatch::Cookies.const_set(:MAX_COOKIE_SIZE, 10240)
+
+    # Generates 4 KB worth of parameters
+    long_params = 4.times.map do |i|
+      ["param#{i}", 'a' * 993] # 6 bytes + 1 byte ('=') + 993 bytes = 1000 bytes each
+    end.to_h
+
+    get users_path, params: long_params
+    assert_redirected_to new_user_session_path
+    assert_equal "#{users_path}?#{long_params.to_param}", session[:user_return_to]
+
+    # Set the original value for MAX_COOKIE_SIZE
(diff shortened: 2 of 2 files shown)


phrsouza:
It seems to be related to https://github.com/heartcombo/devise/pull/3347.

ahukkanen:
@phrsouza Yes, you are correct that #3347 addressed the exact same issue.

There are some differences in this PR, however:
- This does not affect the return to path for non-cookie based session stores
- If the path to be stored is too long, this still stores it but it strips the URL parameters off the URL which usually cause the overflow
- This reads the cookie size from `ActionDispatch::Cookies::MAX_COOKIE_SIZE` and leaves 1 KB for other possible session variables. This seemed to be one of the issues mentioned by @danielfone

carlosantoniodasilva (commented):
Thanks for the PR, I'll need to give this some more thought, but I think some of the original concerns persist, in the sense that we'll be swallowing an error and turning it into a possibly unexpected redirect now, in other words, by dropping the query params the redirect may be less relevant for the application. (you could lose things like utm_* attributes and such)

There's also the fact that while this could avoid some of the cookie overflow errors, it will not eliminate all of them. If you have a path that's 3kb but have more than another 1kb in the session, you're still going to hit the upper limit and overflow, causing the error.

I'm not sure we can fix all of these things together in Devise, given that's a broader scenario/problem with max cookie size.

Anyway, I wanted to give you some initial feedback here, I'll circle back on it after giving it better consideration. Thanks again!

ahukkanen:
@carlosantoniodasilva Sure thing, I understand the concerns. I've already fixed the issue for the application but I thought it would be good to fix "upstream" in Devise as well. As seen by the other PR, this is an issue that may concern a broader audience.

I understand your concern regarding the length of the path but I would say it is more common to have the parameters causing the overflow than the path itself. I've never in my life seen URL paths exceeding 3 KB, although it is technically possible. If you feel it should be checked too, of course we can add the check - not a problem. Just saying it is not as common.

Regarding the other concern about losing some URL parameters, understood, but I think the application maintainer would rather have the user to see the page they were trying to request than seeing a HTTP 500 error page. If the bigger concern is that the maintainer is unaware of this situation as it happens "silently", I think a better approach would be to log a message when this happens OR to make it configurable.

The main idea of this PR is not necessarily to handle all the cases related to this but to make the overflow exception less likely to happen.

carlosantoniodasilva:
@kylefox appreciate the feedback. You don't need this to be a monkey patch if you are writing your own failure app, you can simply override `store_location_for` there based on the gist code.

Either way, the original concern from José still applies: https://github.com/heartcombo/devise/pull/3347#issuecomment-64869120... if this is an invalid path for the app than it should probably fail hard, instead of silently pass, in my opinion.

jcoyne:
@carlosantoniodasilva We overrode `store_location_for` (https://github.com/sul-dlss/happy-heron/pull/1919/files)  as @danielfone provided, but we still see this error occurring:

This happens when the client is using the Nuclei vulnerability scanner: github.com/projectdiscovery/nuclei

and the request url looks like this:
```
/?Page=%0D%0ASet-Cookie:crlfinjection=crlfinjection&_url=%0D%0ASet-Cookie:crlfinjection=crlfinjection&callback=%0D%0ASet-Cookie:crlfinjection=crlfinjection&checkout_url=%0D%0ASet-Cookie:crlfinjection=crlfinjection&content=%0D%0ASet-Cookie:crlfinjection=crlfinjection&continue=%0D%0ASet-Cookie:crlfinjection=crlfinjection&continueTo=%0D%0ASet-Cookie:crlfinjection=crlfinjection&counturl=%0D%0ASet-Cookie:crlfinjection=crlfinjection&data=%0D%0ASet-Cookie:crlfinjection=crlfinjection&dest=%0D%0ASet-Cookie:crlfinjection=crlfinjection&dest_url=%0D%0ASet-Cookie:crlfinjection=crlfinjection&dir=%0D%0ASet-Cookie:crlfinjection=crlfinjection&document=%0D%0ASet-Cookie:crlfinjection=crlfinjection&domain=%0D%0ASet-Cookie:crlfinjection=crlfinjection&done=%0D%0ASet-Cookie:crlfinjection=crlfinjection&download=%0D%0ASet-Cookie:crlfinjection=crlfinjection&feed=%0D%0ASet-Cookie:crlfinjection=crlfinjection&file=%0D%0ASet-Cookie:crlfinjection=crlfinjection&host=%0D%0ASet-Cookie:crlfinjection=crlfinjection&html=%0D%0ASet-Cookie:crlfinjection=crlfinjection&http=%0D%0ASet-Cookie:crlfinjection=crlfinjection&https=%0D%0ASet-Cookie:crlfinjection=crlfinjection&image=%0D%0ASet-Cookie:crlfinjection=crlfinjection&image_src=%0D%0ASet-Cookie:crlfinjection=crlfinjection&image_url=%0D%0ASet-Cookie:crlfinjection=crlfinjection&imageurl=%0D%0ASet-Cookie:crlfinjection=crlfinjection&include=%0D%0ASet-Cookie:crlfinjection=crlfinjection&media=%0D%0ASet-Cookie:crlfinjection=crlfinjection&navigation=%0D%0ASet-Cookie:crlfinjection=crlfinjection&next=%0D%0ASet-Cookie:crlfinjection=crlfinjection&open=%0D%0ASet-Cookie:crlfinjection=crlfinjection&out=%0D%0ASet-Cookie:crlfinjection=crlfinjection&page=%0D%0ASet-Cookie:crlfinjection=crlfinjection&page_url=%0D%0ASet-Cookie:crlfinjection=crlfinjection&pageurl=%0D%0ASet-Cookie:crlfinjection=crlfinjection&path=%0D%0ASet-Cookie:crlfinjection=crlfinjection&picture=%0D%0ASet-Cookie:crlfinjection=crlfinjection&port=%0D%0ASet-Cookie:crlfinjection=crlfinjection&proxy=%0D%0ASet-Cookie:crlfinjection=crlfinjection&redir=%0D%0ASet-Cookie:crlfinjection=crlfinjection&redirect=%0D%0ASet-Cookie:crlfinjection=crlfinjection&redirectUri&redirectUrl=%0D%0ASet-Cookie:crlfinjection=crlfinjection&reference=%0D%0ASet-Cookie:crlfinjection=crlfinjection&referrer=%0D%0ASet-Cookie:crlfinjection=crlfinjection&req=%0D%0ASet-Cookie:crlfinjection=crlfinjection&request=%0D%0ASet-Cookie:crlfinjection=crlfinjection&retUrl=%0D%0ASet-Cookie:crlfinjection=crlfinjection&return=%0D%0ASet-Cookie:crlfinjection=crlfinjection&returnTo=%0D%0ASet-Cookie:crlfinjection=crlfinjection&return_path=%0D%0ASet-Cookie:crlfinjection=crlfinjection&return_to=%0D%0ASet-Cookie:crlfinjection=crlfinjection&rurl=%0D%0ASet-Cookie:crlfinjection=crlfinjection&show=%0D%0ASet-Cookie:crlfinjection=crlfinjection&site=%0D%0ASet-Cookie:crlfinjection=crlfinjection&source=%0D%0ASet-Cookie:crlfinjection=crlfinjection&src=%0D%0ASet-Cookie:crlfinjection=crlfinjection&target=%0D%0ASet-Cookie:crlfinjection=crlfinjection&to=%0D%0ASet-Cookie:crlfinjection=crlfinjection&uri=%0D%0ASet-Cookie:crlfinjection=crlfinjection&url=%0D%0ASet-Cookie:crlfinjection=crlfinjection&val=%0D%0ASet-Cookie:crlfinjection=crlfinjection&validate=%0D%0ASet-Cookie:crlfinjection=crlfinjection&view=%0D%0ASet-Cookie:crlfinjection=crlfinjection&window=%0D%0ASet-Cookie:crlfinjection=crlfinjection&redirect_to=%0D%0ASet-Cookie:crlfinjection=crlfinjection
```

PR #5347, opened 2021-02-16
Support Ruby 3+ officially, remove final Rails 6.1 deprecations, drop test dependency for XML serialization

carlosantoniodasilva:
Adding Ruby 3 to the matrix caused some issues with Rails 6.0 to show up, which required us to point to the 6-0-stable branch for now. We'll switch back to a stable release version once it gets out there.

And in order to remove Rails 6.1 deprecation with `ActiveMode::Errors#to_xml`, I went ahead and changes the tests that relied on XML serialization to use JSON instead, so we would no longer have that deprecation and could remove the dependency on the extracted `activemodel-serializers-xml`, simplifying things. Note that this doesn't mean XML isn't supported by Devise, it simply means we're using JSON to test non-navigatable (HTML & such) flows.


waissbluth:
thanks @carlosantoniodasilva. is there anything else preventing the rails 6.1 support to make it into a release?

carlosantoniodasilva:
@waissbluth kinda, I was hoping to prepare a few other things since the OmniAuth change that is on master may be considered a bit of a breaking change, but I might go ahead and release what's in there as 4.8 instead. (I'm not ready for a major bump.) Can't give you an ETA for now though. Thanks!

carlosantoniodasilva:
@Dreamersoul appreciate the feedback.

I don't think any changes are required to run with Rails 6.1 right now, which means the current stable version should work. This just helped ensure our test suite was green with that version of Rails.

That being said, the changes for omniauth aren't as breaking to require a 5.0 release (and all the work involved in reverting those from master to exclude from a next release, etc.), I just wanted to wrap up a few other things to be able to release something like a 4.8 including those changes. Anyway, I'm hoping to cut one soon including this (master) even if I don't get to include everything I wanted.

carlosantoniodasilva:
I just pushed v4.8 including this change, please give it a try and report back if you have any issues. Thanks.

PR #5349, opened 2021-02-20
Cleaned up notices in locales/en.yml

dcurtis:
1) Improved grammar.
2) Made some messages more neutral by default.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -4,8 +4,8 @@ en:
   devise:
     confirmations:
       confirmed: "Your email address has been successfully confirmed."
-      send_instructions: "You will receive an email with instructions for how to confirm your email address in a few minutes."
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva (commented):
Thanks for your contribution, I like the copy updates you're proposing here.

I'll have to see which version to include them, and we'll have to fix the tests, please take a look at those if you can, otherwise I'll circle back on these and get everything cleaned up & merged.

dcurtis:
> I'll have to see which version to include them, and we'll have to fix the tests, please take a look at those if you can, otherwise I'll circle back on these and get everything cleaned up & merged.

Okay, I further cleaned up the notices for paranoid mode and updated the tests so that they pass. I believe it can be safely merged now.

PR #5356, opened 2021-03-09
Change in README

bibinvenugopal:
Changed `rails db:migrate` to `rake db:migrate`


carlosantoniodasilva:
Since Rails 5, all `rake` tasks are callable through `rails` so that you can use `rails` for everything, e.g. `rails db:migrate`.

https://github.com/rails/rails/pull/22288#issuecomment-164296649

Devise still supports Rails 4, but since this should change soon I'll leave the Readme as is.

Thanks for your contribution.

PR #5366, opened 2021-03-20
Trackable: save with touch: false

jscissr:
With this, the `updated_at` timestamp won't be updated. We already have a tracking timestamp in `current_sign_in_at`, it doesn't make sense to also update `updated_at`.

See https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save

### DIFF
--- a/lib/devise/models/trackable.rb
+++ b/lib/devise/models/trackable.rb
@@ -37,7 +37,7 @@ module Devise
         return if new_record?
 
         update_tracked_fields(request)
-        save(validate: false)
+        save(validate: false, touch: false)
       end
 
       protected


carlosantoniodasilva:
The `updated_at` column is usually bumped with any updates to the record, no matter what updates are being done, part of the default lifecycle of the record. Skipping it in Devise might break that expectation in ways we can't foresee, for example if someone is relying on that timestamp as a cache key in a view fragment, and would like to re-generate the cache, we'd possibly break that behavior by not bumping `updated_at` here.

Just skipping that bumping sounds riskier than leaving it as is, since there's basically no penalty with bumping it (that I can think of right now), so I'll keep it. Thanks for your contribution.

Issue #5133, opened 2019-09-04
Updating User model with GET requests is not compatible with Rails 6 read/write splitting


vitobotta:
## Environment

- Ruby 2.6.4
- Rails 6.0.0
- Devise 4.7.0

## Current behavior

Not sure of where to post this. I am trying to set up read/writing splitting of SQL queries with Rails 6 since it now has support for this thanks to the multiple database features, and am having problems with some Devise actions. Rails' new auto connection switching sends writes to a primary database and reads to replicas and whether or not writes are allowed depends on the HTTP request. If it's a GET request for example, only reads are allowed by default and sent to the replicas, while for POST/PUT/DELETE queries are sent to the master/primary.

In some cases Devise performs writes even with GET requests, for example when user follows the link in the confirmation email which confirms the user using a token. In this case Devise updates `confirmed_at` and `updated_at` columns of the user but because it's a GET request, the write is blocked by Rails 6 using the automatic connection switching based on the HTTP verb.

## Expected behavior

In theory Devise should follow HTTP semantics and not make changes to data with GET requests, but I am not sure of how to work around this. I am thinking that I could load the confirmation page and then submit a POST request with JavaScript that actually confirms the user, but there are other cases where Devise makes changes with GET requests. 

Any help/idea much appreciated.

colinross:
I don't necessarily want to get into a discussion about the rails 6 read/write stuff, but I don't (personally) think that is it that black and white, especially when dealing with the meta-record like the logged in user, assuming you want to track their actions. I would also argue that we are talking about RESTful semantics, and not HTTP. 

Specifically with the confirm request, you can't do a POST from inside an email, at least directly. Also, semantically, it _is_ the show action of the 'confirmation' resource.

That all said, I believe the cleanest way of to overcome this situation would be to always use the write db for devise actions, or maybe just the ones you know have problems. Again, I'm not sure of how granular you can configure the 

Alternatively you could override the pertinent methods on the model
```
# User
def confirm
  User.connected_to(hander: :writing) do
    super
  end
end
```

I don't think this is a bug with devise, at a certain point there is a difference between 'written for rails' and 'written to support every possible configuration of rails'.

It is clearly a personal opinion, but I would prefer devise to be a solid, tested, and customizable 90% solution rather than muck it up with large amounts of functionality not required by 95+% of installations.

vitobotta:
Hi @colinross , in the meantime I ended up doing something similar with an around filter for all the Devise actions just to be sure:

[...]

PR #5369, opened 2021-03-22
Create a model hook around the lockable warden hook to reset attempts

carlosantoniodasilva:
Resetting failed attempts after sign in happened inside a warden hook
specific for the lockable module, but that was hidden inside the hook
implementation and didn't allow any user customization.

One such customization needed for example is to direct these updates to
a write DB when using a multi-DB setup. With the logic hidden in the
warden hook this wasn't possible, now that it's exposed in a model
method much like trackable, we can override the model method to wrap it
in a connection switch block for example, point to a write DB, and
simply call `super`.

Closes #5310
Related to #5264 and #5133


geoffharcourt (approved):
Awesome. Thank you for taking this on, my apologies for not getting to it sooner! I think this change will help a lot of Rails 6 users with a replica database setup.

mhenrixon:
This looks great and not as complicated as I thought it would be. Kudos for tackling this problem! It is the one thing that prevents me from making full (and easy) use of the replica reads.

geoffharcourt:
We've been using this change in production and haven't seen any exceptions from attempting to write to the read-only replica since applying it.

carlosantoniodasilva:
Thanks @geoffharcourt @mhenrixon, this is now on master and will be on the next release, which I'm hoping to cut soon.

paulspringett:
Hey @carlosantoniodasilva Please could I ask when you're planning on the next release to RubyGems? I've just come across this as we're looking for a fix to this issue with Rails 6.1 multi-database switching. Many thanks!

carlosantoniodasilva:
I just pushed v4.8 including this change, please give it a try and report back if you have any issues. Thanks.

paulspringett:
@carlosantoniodasilva Thanks so much, really appreciate it 🙌

PR #5378, opened 2021-05-06
Add metadata for RubyGems

nickhammond:
Add the various links to RubyGems so people can go directly to various places such as the changelog when upgrading.

### DIFF
--- a/devise.gemspec
+++ b/devise.gemspec
@@ -14,6 +14,14 @@ Gem::Specification.new do |s|
   s.homepage    = "https://github.com/heartcombo/devise"
   s.description = "Flexible authentication solution for Rails with Warden"
   s.authors     = ['José Valim', 'Carlos Antônio']
+  s.metadata    = {
+    "homepage_uri"      => "https://github.com/heartcombo/devise",
+    "documentation_uri" => "https://rubydoc.info/github/heartcombo/devise",
+    "changelog_uri"     => "https://github.com/heartcombo/devise/blob/master/CHANGELOG.md",
+    "source_code_uri"   => "https://github.com/heartcombo/devise",
+    "bug_tracker_uri"   => "https://github.com/heartcombo/devise/issues",
+    "wiki_uri"          => "https://github.com/heartcombo/devise/wiki"
+  }
 
   s.files         = Dir["{app,config,lib}/**/*", "CHANGELOG.md", "MIT-LICENSE", "README.md"]
   s.require_paths = ["lib"]


devise.gemspec
```diff
@@ -14,6 +14,13 @@ Gem::Specification.new do |s|
   s.homepage    = "https://github.com/heartcombo/devise"
   s.description = "Flexible authentication solution for Rails with Warden"
   s.authors     = ['José Valim', 'Carlos Antônio']
+  s.metadata    = {
+    "homepage_uri"     => "https://github.com/heartcombo/devise",
+    "changelog_uri"    => "https://github.com/heartcombo/devise/blob/master/CHANGELOG.md",
+    "source_code_uri"  => "https://github.com/heartcombo/devise/issues",
```
carlosantoniodasilva, line 20: This should be either: 

```suggestion
    "source_code_uri"  => "https://github.com/heartcombo/devise",
```

Or:

```suggestion
    "bug_tracker_uri"  => "https://github.com/heartcombo/devise/issues",
```

devise.gemspec
```diff
@@ -14,6 +14,13 @@ Gem::Specification.new do |s|
...
+    "wiki_uri"         => "https://github.com/heartcombo/devise/wiki",
+    "mailing_list_uri" => "http://stackoverflow.com/questions/tagged/devise"
```
carlosantoniodasilva, line 22: We actually have a mailing list under heartcombo: https://groups.google.com/g/heartcombo... I'm not sure I want to point people to there though as I have been thinking about killing it, but maybe it's a better option to start. I'm also not sure about pointing to stackoverflow here, but I guess that's where we point people from the readme anyway... just seems a little odd to have it here for some reason 😅

carlosantoniodasilva (commented):
Hey, thanks, I've had that noted down somewhere to review at some point. I had a couple of comments, mainly the source code vs bug tracker URI (and we may need both?), let me know what you think.

devise.gemspec
nickhammond: Ah, yah I was just grabbing it from the readme but definitely up to the core contributors. I think the Google Groups mailing list must be private as it redirects me to https://groups.google.com/access-error?continue=https://groups.google.com/g/heartcombo. We could of course just remove it for now, I was just trying to enhance the metadata with [what's available](https://guides.rubygems.org/specification-reference/#metadata).

nickhammond:
@carlosantoniodasilva I think we want both `source_code_uri` and `bug_tracker_uri`, I tried to batch commit both of your suggestions but had to manually do the other one. I've also added the `documentation_uri` to point to https://rubydoc.info/github/heartcombo/devise. 

Thanks for making such a great gem for the community over all these years!

carlosantoniodasilva:
Oh right, I think google groups now require you to be logged in to google or something... anyway let's not worry about it, like I said I'm probably gonna kill it at some point, maybe come up with something else instead.

And no worries about the batch commit, if you can't squash everything I can do it when merging. I'll take another look later/tomorrow. Thanks again!

nickhammond:
`mailing_list_uri` removed and commits squashed!

carlosantoniodasilva:
Perfect, thank you!

PR #5382, opened 2021-06-10
Mirror rails default_sender logic

jmschneider:
Update `default_sender` logic to mirror the [Rails `compute_default` method.](https://github.com/rails/rails/blob/9e66e5d027f32b2dc3b695471e9b99f214af3f0f/actionmailer/lib/action_mailer/base.rb#L913-L921) 

The existing code uses `instance_eval` which always passes `self` in as the first argument. The examples in the Rails guides and API docs use lambdas with no arguments. With the existing Devise code this raises an `ArgumentError` exception since lambda expects no arguments and one is passed.

### DIFF
--- a/lib/devise/mailers/helpers.rb
+++ b/lib/devise/mailers/helpers.rb
@@ -53,7 +53,13 @@ module Devise
       def mailer_sender(mapping, sender = :from)
         default_sender = default_params[sender]
         if default_sender.present?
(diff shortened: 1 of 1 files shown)


jmschneider:
Are there any roadblocks for merging this? It would be nice if devise were compatible with the current rails docs.

dilshod:
we need this fix, experiencing the same issue. I'll have to create another mailer class as a workaround

dilshod:
@jmschneider yes, that works, thanks

PR #5386, opened 2021-06-26
Fix code block in doc for after_sign_in_path_for

kelvich:


lib/devise/controllers/helpers.rb
```diff
@@ -210,6 +210,7 @@ def signed_in_root_path(resource_or_scope)
       #       else
       #         super
       #       end
+      #     end
```
nashby, line 213: there's no need for that `end`. We need one `end` for `def`  and one for `if`

kelvich: Ah, I confused `||` with the beginning of code block. Sorry for the noise.

PR #5392, opened 2021-07-29
per-Controller auth strategies

braindeaf:
Adding BlahController.strategies config to allow only specific  strategies to be run on a per-controller basis. The best example of this is we have at least 5 strategies for our User class, but when it comes to our API controllers we only want :jwt to run, ignoring database_authenticatable, yubi_key, etc.

This was my initial solution, before this pull request.

https://robl.me/posts/the-magical-devise-journey

### DIFF
--- a/lib/devise/controllers/helpers.rb
+++ b/lib/devise/controllers/helpers.rb
@@ -9,6 +9,15 @@ module Devise
       include Devise::Controllers::StoreLocation
 
       included do
+        # inside any controller the strategies that are consulted can be configured
+        # such that only whitelisted strategies are run. 
+        # Example:
+        #   self.strategies = :authenticatable
+        #   self.strategies = [:authenticatable, :database_authenticatable]
+        #   self.strategies = { user: :authenticatable, admin: :database_authenticatable }
+        #
+        class_attribute :strategies
+
         if respond_to?(:helper_method)
           helper_method :warden, :signed_in?, :devise_controller?
         end
@@ -46,14 +55,14 @@ module Devise
                 mappings.unshift mappings.delete(favorite.to_sym) if favorite
                 mappings.each do |mapping|
                   opts[:scope] = mapping
-                  warden.authenticate!(opts) if !devise_controller? || opts.delete(:force)
+                  warden.authenticate!(*self.class.strategies_for(mapping), opts) if !devise_controller? || opts.delete(:force)
                 end
               end
             end
 
             def #{group_name}_signed_in?
               #{mappings}.any? do |mapping|
-                warden.authenticate?(scope: mapping)
+                warden.authenticate?(*self.class.strategies_for(mapping), scope: mapping)
               end
             end
 
@@ -61,7 +70,7 @@ module Devise
               mappings = #{mappings}
               mappings.unshift mappings.delete(favorite.to_sym) if favorite
               mappings.each do |mapping|
-                current = warden.authenticate(scope: mapping)
+                current = warden.authenticate(*self.class.strategies_for(mapping), scope: mapping)
                 return current if current
               end
               nil
@@ -69,7 +78,7 @@ module Devise
 
             def current_#{group_name.to_s.pluralize}
               #{mappings}.map do |mapping|
-                warden.authenticate(scope: mapping)
+                warden.authenticate(*self.class.strategies_for(mapping), scope: mapping)
               end.compact
             end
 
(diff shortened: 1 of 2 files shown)


xijo:
I'd like to bump this PR. 

We have the same use case: API should only use token authentication, rest of the app tries all approaches. And I'd guess we're not be the only ones..

Anything speaking against integrating it into devise? Happy to help with the PR in case anything is still missing.

braindeaf:
We are using this PR in production so it's been really invaluable.

lib/devise/controllers/helpers.rb
```diff
@@ -9,6 +9,15 @@ module Helpers
...
 
       included do
+        # inside any controller the strategies that are consulted can be configured
+        # such that only whitelisted strategies are run. 
+        # Example:
+        #   self.strategies = :authenticatable
+        #   self.strategies = [:authenticatable, :database_authenticatable]
+        #   self.strategies = { user: :authenticatable, admin: :database_authenticatable }
+        #
+        class_attribute :strategies
```
oehlschl, line 19: Thoughts on making this a controller instance method as opposed to a class attribute? Agreed that this would be useful -- I'm in the middle of implementing something similar for a multi-tenant app with a single scope but a few separate auth mechanisms -- but I think it could be helpful to have the request context available if possible.

I realize Warden encourages you to locate strategy eligibility in the strategy itself, but if we did want to pick explicitly in the controller based on the request or some runtime information, that would be possible with an instance method but not a class attribute.

Making this an instance method would also make this function a little closer to the way the `auth_options` work in `Devise::SessionsController#create.`: https://github.com/heartcombo/devise/blob/main/app/controllers/devise/sessions_controller.rb#L47

lib/devise/controllers/helpers.rb
```diff
@@ -46,30 +55,30 @@ def authenticate_#{group_name}!(favorite = nil, opts = {})
                 mappings.unshift mappings.delete(favorite.to_sym) if favorite
                 mappings.each do |mapping|
                   opts[:scope] = mapping
-                  warden.authenticate!(opts) if !devise_controller? || opts.delete(:force)
+                  warden.authenticate!(*self.class.strategies_for(mapping), opts) if !devise_controller? || opts.delete(:force)
```
oehlschl, line 58: You'll likely also want to respect any per-controller strategies here (and add `self.class.strategies_for` to this `.authenticate` call):

https://github.com/heartcombo/devise/blob/main/app/controllers/devise/sessions_controller.rb#L19

oehlschl: I guess that `Devise::SessionsController` probably doesn't include these controller helpers, but as part of this work, it may be worth considering how to allow custom strategies in that controller. (It would be a mildly more complete solution, even if it's just adding an empty `auth_strategies` or `strategies` method to that controller and respecting it in the `authenticate` call.)

lib/devise/controllers/helpers.rb
braindeaf: That seems perfectly reasonable. I'll take a look.

oehlschl:
Just for the record, another solution I've seen encouraged for this problem is to set something in the Rails/request env in the controller and check it in the Strategy. Ex:

```ruby
class Api::Users::SessionsController
  prepend_before_action :skip_token_strategy!, only: :create
  
  private def skip_token_strategy!
    request.env['devise.skip_token_strategy'] = true
  end
end

class Devise::Strategies::Token
  def valid?
    !env['devise.skip_token_strategy']
  end
end
```

Not suggesting anything one way or the other; just anticipating what other responses to this proposal might be. (Ex: "This can already be done, and strategy eligibility / validity should be determined in the Strategy itself.")

braindeaf:
Not sure how I feel about that solution. A Strategy should be able to understand what is valid or not sure. Say, if it's a JWT token request and maybe forcing the request to be JSON. That I could understand. But deciding on whether to force the request to be invalid for some other arbitrary reason seems unreasonable to inject a reason not to cause the Strategy to fail, not to mention that it will run every Strategy which is unnecessary if you know you only want to run one of them.

Not sure what the hell has happened after my merge of upstream, now I have 23 file changes :S

lib/devise/controllers/helpers.rb
braindeaf: changed on your recommendation

lib/devise/controllers/helpers.rb
braindeaf: it is now possible to override the SessionsController too :)

PR #5394, opened 2021-08-09
Redirect if reset password token is expired

c960657:
If the user loads the password reset page with an expired or invalid password token, there is no point in showing the password+password_confirmation fields. The user would most likely want to request a new token, so redirect to the page for requesting a new token.


c960657:
Tests are currently failing due to #5745.

PR #5410, opened 2021-10-23
Return 303 on redirect for Turbo

coder2000:
Turbo uses the `DELETE` verb directly so when redirecting with a 302 it retains that verb and the routing fails. This change will use status 303 which will force a change to `GET` which will allow proper redirecting after sign out.

More information can be found in the following issues:
https://github.com/hotwired/turbo/issues/84
https://github.com/hotwired/turbo-rails/issues/259
https://github.com/rails/rails/pull/43430

### DIFF
--- a/app/controllers/devise/sessions_controller.rb
+++ b/app/controllers/devise/sessions_controller.rb
@@ -77,7 +77,7 @@ class Devise::SessionsController < DeviseController
     # support returning empty response on GET request
     respond_to do |format|
       format.all { head :no_content }
-      format.any(*navigational_formats) { redirect_to after_sign_out_path_for(resource_name) }
+      format.any(*navigational_formats) { redirect_to after_sign_out_path_for(resource_name), status: 303 }
     end
   end
 end


JunichiIto:
This PR helped my issue. Thank you very much.

However, I had to add `notice:` option in order to display signed-out message:

```ruby
  def respond_to_on_destroy
    # We actually need to hardcode this as Rails default responder doesn't
    # support returning empty response on GET request
    respond_to do |format|
      format.all { head :no_content }
      format.any(*navigational_formats) { redirect_to after_sign_out_path_for(resource_name), status: 303, notice: find_message(:signed_out) }
    end
  end
```

And the same issue can happen `registrations#destroy` action. I override destroy action in Devise::RegistrationsController too:

```ruby
  def destroy
    resource.destroy
    Devise.sign_out_all_scopes ? sign_out : sign_out(resource_name)
    set_flash_message! :notice, :destroyed
    yield resource if block_given?
    respond_with_navigational(resource){ redirect_to after_sign_out_path_for(resource_name), status: 303, notice: find_message(:destroyed) }
  end
```

Genkilabs:
I think this is still relevant, as seen here https://github.com/heartcombo/devise/issues/5458

carlosantoniodasilva:

PR #5415, opened 2021-10-30
Add description after adding `before_action :authenticate_user!`.

R-Tsukada:
Add description of the transition after adding `before_action:authenticate_user!`
https://github.com/heartcombo/devise#controller-filters-and-helpers


yuseitahara:
Since it was not that clear for beginners about the effect of adding `before_action :authenticate_user!`, thus this additional sentence should be useful.

PR #5416, opened 2021-11-01
Make copying and pasting terminal commands easier

DerekCrosson:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] QA Issue Fix
- [ ] Optimization
- [x] Documentation Update

## Description
Removes the dollar sign from the beginning of terminal commands.

### Problem
When the commands that have the terminal beginning of line symbol in front of them are copied to the clipboard using the copy button in the code block, the user has to first remove the symbol to run the command. Running the command without first removing this symbol will result in an error.

### Solution
Remove the terminal beginning of line symbol so that the user can click the button in the code block to copy to clipboard and then simply past the command in the terminal and press the return key to run it.

## Related Issues, Tickets & Documents
N/A

### UI accessibility concerns?
N/A

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: documentation update, tests not necessary.
- [ ] I need help with writing tests

## [optional] Are there any post deployment tasks we need to perform?
No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ Then run `bundle install`
 Next, you need to run the generator:
 
 ```console
-$ rails generate devise:install
+rails generate devise:install
 ```
 
 At this point, a number of instructions will appear in the console. Among these instructions, you'll need to set up the default URL options for the Devise mailer in each environment. Here is a possible configuration for `config/environments/development.rb`:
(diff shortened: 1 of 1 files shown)


p8:
With github's copy button this does seem like a nice change.
It seems most Ruby projects still have the prompt.
The only example I could find of removing it for easy copying, is the Rust README: https://github.com/rust-lang/rust/issues/76863

mcmuralishclint:
Good one

PR #5426, opened 2021-11-29
Improved failure_message to interpret OAuth2::Error

coezbek:
OAuth2::Error Exceptions do not have `error_reason` and `error` as the Omniauth::CallbackError, but rather use the fields `description` and `code`. 

This is particularly useful if the error is returned server-side and handled by OAuth2.

See: `callback_phase` can return `OAuth2::Error` or `CallbackError`
https://github.com/omniauth/omniauth-oauth2/blob/561ddcef4d7791bc46ff1d718de5db13894d8033/lib/omniauth/strategies/oauth2.rb#L84

See: Definition of `CallbackError`
https://github.com/omniauth/omniauth-oauth2/blob/561ddcef4d7791bc46ff1d718de5db13894d8033/lib/omniauth/strategies/oauth2.rb#L149

See: Definition of `OAuth2::Error`
https://github.com/oauth-xx/oauth2/blob/1940e9ad6f395b2fb80f09e4140d17fe0344325a/lib/oauth2/error.rb#L4

### DIFF
--- a/app/controllers/devise/omniauth_callbacks_controller.rb
+++ b/app/controllers/devise/omniauth_callbacks_controller.rb
@@ -21,7 +21,11 @@ class Devise::OmniauthCallbacksController < DeviseController
   def failure_message
     exception = request.respond_to?(:get_header) ? request.get_header("omniauth.error") : request.env["omniauth.error"]
     error   = exception.error_reason if exception.respond_to?(:error_reason)
+    error ||= exception.description  if exception.respond_to?(:description)
+    return error.to_s if error # Description and Error_reason are normal text    
+    
     error ||= exception.error        if exception.respond_to?(:error)
+    error ||= exception.code         if exception.respond_to?(:code)
     error ||= (request.respond_to?(:get_header) ? request.get_header("omniauth.error.type") : request.env["omniauth.error.type"]).to_s
     error.to_s.humanize if error
   end


thatandyrose:
hi @coezbek I'm just looking at this part of the code now, and I've noticed that in certain situations an actual exception object is returned whcih doesn't have reason, error, code or description BUT does have `.message`, as in `error.message`.

app/controllers/devise/omniauth_callbacks_controller.rb
```diff
@@ -21,7 +21,11 @@ def failed_strategy
   def failure_message
     exception = request.respond_to?(:get_header) ? request.get_header("omniauth.error") : request.env["omniauth.error"]
     error   = exception.error_reason if exception.respond_to?(:error_reason)
+    error ||= exception.description  if exception.respond_to?(:description)
+    return error.to_s if error # Description and Error_reason are normal text
```
carlosantoniodasilva, line 25: This might change behavior slightly in that we were humanizing the error reason before... I'd prefer if we just let all messages read that I think.

carlosantoniodasilva (commented):
@coezbek hey, catching up here :)

Are you still interested in this change? If so, would you mind adding tests to it? We have a couple already: https://github.com/heartcombo/devise/blob/90f46bac373416d6ad4b3541e6ca144b1f7ee527/test/integration/omniauthable_test.rb#L144-L159

PR #5429, opened 2021-12-10
Delegate sign_in_after_reset_password to resource class

mattr:
Allows resource class scopes to overrides the global configuration for sign in after reset password behaviour.


carlosantoniodasilva:
Thanks!

PR #5440, opened 2021-12-18
Add :turbo_stream as a default navigational format

### LINKED ISSUE #5439: Devise 4.8.1 + Rails 7.0.0 | Undefined method 'user_url'
## Pre-check

- Created a new rails-app with `rails new app-name -a propshaft -j importmap -c tailwind`
- Added `gem 'devise', '~> 4.8', '>= 4.8.1'` to `Gemfile` and followed installation instructions

## Environment

- Ruby **3.0.3p157**
- Rails **7.0.0**
- Devise **4.8.1**

## Current behavior

- Receiving `NoMethodError (undefined method 'user_url' for #<Devise::RegistrationsController:0x0000000000d638>)` upon successful user registration. 
- User gets persisted in database

Demo app to reproduce the error is available [here](https://github.com/erik-brueggemann/devise-issue-app)

## Expected behavior
- Redirect to `root_path` without error

nickrivadeneira:
Fixes #5439

Rails 7 returns `:turbo_stream` as a request format out-of-the-box. I'm on the fence whether this is the right change to make or if it should be left up to the user to add it as a configuration in their initializer. I opted to open the PR since it is a format that a vanilla Rails 7 app will use, which means vanilla Devise will break with vanilla Rails without this change.

Feel free to close if it's best left as-is.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -217,7 +217,7 @@ module Devise
 
   # Which formats should be treated as navigational.
   mattr_accessor :navigational_formats
-  @@navigational_formats = ["*/*", :html]
+  @@navigational_formats = ["*/*", :html, :turbo_stream]
 
   # When set to true, signing out a user signs out all other scopes.
   mattr_accessor :sign_out_all_scopes


jackparsons93:
I tried your version of devise gem and I am still getting an error on user registration 
devise

I think it would be best to solve this problem by adding to lib/generators/templates/devise.rb. 

config.navigational_formats = ['*/*', :html, :turbo_stream]

jackparsons93:
check out this commit I made [https://github.com/heartcombo/devise/compare/main...jackparsons93:devise:main](url)

I tested it and it works with both deleting users and fixes the sign in issue

carlosantoniodasilva:

PR #5450, opened 2022-01-07
Add Ruby 3.1 to CI matrix

petergoldstein:
As part of enabling Ruby 3.1 and adding it to the CI matrix this PR:

- Updates the CI file so that '3.0' is a string, and thus runs Ruby 3.0.x for those test entries exclusively.
- Locks Nokogiri to versions less than 1.13 because of a webrat incompatibility with Nokogiri 1.13 and higher.
- Updates the root Gemfile to use Rails 7.0.1 (which is Ruby 3.1 compatible)
- Adds conditional inclusion of the net-imap, net-smtp, and net-pop gems in the Rails 6.1 Gemfile when the Ruby version is 3.1.0 or higher
- Assorted other Gem updates in the base Gemfile


petergoldstein:
Note that webrat is seemingly unmaintained (last commit was almost 8 years ago) so it would likely be desirable to remove it as a dependency going forward.  For the moment, locking the Nokogiri version makes everything run green.

.github/workflows/test.yml
```diff
@@ -94,16 +95,32 @@ jobs:
...
+          - ruby: 3.1
+            gemfile: gemfiles/Gemfile-rails-4-2
+          - ruby: 3.1
+            gemfile: gemfiles/Gemfile-rails-5-0
+          - ruby: 3.1
+            gemfile: gemfiles/Gemfile-rails-5-1
+          - ruby: 3.1
+            gemfile: gemfiles/Gemfile-rails-5-2
+          - ruby: 3.1
+            gemfile: gemfiles/Gemfile-rails-6-0
```
aried3r, line 123: This should work if you add `gem "psych", "~> 3.0"` to the appraisal I believe.
https://gist.github.com/yahonda/2776d8d7b6ea7045359f38c10449937b#rails-60z

petergoldstein: Is it supposed to work?  My understanding of the Rails support matrix is that Rails 6.0 fell out of general support (it's on security only) before Ruby 3.1 was released.  So I think that would be an unsupported configuration.

aried3r: Neither is 6.1 officially supported (so far), according to https://github.com/rails/rails/issues/44090.

yahonda is a Rails contributor so I followed their advice for [rpush](https://github.com/rpush/rpush/commit/e628102667a3d101f33c8e26f61c5b900e12f095#diff-635b20966b1d2fea76ee0565254eab939d55f6d22e3c4727cf9d453f589b87a0R20-R22) but you're correct that this might work, but is not guaranteed to work.

Just thought I'd mention it for people who run that kind of setup.

petergoldstein:
Are there any changes that would be helpful in getting this merged?

carlosantoniodasilva:
@petergoldstein thanks, I'm looking into it as well.

Do you have any reference on the Nokogiri + webrat problem, or just the tests here that failed with a newer Nokogiri version?

petergoldstein:
@carlosantoniodasilva When the Nokogiri restriction isn't added, we get failures in the specs with stack traces like:

```
Error:
DeviseHelperTest#test_test_errors.messages.not_saved_with_single_error_from_i18n:
ArgumentError: wrong number of arguments (given 0, expected 2)
    nokogiri-1.13.3-x86_64 (darwin) lib/nokogiri/css/node.rb:26:in `to_xpath'
    webrat (0.7.3) lib/webrat/core/matchers/have_selector.rb:40:in `block in query'
 ```

You can see the relevant lines here:

[Webrat](https://github.com/brynary/webrat/blob/8947ea89c16a4ffef06ccbb871e3452fa3aeb169/lib/webrat/core/matchers/have_selector.rb#L40)[
Nokogiri](https://github.com/sparklemotion/nokogiri/blob/dcccf7206b38daeb7ab60e290f3c1a618a17d982/lib/nokogiri/css/node.rb#L26)

This [commit made ~2 months ago](https://github.com/sparklemotion/nokogiri/commit/b24efcca08cc93881c29c95a31c2068002b591ba) removed the default arguments from `to_xpath` and introduced the incompatibility.

petergoldstein:
I've opened an issue against nokogiri [here](https://github.com/sparklemotion/nokogiri/issues/2469).

carlosantoniodasilva (approved):
@petergoldstein thanks for the details, that makes sense. This looks good to me.

I will push a follow up change reorganizing the test yaml a bit.

PR #5453, opened 2022-01-12
Better handle redirection on sign out

spaquet:
Issue: When a user signs out from a protected page (aka a page only accessible when signed in), Rails 7 (tested on 7.0.1) does not properly redirect to the root_path or another page specified by the user in the `after_sign_out_path_for` method as it uses a 302 status.

Changing to a status 303 enable proper redirection to the root page (or any other page assigned using `after_sign_out_path_for`).


spaquet:
Looks like a duplicate of #5410

PR #5340, opened 2021-02-02
Return 422 status when login fails + treat `turbo_stream` as a navigational format

ghiculescu:
Devise `FailureApp` now returns a `422` status when running the `recall` flow. This, combined with https://github.com/heartcombo/responders/pull/223, enables Devise to work correctly with the [new Rails defaults](https://github.com/rails/rails/pull/41026) for form errors, as well as with new libraries like Turbo.

By default, the `recall` flow only runs on the [`SessionsController`](https://github.com/heartcombo/devise/blob/v4.7.3/app/controllers/devise/sessions_controller.rb#L48).

`:turbo_stream` is also treated as a navigational format which means Devise will return redirects where appropriate rather than `401`s for Turbo requests.

Fixes https://github.com/heartcombo/devise/pull/5325

PR #5467, opened 2022-02-09
turbo_stream fix bugts for rails 7 and above apps

Salanoid:
I tried to fix this two issues: https://github.com/heartcombo/devise/issues/5448, https://github.com/heartcombo/devise/issues/5446.

### DIFF
--- a/lib/generators/templates/devise_turbo.rb
+++ b/lib/generators/templates/devise_turbo.rb
@@ -1,0 +1,311 @@
+# frozen_string_literal: true
+
+# Assuming you have not yet modified this file, each configuration option below
+# is set to its default value. Note that some are commented out while others
+# are not: uncommented lines are intended to protect your configuration from
+# breaking changes in upgrades (i.e., in the event that future versions of
+# Devise change the default values for those options).
+#
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
+Devise.setup do |config|
+  # The secret key used by Devise. Devise uses this key to generate
+  # random tokens. Changing this key will render invalid all existing
+  # confirmation, reset password and unlock tokens in the database.
+  # Devise will use the `secret_key_base` as its `secret_key`
+  # by default. You can change it below and use your own secret key.
+  # config.secret_key = '<%= SecureRandom.hex(64) %>'
+
+  # ==> Controller configuration
+  # Configure the parent class to the devise controllers.
+  # config.parent_controller = 'DeviseController'
+
+  # ==> Mailer Configuration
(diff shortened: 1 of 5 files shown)


clemensg:
@carlosantoniodasilva Would be great to get this or something similar merged

lib/generators/templates/devise_turbo.rb
```diff
@@ -0,0 +1,311 @@
...
+  # When using OmniAuth, Devise cannot automatically set OmniAuth path,
+  # so you need to do it manually. For the users scope, it would be:
+  # config.omniauth_path_prefix = '/my_engine/users/auth'
+
+  # ==> Turbolinks configuration
+  # If your app is using Turbolinks, Turbolinks::Controller needs to be included to make redirection work correctly:
+  #
+  # ActiveSupport.on_load(:devise_failure_app) do
+  #   include Turbolinks::Controller
+  # end
```
sedubois, line 304: I suppose this part can be omitted as Turbolinks is deprecated and superseded by Turbo, where there is no such `Turbolinks::Controller` and such an include appears unnecessary.

Also see https://github.com/heartcombo/devise/pull/5324

tmaier:
PR #5340 looks like it aims for a deeper integration into devise than what is proposed in this pull request.

Besides this, using turbo is also valid in a rails 6.1 context. So just checking for "> Rails 7" would exclude some valid use cases

msducheminjr:
@tmaier Would something along the lines of a `class_option` boolean on the generator allow for less surprising behavior than checking for a Rails version?

```ruby
  class_option :turbo, type: :boolean, default: false
``` 

That way somebody who wants Turbo could run `rails generate devise:install --turbo` to get Turbo behavior generated irrespective of Rails version.

You could make the default `true` for 5.x, but keep behavior as-is unless somebody specifically requests generation with Turbo in 4.x.

tmaier:
I am not sure what's the right approach in the context of devise.

I think I would have checked, if `Turbo` is defined with `defined?(::Turbo)`

Ref. https://github.com/hotwired/turbo-rails/blob/main/lib/turbo-rails.rb

shaunbruno:
> PR #5340 looks like it aims for a deeper integration into devise than what is proposed in this pull request.
> 
> Besides this, using turbo is also valid in a rails 6.1 context. So just checking for "> Rails 7" would exclude some valid use cases

Perhaps Devise should just be versioned with incremental support for Rails 6 dropped, in favor of a cleaner implementation for Rails 7+?

earnold:
Forgive me for chiming in on this, but I am getting confused by the different issues. 

Is there a canonical (at at least blessed) way to get Devise and Rails 7 playing nicely together?

carlosantoniodasilva:

PR #5469, opened 2022-02-10
Fix bug with sign_out_all_scopes trying to edit resource after it has already been destroyed

jcdogo:
When destroying a resource and resource is rememberable, `Devise.sign_out_all_scopes` attempts to edit the resource after resource has already been destroyed.  The reason is because `lib/devise/models/rememberable.rb:60: forget_me!` sets the `remember_token` and `remember_created_at` fields and then calls `save(validate: false)`.

The solution is pretty straightforward. We should first call `Devise.sign_out_all_scopes` and then `resource.destroy`.

### DIFF
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -63,8 +63,8 @@ class Devise::RegistrationsController < DeviseController
 
   # DELETE /resource
   def destroy
-    resource.destroy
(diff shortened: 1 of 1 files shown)


jcdogo:
By the way, we're using Devise with Mongoid and getting a `Exception: FrozenError: can't modify frozen BSON::Document` when destroying resources because of this bug.

PR #5474, opened 2022-02-25
Fixing compatibility with Rails 7 and API mode

morenocarullo:
This PR addresses this issue https://github.com/heartcombo/devise/issues/5443 . 

The test coverage for it is not ideal yet, as we are working to find a good way to have API-mode tests properly integrated (quick hack of forcing it to true does not work).

### DIFF
--- a/app/controllers/devise_controller.rb
+++ b/app/controllers/devise_controller.rb
@@ -3,6 +3,7 @@
 # All Devise controllers are inherited from here.
 class DeviseController < Devise.parent_controller.constantize
   include Devise::Controllers::ScopedViews
+  include Devise::Controllers::Rails7ApiMode
 
   if respond_to?(:helper)
     helper DeviseHelper
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -22,6 +22,7 @@ module Devise
 
   module Controllers
     autoload :Helpers,        'devise/controllers/helpers'
+    autoload :Rails7ApiMode,  'devise/controllers/rails7_api_mode'
     autoload :Rememberable,   'devise/controllers/rememberable'
     autoload :ScopedViews,    'devise/controllers/scoped_views'
     autoload :SignInOut,      'devise/controllers/sign_in_out'
--- a/lib/devise/controllers/rails7_api_mode.rb
+++ b/lib/devise/controllers/rails7_api_mode.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Devise
+  module Controllers
+    module Rails7ApiMode
+      extend ActiveSupport::Concern
+
+      class FakeRackSession < Hash
+        def enabled?
+          false
+        end
+      end
+
+      included do
+        before_action :set_fake_rack_session_for_devise
+        
+        private
+    
+        def set_fake_rack_session_for_devise
+          if Rails.configuration.respond_to?(:api_only) && Rails.configuration.api_only
+            request.env['rack.session'] ||= FakeRackSession.new
+          end
+        end
+      end
+    end
+  end
+end
--- a/test/controllers/rails7_api_mode_test.rb
+++ b/test/controllers/rails7_api_mode_test.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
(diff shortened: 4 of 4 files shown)


morenocarullo:
Hey @carlosantoniodasilva , this is a first part of the work to allow API mode to work as well. See https://github.com/heartcombo/devise/issues/5443.

hms:
Why the hard lock to Rails 7+ ?  The PR looks like it will work with Rails 6 without a hitch.

morenocarullo:
> Why the hard lock to Rails 7+ ? The PR looks like it will work with Rails 6 without a hitch.

you are right: but in Rails 6 it would not have any effect because `rack.session` will be a valid one.

dnpp73:
I think this is a good hack as bandaid. but I don't think this is a good solution. Some users may be using it in API mode and with sessions enabled. I think it would be broken from those users.

`DisabledSessionError` is useful information. because we can realize an implicit dependency on the `"rack.session"`

I think the problem is that devise implicitly depends on session. 
following lines ...
- https://github.com/heartcombo/devise/blob/v4.8.1/lib/devise/controllers/sign_in_out.rb#L48
- https://github.com/heartcombo/devise/blob/v4.8.1/lib/devise/controllers/sign_in_out.rb#L68
- https://github.com/heartcombo/devise/blob/v4.8.1/lib/devise/test/controller_helpers.rb#L81
- https://github.com/heartcombo/devise/blob/v4.8.1/lib/devise/test/controller_helpers.rb#L96

How about providing an option called `ignore_session`? like this

```ruby
Devise.setup do |config|
  config.ignore_session = true # default false
end
```

As a side note, there seems to be a similar discussion on warden's side.
https://github.com/wardencommunity/warden/pull/201

morenocarullo:
> I think this is a good hack as bandaid. but I don't think this is a good solution. Some users may be using it in API mode and with sessions enabled. I think it would be broken from those users.
> 
> `DisabledSessionError` is useful information. because we can realize an implicit dependency on the `"rack.session"`
> 
> I think the problem is that devise implicitly depends on session. following lines ...
> 
>     * https://github.com/heartcombo/devise/blob/v4.8.1/lib/devise/controllers/sign_in_out.rb#L48
> 
>     * https://github.com/heartcombo/devise/blob/v4.8.1/lib/devise/controllers/sign_in_out.rb#L68
> 
>     * https://github.com/heartcombo/devise/blob/v4.8.1/lib/devise/test/controller_helpers.rb#L81
> 
>     * https://github.com/heartcombo/devise/blob/v4.8.1/lib/devise/test/controller_helpers.rb#L96
> 
> 
> How about providing an option called `ignore_session`? like this
> 
> ```ruby
> Devise.setup do |config|
>   config.ignore_session = true # default false
> end
> ```
> 
> As a side note, there seems to be a similar discussion on warden's side. [wardencommunity/warden#201](https://github.com/wardencommunity/warden/pull/201)

I agree with you. I believe that, before beginning a journey to the "right thing" it would be great to have some blessing from the mantainers, or it will result in a lot of wasted work.

frg-digital:
@morenocarullo I forked and merged in my Devise version for an API ( Rails 7, API Only, JWT Auth). Works like a charm, thanks for your insight.

alhajrahmoun:
Getting 
```
NoMethodError (undefined method `destroy' for {}:Devise::Controllers::Rails7ApiMode::FakeRackSession):
```
when trying to sign out. Any thoughts?

internet-exploder:
Didn't work on rails 7.0.2.3 for me. Works only if comment "load_for_write" line in actionpack's gem at lib/action_dispatch/request/session.rb

```
      # Writes given value to given key of the session.
      def []=(key, value)
        # load_for_write!
        @delegate[key.to_s] = value
      end
```

I don't know if this is correct in any way.

justinshaw:
Is this likely to be merged any time soon? Or has anyone discovered an alternative for dealing with the problem this PR solves?

alexjoseps:
Any news?

graveetone:
any updates on merging this soultion or providing another one?

bussularf (approved):

morenocarullo:
This is lingering for a while (missing review approval from a reviewer with write access), I am closing it ...

PR #5475, opened 2022-02-25
Monkeypatch webrat for Nokogiri compatibility

petergoldstein:
This is an attempt to address the Webrat / Nokogiri compatibility issue discussed [here](https://github.com/sparklemotion/nokogiri/issues/2469).  It monkeypatches Webrat to explicitly add the old default arguments to the invocation of `to_xpath`.

This PR picked an unrelated failure in Rails main specs because of this [commit](https://github.com/rails/rails/commit/c2e756a944fd3ca2efa58bd285c0e75e0b4794ab), which removes the body content from a redirect response.  Devise has tests that check that body content, and those tests are now failing.  I would suggest we remove those specs, and I'm happy to put up a PR to do so.  Just let me know if such a PR is desired.

UPDATE: As the unrelated failure is now fixed in `main`, I've rebased this PR and it now exclusively addresses the Webrat issue.

cc: @carlosantoniodasilva, @flavorjones

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -15,6 +15,20 @@ I18n.load_path << File.expand_path("../support/locale/en.yml", __FILE__)
 require 'mocha/minitest'
 require 'timecop'
 require 'webrat'
+
+# Monkey patch for Nokogiri changes - https://github.com/sparklemotion/nokogiri/issues/2469
+module Webrat
+  module Matchers
+    class HaveSelector
+      def query
+          Nokogiri::CSS.parse(@expected.to_s).map do |ast|
+            ast.to_xpath("//", Nokogiri::CSS::XPathVisitor.new)
+          end.first
+      end
+    end
+  end
+end
+
 Webrat.configure do |config|
   config.mode = :rails
   config.open_error_files = false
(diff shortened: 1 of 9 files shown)


test/test_helper.rb
```diff
@@ -15,6 +15,20 @@
...
 require 'timecop'
 require 'webrat'
+
+# Monkey patch for Nokogiri changes - https://github.com/sparklemotion/nokogiri/issues/2469
+module Webrat
+  module Matchers
+    class HaveSelector
+      def query
+          Nokogiri::CSS.parse(@expected.to_s).map do |ast|
+            ast.to_xpath("//", Nokogiri::CSS::XPathVisitor.new)
```
flavorjones, line 25: An alternative to the naked `"//"` might be to reference `Nokogiri::XML::XPath::GLOBAL_SEARCH_PREFIX` in Nokogiri v1.13.0 and later.

petergoldstein: Because of the wide support matrix for Rails, we actually need to be able to run this with Nokogiri v1.9.1.  So I think the naked `"//"` is preferable until those Rails versions are dropped.

flavorjones: :+1: That monkeypatch should work on any Nokogiri version >= v1.6.2 and < 1.13.30

carlosantoniodasilva:
Merged in 90f46bac373416d6ad4b3541e6ca144b1f7ee527, appreciate it!

Side note: I really need to get rid of webrat at some point, indeed. :)

Issue #5140, opened 2019-09-24
Rails6 without ActionMailer won't boot with zeitwerk eager-loading


sbull:
## Environment

- Ruby 2.6.3
- Rails 6.0.0
- Devise 4.7.1

## Current behavior

In a Rails 6 app without ActionMailer included (`rails new testapp --skip-action-mailer`), the default zeitwerk loader errors when eager-loading due to mailers/devise/mailer.rb not defining Devise::Mailer. Output from `rails zeitwerk:check`:
`expected file /usr/local/bundle/gems/devise-4.7.1/app/mailers/devise/mailer.rb to define constant Devise::Mailer,`

## Expected behavior

Zeitwerk eager-loading should work without ActionMailer. Suggested options:
- (easy) Define a dummy Devise::Mailer when ActionMailer is not defined.
- (hard?) Customize the zeitwerk loader for when ActionMailer is not used.

fxn:
Idiomatically, you want to ignore that file. It could done with something like this in Devise (off the top of my head):

```ruby
initializer "devise.configure_zeitwerk_if_enabled" do |app|
  if app.zeitwerk_enabled? && !defined?(ActionMailer)
    app.autoloaders.main.ignore("#{__dir__}/relative/path/to/devise/mailer.rb")
  end
end
```

mracos:
@fxn I would only add a check for `Rails::Version::MAJOR >= "6"` so it does not break on older versions of Rails where the method `zeitwerk_enabled?` does not exist, wdyt?

fxn:
Oh yes, totally.

jeppester:
In case others come here in search for a quick solution:

It's possibly to work around the issue by simply actionmailer in `config/application.rb` along with the other frameworks:
```ruby
require "action_mailer/railtie"
```

gkats:
For what it's worth, in a Rails 6.0.1 app without the `ActionMailer` railtie, it seems that defining `Devise::Mailer` as an empty class does not break anything. The app sends emails via a 3rd party service.

```ruby
# <path-to-gems>/devise-4.7.1/app/mailers/devise/mailer.rb

if defined?(ActionMailer)
  class DeviseMailer < Devise.parent_mailer.constantize
    # whatever was already there
  end
else
  class DeviseMailer
  end
end
```
Tested authentication, registration, validation, forgot password flows.

Obviously inferior solution to explicitly ignoring the file in an initializer, but thought I'd share.

[...]

PR #5485, opened 2022-04-23
Fixing Zeitwerk errors when running Devise without ActionMailer

aldent95:
This fixes the longstanding issue #5140 where Zeitwerk throws errors around when using Devise without ActionMailer.

Overall behaviour shouldn't be effected. The new empty class is only used when the application would otherwise error our (ActionMailer not available and Zeitwerk enabled)

### DIFF
--- a/app/mailers/devise/mailer.rb
+++ b/app/mailers/devise/mailer.rb
@@ -28,6 +28,8 @@ if defined?(ActionMailer)
     end
   end
 else
-  class Devise::Mailer
+  if Rails.autoloaders.zeitwerk_enabled?
+    class Devise::Mailer
+    end
   end
 end
--- a/app/mailers/devise/mailer.rb
+++ b/app/mailers/devise/mailer.rb
@@ -27,4 +27,7 @@ if defined?(ActionMailer)
       devise_mail(record, :password_change, opts)
     end
   end
+else
+  class DeviseMailer
+  end
 end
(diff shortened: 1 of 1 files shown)


lxxxvi:
What is the use for `.zeitwerk_enabled?` clause in the first place? Wouldn't it be possible just to have the empty class if `ActionMailer` is not there?

`.zeitwerk_enabled?` was introduced around Rails 6.0 and devise is backward compatible with Rails 4 (I guess), so this condition may cause issues in earlier versions.

lomefin:
Is there any real damage if it just defines the class if ActionMailer is not defined?
I understand that adding the rule makes it more focused on solving the error.
But it adds a new dependency, so now we don't need ActionMailer but we need Zeitwerk.

I added `Devise::Mailer` on an initializer and seems to work alright.

If possible, I would just define the class and leave it empty if there is no ActionMailer

rafaelfranca:
Would devise even work without Action Mailer? So many of its default modules requires a mailer to exist.

Can you provide how you are planning to use Devise without a mailer?

Fs00:
@rafaelfranca In the company I work for, we are using Devise without Action Mailer in a Rails app that powers an internal portal where users log in via LDAP. In that scenario we don't need mailers at all, since user registration/user blocking/password change are all handled by AD.

nashby:
@aldent95 thank for PR! I've just merged https://github.com/heartcombo/devise/pull/5731 with a bit different approach.

PR #5487, opened 2022-04-29
Disable turbo by default

n-studio:
Until `devise` is compatible with `Turbo` out of the box, I suggest we disable it by default.

A lot of newcomers installing Rails and devise for the first time might become discouraged when they can't figure out why an essential feature such as authentication doesn't work without extensive efforts.


carlosantoniodasilva:

PR #5490, opened 2022-05-05
Make the unlock URL itempotent

neilvilela:
The path `GET /resource/unlock?unlock_token=TOKEN` is not currently
idempotent, which brings some problems when the page is hit by an
antivirus, prefetch, or anything other than the user. For example,
Microsoft SafeLinks technology visits URLs sent in emails to make sure
that the URL doesn't have malware.

This change keeps, when configured, the unlock token after the resource
has been unlocked. If the user happens to click on the unlock link after
it has already been visited, it will behave as if this was the first
visit, even though the resource is technically unlocked since the first
visit. This makes the GET request idempotent and doesn't break the
current behavior.

This may add a side effect on database queries that check how many
resources have `unlock_token` not null to count how many resources are
unlocked. To address this side effect, the new behavior is turned off by
default.

The [first version] of the Lockable module logged the user in, but this
is not true since https://github.com/heartcombo/devise/pull/1493 was merged. When a resource is locked again,
another unlock token is created and overwrites the older one. Keeping
the most recent unlock token, then, doesn't represent a security issue.

[first version]:
https://github.com/heartcombo/devise/commit/d2fa737aa094f2e4808d9d4388b20d7081861511


neilvilela:
Oh, I noticed a possible security issue: by possessing the unlock token, someone may reset the count of login attempts and try to login indefinitely, as long as the account isn't locked. I'll address this issue.

test/models/lockable_test.rb
```diff
@@ -110,6 +110,30 @@ def setup
     assert_equal 0, user.reload.failed_attempts
   end
 
+  test "unlocks a user keeping its last unlock_token when configured" do
```
marivaldo, line 113:
```suggestion
  test "unlocks an user keeping its last unlock_token when configured" do
```

neilvilela: That's an interesting thing in the English language!

In this case, _a_ user is correct, as the a/an usage depends on the _sound_ of the next word. It's pronounced something like _yuzer_, so it has a consonant initial sound.

Reference: https://forum.wordreference.com/threads/a-an-user.480322/#:~:text=A%20user%20is%20correct.,vowel%20as%20the%20first%20letter.

PR #5492, opened 2022-05-16
Change define_helpers parameters to String

gardnerapp:
The `define_helpers` method takes an instance of `Devise::Mapping` as an argument, which is a very large object. Within the method a local variable `mapping`,  which is set to the name instance variable of the mapping object. This variable is then used in the method to meta program helper methods like` user_signed_in?` etc.

It is unnecessary to pass an entire `Devise::Mapping ` to `define_helpers`. Instead I propose passing the @name of the mapping directly to `define_helpers`. The assignment of the `mapping` variable can now also be removed. This change simplifies code and speeds up the program by avoiding passing large objects to functions.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -354,7 +354,7 @@ module Devise
     mapping = Devise::Mapping.new(resource, options)
     @@mappings[mapping.name] = mapping
     @@default_scope ||= mapping.name
-    @@helpers.each { |h| h.define_helpers(mapping) }
+    @@helpers.each { |h| h.define_helpers(mapping.name) }
     mapping
   end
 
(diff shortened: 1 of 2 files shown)


p8:
Hey @gardnerapp, thanks for contributing to Devise!
It seems like passing the mapping instead of the name was changed in:
https://github.com/heartcombo/devise/commit/ef841ca17d37040e89072debb1e1d8b5f6726c34
It doesn't seem to be a requirement for the change, just a refactoring, and the method is marked as `:nodoc:` so I'm guessing it's ok to change it.

p8:
@gardnerapp Sorry, `:nodoc:` means it's not documented as public API.
Third parties shouldn't use it because it might change in the future.

> :nodoc: / :nodoc: all
> Don't include this element in the documentation. For classes and modules, the methods, aliases, constants, and attributes directly within the affected class or module will also be omitted. By default, though, modules and classes within that class of module will be documented. This is turned off by adding the all modifier.

https://ruby-doc.org/stdlib-1.9.1/libdoc/rdoc/rdoc/index.html#label-Directives

PR #5493, opened 2022-05-19
View helper methods

gardnerapp:
If/Else conditional statements based on a users session status are relatively common at the controller and view level of applications. These conditional statements are repetitive, accumulate quickly and make our code bulky . Great programs are DRY and this pull request seeks to DRY up our code by creating helper methods that abstract if/else conditionals into methods where a block is called based on a resources sign_in/out state. 

Examples: 

View -> 

`user_signed_out do
   <%= link_to "Sign In", new_user_session_path %>
end`

Controller ->

`
def some_action;
   admin_signed_in { render @posts };
   admin_signed_out { redirect_to root_path };
end 
`

### DIFF
--- a/test/views/helper_methods_test.rb
+++ b/test/views/helper_methods_test.rb
@@ -1,11 +1,28 @@
 # frozen_string_literal: true
 require 'test_helper'
 
+module ActionView
+  include Devise::Views::Helpers
+end
 
-class ViewsHelperMethodsTest < Devise::ControllerTestCase
+class ViewsHelperMethodsTest < Devise::IntegrationTest
 
-  #test 'defines methods like foo' do
-  #  assert_respond_to ActionView, :foo
-#  end
+  # Figure out which Rails Test class/module to inherit from to properly test
+  # setup test to run through devise initialization
+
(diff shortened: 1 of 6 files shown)


test/views/helper_methods_test.rb
```diff
@@ -0,0 +1,22 @@
...
+require 'test_helper'
+
+module ActionView
+  include Devise::Views::Helpers
+end
+
+class ViewsHelperMethodsTest < Devise::IntegrationTest
+
+  test 'Action View includes Devise::Views::Helpers' do
+    assert_includes  ActionView.ancestors, Devise::Views::Helpers
```
marivaldo, line 11:
```suggestion
    assert_includes ActionView.ancestors, Devise::Views::Helpers
```

gardnerapp: Can we get this pulled?

gardnerapp: Thank you for the review. Could we get this pull request submitted?

nashby:
Hey @gardnerapp! Thanks for the PR but I think it's something that you can have in your application as a helper

PR #5494, opened 2022-05-24
Replaced `<br/>` with `<br>`

iainbeeston:
In regular HTML `<br>` is a void element, so it doesn't
need a closing tag or a closing slash. See the WhatWG
spec here: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element

Many of the shared templates used by devise use `<br/>`
to separate lines. To make that more correct I've replaced
them with `<br>` instead.

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -4,8 +4,8 @@
   <%= render "devise/shared/error_messages", resource: resource %>
 
   <div class="field">
-    <%= f.label :email %><br />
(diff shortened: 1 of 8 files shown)


iainbeeston:
@p8 Thanks, I've replaced those with paragraph tags now. I've also updated the commit message to explicitly say that this reverts b86c1c241b9bc6f6679ff1da78ffda880dd1bf9c. Could you please take another look? (Especially at the HTML changes because I did them manually and it would be easy for me to make a mistake)

PR #5499, opened 2022-06-16
fix: add status codes to redirects and failure app

KonnorRogers:
- Adds 303 (see_other) status codes to redirects
- Adds a 401 status code when Warden fails to authenticate.

- This does not directly allow Turbo streams, but does allow other solutions to hook into status codes to handle responses properly via Ajax. (Things like Mrujs or Request.js)

## Additional notes

totally open to discussion on this. This arised from a personal fork I was using and am happy to close the PR should another solution arise.

### DIFF
--- a/app/controllers/devise/confirmations_controller.rb
+++ b/app/controllers/devise/confirmations_controller.rb
@@ -14,7 +14,7 @@ class Devise::ConfirmationsController < DeviseController
     if successfully_sent?(resource)
       respond_with({}, location: after_resending_confirmation_instructions_path_for(resource_name))
     else
-      respond_with(resource)
+      respond_with(resource, status: Devise.failure_status_code)
     end
   end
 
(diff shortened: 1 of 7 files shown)


rafaelfranca:
This new behavior needs to be configurable, but default. Can you work towards that direction?

ParamagicDev:
@rafaelfranca I'll see about making a configurable status code. I previously tweeted about Devise + Warden authentication failures and @carlosantoniodasilva seemed apprehensive about the "failure_app" here. With its state, it can be configured to return a 401 or 200 (current behavior). 

Do you think the failure_app as is should be left in with a configurable status code? Or leave it to a user to custom configure their own "failure_app" with a 401 status code?

ParamagicDev:
I havent added it to the generators yet nor added tests, but I reworked this PR so it does not add the "redirect_status_code" to "POST" requests since its not needed. It may be confusing since its not applied everywhere, but this is the change that enables the ability to use "fetch" requests with Devise.

leastbad:
So, if I understand correctly, I could add this to my initializer:
```ruby
Devise.setup do |config|
  config.redirect_status_code = 303
  config.failure_status_code = 422
  config.authentication_failure_status_code = 401
end
```
And suddenly, all of the problems that we have to come up with hacky workarounds for go away?

Magic. Lighting all of my black candles for this one.

printfinn:
The proposed fix looks clean and flexible!

carlosantoniodasilva:

PR #5506, opened 2022-07-03
5505 - trigger setter instead of setting local var

yasirazgar:
Fixes: https://github.com/heartcombo/devise/issues/5505

Changed `resource = warden.user(resource_name)` to `self.resource = warden.user(resource_name)` in [devise_controller#require_no_authentication](https://github.com/heartcombo/devise/blob/main/app/controllers/devise_controller.rb#L114) so that [setter resource =](https://github.com/heartcombo/devise/blob/main/app/controllers/devise_controller.rb#L94) is triggered and instance variable @user is populated.

### DIFF
--- a/app/controllers/devise_controller.rb
+++ b/app/controllers/devise_controller.rb
@@ -111,7 +111,7 @@ MESSAGE
       warden.authenticated?(resource_name)
     end
 
-    if authenticated && resource = warden.user(resource_name)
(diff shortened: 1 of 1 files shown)


nashby:
@yasirazgar thanks for the PR! This is intentional that we set `resource` here and not `self.resource` as was explained in https://github.com/heartcombo/devise/issues/5505

PR #5507, opened 2022-07-04
Remeberable: Use has_changes_to_save? instead of changed?

theRealNG:
Created pull request for this issue: https://github.com/heartcombo/devise/issues/5421


carlosantoniodasilva:
Please see https://github.com/heartcombo/devise/issues/5421#issuecomment-1451094724, `changed?` is not really deprecated, just the confusing behavior of it was, but it is interchangeable with `has_changes_to_save?` in that context. Thanks!

PR #5514, opened 2022-08-29
Bug Fix: Devise Checkbox TRUE VALUES not compatible with HTML specs (missing "on" value)

itay-grudev:
See:

* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#attr-value
* https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type%3Dcheckbox)
* https://html.spec.whatwg.org/multipage/input.html#dom-input-value-default-on

This issue causes the remember me functionality not to work correctly, especially when overriding the default styles as check boxes by default in all major browsers send the value `on` as required by the HTML specifications.

See also:

https://stackoverflow.com/questions/5987075/devise-remember-me-not-working-localhost-issue

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -60,7 +60,7 @@ module Devise
   NO_INPUT = []
 
   # True values used to check params
-  TRUE_VALUES = [true, 1, '1', 't', 'T', 'true', 'TRUE']
+  TRUE_VALUES = [true, 1, '1', 'on', 'ON', 't', 'T', 'true', 'TRUE']
 
   # Secret key used by the key generator
   mattr_accessor :secret_key


itay-grudev:
@carlosantoniodasilva

itay-grudev:
Ping! Any progress on this?

itay-grudev:
@carlosantoniodasilva I think this project needs new contributors or it's of risk of dying. Do you need help?

tarellel (approved):

itay-grudev:
@nashby Done.

itay-grudev:
@nashby I rebased against main. This probably will fix the test failures.

nashby:
@itay-grudev thanks, failing specs are not related to this change so this is fine. Could you please squash your commits?

itay-grudev:
GitHub does that when you click merge and create one merge commit on the entire PR diff.

itay-grudev:
If not - enable it. It keeps your repo clean without using extra human time.

itay-grudev:
image

nashby:
@itay-grudev the thing is that I want to have a merge commit but I don't want to have 3 commits on top (it's fine to have multiple commits in one PR but not in this case). So the only way to do it if you squash them, sorry for the trouble

itay-grudev:
I wish to inform you that because of the non-linearity of the commits applied not in order over the time-span of over **2 years** - it took me 10min to understand / resolve the situation and them. And I am git skilled. I knew what I was doing.

This is as opposed to the UI-based squash merge that would have taken less than 30s if enabled.

As a true software engineer I despise this bureaucratic inefficiency hence this complaint.

itay-grudev:
@nashby Done.

PR #5516, opened 2022-09-03
Add use_turbo configuration with default to false.

gobijan:
This PR enables Rails 7 and Devise to work out of the box.
Devise now opts out of turbo by default for its forms by making use of
data: { turbo: Devise.use_turbo} setting on forms.

I've seen this PR https://github.com/heartcombo/devise/pull/5487 but it is not configurable and hardcoded. So people that have opted into turbo can do this via a setting (maybe they fixed notifications by patching devise etc.).
I went with an approach that conforms to the style I've seen by inspecting the devise codebase.

My PR defaults to not using turbo in devise forms by default. I know this is a minor breaking change for people that might have patched the problem otherwise but it's the way to offer a sane default experience with Devise and Rails 7 until the maintainers decide on a real solution.

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -1,6 +1,6 @@
 <h2>Resend confirmation instructions</h2>
 
-<%= form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: { method: :post }, data: { devise_mapping.use_turbo }) do |f| %>
+<%= form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: { method: :post }, data: { turbo: devise_mapping.use_turbo }) do |f| %>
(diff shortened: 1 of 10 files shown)


gobijan:
Needs some fixing. I close.

guihatano (approved):

msducheminjr:
On the `registrations/edit` view, it is also necessary to handle the confirm on `Cancel my account` so that a user doesn't accidentally misclick and bypass the confirmation in Turbo.

Something like this:
```erb
<% confirm_attribs = devise_mapping.use_turbo ? { form: { data: { turbo_confirm: "Are you sure?" } } } : { data: { confirm: "Are you sure?" } } %>
<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), **confirm_attribs, method: :delete %>
```

carlosantoniodasilva:

PR #5517, opened 2022-09-06
Critical security update Rack:  SNYK-RUBY-RACK-2848599

gentlemich:


rafaelfranca:
Thank you for the pull request but the Gemfile.lock file is only used for development of this gem. There is no security risk and no reason to upgrade this.

PR #5531, opened 2022-10-20
Passkeys support: Rename Devise::Strategies::Authenticatable => PasswordAuthenticatable

tcannonfodder:
Rename Devise::Strategies::Authenticatable => PasswordAuthenticatable
* To begin the process of refactoring Devise to support passkeys (https://github.com/heartcombo/devise/issues/5527), we need to decouple authentication from being specifically tied to passwords.
	* The first step is explicitly prefixing generically-named components like `Authenticatable` with `Password`, since they concern password authentication

Add Devise::Strategies::Authenticatable shim w/ deprecation warning
* To maintain compatability with exisiting installations, we need the `Devise::Strategies::Authenticatable` class.
	* This re-adds the class in, subclassing from `PasswordAuthenticable` and printing a deprecation warning

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ pkg
 log
 test/tmp/*
 gemfiles/*.lock
+.DS_Store
--- a/lib/devise/strategies/authenticatable.rb
+++ b/lib/devise/strategies/authenticatable.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'devise/strategies/password_authenticatable'
+
+module Devise
+  module Strategies
+    class Authenticatable < PasswordAuthenticatable
+      ActiveSupport::Deprecation.warn <<-DEPRECATION.strip_heredoc
+      [Devise] `Devise::Strategies::Authenticatable` is deprecated and will be
+      removed in the next major version.
+      Use `Devise::Strategies::PasswordAuthenticatable` instead.
+      DEPRECATION
+    end
+  end
+end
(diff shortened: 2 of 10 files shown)


lib/devise/strategies/authenticatable.rb
```diff
@@ -1,178 +1,15 @@
...
+    class Authenticatable < PasswordAuthenticatable
+      ActiveSupport::Deprecation.warn <<-DEPRECATION.strip_heredoc
+      [Devise] `Devise::Strategies::Authenticatable` is deprecated and will be
+      removed in the next major version.
+      Use `Devise::Strategies::PasswordAuthenticatable` instead.
+      DEPRECATION
     end
   end
-end
+end
```
matiasgarciaisaia, line 15: Cosmetic, but we should fix these missing trailing newlines.

.gitignore
```diff
@@ -9,3 +9,4 @@ pkg
 log
 test/tmp/*
 gemfiles/*.lock
+.DS_Store
```
matiasgarciaisaia, line 12: This change doesn't belong to this PR.

matiasgarciaisaia (commented):
The PR looks internally consistent 👍 

The name migrations are clean (ie, `DatabasePasswordAuthenticable` is exactly what `DatabaseAuthenticable` was, except the name changes - and the same for the other two strategies).

I'm wondering if the PR needs refreshing since a while passed - any new strategy that needs to be name-scoped with `Password`? Did any of the strategies change (and we should update the new classes)?

PR #5539, opened 2022-12-13
Handle multiple ORMs used by the parent app

Nikko-barbier:
**What?**
Update the forked Devise gem until the version 4.8.1
**Why?**
In order to update the rails version to version 6, there is some gem to update before. Inncluding Devise. 
**Jira**
https://universedev.atlassian.net/browse/GLB-488


carlosantoniodasilva:
Can you please add more details about the change, the reasoning, what it fixes, etc? Thanks.

PR #5545, opened 2023-01-11
Apply changes for rails-ujs and Turbo compatibility

JunichiIto:
## Purpose
I want to use Devise easily for both Rails 6.1(rails-ujs) and Rails 7.0(Turbo). Typical steps are like this:

1. rails new
1. echo 'gem "devise"' >> Gemfile
1. bundle install
1. rails g devise:install
1. rails g devise User
1. rails db:migrate
1. [edit application controller](https://github.com/JunichiIto/devise-rails-6-1-sandbox/blob/main/app/controllers/application_controller.rb#L2-L8)
1. [add edit account link and sign out button](https://github.com/JunichiIto/devise-rails-6-1-sandbox/blob/main/app/views/home/index.html.erb#L5-L10)
1. rails c

My PR enables it.

### Status codes
- Returns 422 Unprocessable Entity status instead of 200 when validation error occurs for the following actions:

```
POST /resource/confirmation
POST /resource/password
PUT  /resource/password
POST /resource
PUT  /resource
POST /resource/unlock
```

- Returns 303 See Other status instead of 302 for the following actions:

```
DELETE /resource
DELETE /resource/sign_out
```

- Returns 302 status instead of 200 when sign-in fails for turbo_stream requests.

### Views
- Recommends to use `button_to` to sign-out because it works for both rails-ujs and Turbo:

```erb
<%# This works for rails-ujs and Turbo %>
<%= button_to 'Sign out', destroy_user_session_path, method: :delete %>
```

- If you want to use `link_to` to sign-out, please choose one or the other:

```erb
<%# Turbo %>
<%= link_to 'Sign out', destroy_user_session_path, data: { turbo_method: :delete } %>

<%# rails-ujs %>
<%= link_to 'Sign out', destroy_user_session_path, method: :delete %>
```

## Example apps

I created example apps for Turbo and rails-ujs with my version of Devise. They also have system tests.

- [Rails 7.0 + Turbo](https://github.com/JunichiIto/devise-rails-7-sandbox/tree/with-customized-devise) ([diff to setup](https://github.com/JunichiIto/devise-rails-7-sandbox/compare/d4ff16c63bff31c47b9ef126da03396341f8683b...abf02f41f0f4f90ec47030608d1713b90cd778e3))
- [Rails 6.1 + rails-ujs](https://github.com/JunichiIto/devise-rails-6-1-sandbox) ([diff to setup](https://github.com/JunichiIto/devise-rails-6-1-sandbox/compare/438f8f8a0e76ca31dfcd9111fc7e3852a33a28d7...0ae2b80cd6f2678375636ced561ac83226cce785))

Please feel free ask me if you have any questions or requests.

### DIFF
--- a/app/controllers/devise/confirmations_controller.rb
+++ b/app/controllers/devise/confirmations_controller.rb
@@ -14,7 +14,7 @@ class Devise::ConfirmationsController < DeviseController
     if successfully_sent?(resource)
       respond_with({}, location: after_resending_confirmation_instructions_path_for(resource_name))
     else
-      respond_with(resource)
+      respond_with resource, status: :unprocessable_entity
     end
   end
 
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -38,6 +38,6 @@
 
 <h3>Cancel my account</h3>
 
-<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?" }, method: :delete %></p>
+<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?", turbo_confirm: "Are you sure?" }, method: :delete %></p>
 
 <%= link_to "Back", :back %>
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -20,6 +20,6 @@
 
 <%- if devise_mapping.omniauthable? %>
   <%- resource_class.omniauth_providers.each do |provider| %>
-    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), method: :post %><br />
+    <%= button_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), data: { turbo: false } %><br />
   <% end %>
 <% end %>
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -38,7 +38,7 @@ module Devise
       if http_auth?
         http_auth
       elsif warden_options[:recall]
-        recall
+        request_format == :turbo_stream ? redirect : recall
       else
         redirect
       end
@@ -167,7 +167,7 @@ module Devise
     end
 
     def skip_format?
-      %w(html */*).include? request_format.to_s
+      %w(html */* turbo_stream).include? request_format.to_s
     end
 
     # Choose whether we should respond in an HTTP authentication fashion,
--- a/app/controllers/devise/passwords_controller.rb
+++ b/app/controllers/devise/passwords_controller.rb
@@ -18,7 +18,7 @@ class Devise::PasswordsController < DeviseController
     if successfully_sent?(resource)
       respond_with({}, location: after_sending_reset_password_instructions_path_for(resource_name))
     else
-      respond_with(resource)
+      respond_with resource, status: :unprocessable_entity
     end
   end
 
@@ -47,7 +47,7 @@ class Devise::PasswordsController < DeviseController
       respond_with resource, location: after_resetting_password_path_for(resource)
     else
       set_minimum_password_length
-      respond_with resource
+      respond_with resource, status: :unprocessable_entity
     end
   end
 
--- a/app/controllers/devise/registrations_controller.rb
+++ b/app/controllers/devise/registrations_controller.rb
@@ -31,7 +31,7 @@ class Devise::RegistrationsController < DeviseController
     else
       clean_up_passwords resource
       set_minimum_password_length
-      respond_with resource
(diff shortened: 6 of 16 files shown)


JunichiIto:
@carlosantoniodasilva @rafaelfranca Hello, I'd appreciate if you would review my PR.

carlosantoniodasilva:
@JunichiIto really appreciate your work here, but please be a bit patient. If anything there's breaking changes to consider.

I'm looking into some current build failures before anything else, and we'll get to it.

app/controllers/devise/confirmations_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
     if successfully_sent?(resource)
       respond_with({}, location: after_resending_confirmation_instructions_path_for(resource_name))
     else
-      respond_with(resource)
+      respond_with resource, status: :unprocessable_entity
```
carlosantoniodasilva, line 17: Note: We will be able to use responders to handle this for us across the board going forward.

Please see https://github.com/heartcombo/responders/blob/fb9f787055a7a842584ce351793b249676290090/CHANGELOG.md#unreleased (still unreleased at the time of this writing)

app/views/devise/registrations/edit.html.erb
```diff
@@ -38,6 +38,6 @@
 
 <h3>Cancel my account</h3>
 
-<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?" }, method: :delete %></p>
+<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?", turbo_confirm: "Are you sure?" }, method: :delete %></p>
```
carlosantoniodasilva, line 41: This looks like a good option to support both confirm options.

app/views/devise/shared/_links.html.erb
```diff
@@ -20,6 +20,6 @@
 
 <%- if devise_mapping.omniauthable? %>
   <%- resource_class.omniauth_providers.each do |provider| %>
-    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), method: :post %><br />
+    <%= button_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), data: { turbo: false } %><br />
```
carlosantoniodasilva, line 23: Yeah, I think the best option for this one might be a button since this is not something that should go through turbo I guess?

The only other option that could work would be two methods. (`method: :post` and `data: { turbo_method: :post }`), but again I'm not sure about the Turbo behavior for something like this so may be better to disable it entirely.

lib/devise/failure_app.rb
```diff
@@ -38,7 +38,7 @@ def respond
       if http_auth?
         http_auth
       elsif warden_options[:recall]
-        recall
+        request_format == :turbo_stream ? redirect : recall
```
carlosantoniodasilva, line 41: I am afraid that while redirecting works in the sense that it will re-display the form as expected, it will lose some context about the flash / error message and show the more generic "you need to be signed in" instead of the "invalid email or password". 

Ideally we'd still recall, but change the status, so it has that same behavior.

carlosantoniodasilva (commented):
Hey @JunichiIto, appreciate our work here.

I left a few thoughts below as I looked through the changes, but no need to change anything yourself here. I've been working on related changes https://github.com/heartcombo/devise/pull/5548 on top of what you did here and a few other PRs/issues, to try to accommodate for turbo without making it a full breaking change, and integrating with responders. Your updates here were extremely helpful to get a better understanding of what was needed, and get that up and running, so thank you.

Please feel free to give that one a shot, and leave any feedback or issue you may encounter there. Let's also keep this open for now, it can be closed if / when that one gets merged. Thanks again.

app/views/devise/shared/_links.html.erb
JunichiIto: As far as I tested, Turbo didn't work with OAuth because it raises CORS error.

Here is the result when I click the link generated via `<%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), method: :post, data: { turbo_method: :post } %>`:

Screenshot 2023-02-01 at 19 20 54

So I had to change this line from link to button and add `data: { turbo: false }` option.

JunichiIto:
@carlosantoniodasilva Thank you for your feedback. I'm glad to hear you've started working on https://github.com/heartcombo/devise/pull/5548 and looking forward to the day when it gets released.

app/views/devise/shared/_links.html.erb
carlosantoniodasilva: Good point, I hadn't thought about CORS but yeah since it's all happening via JS it will trigger a CORS error when trying to follow the redirect with fetch under the hood.

`button_to` it is then, to support this correctly on all cases, since there's no way to do POST with turbo like rails-ujs does. Thanks!

carlosantoniodasilva:

Issue #5448, opened 2022-01-05
Rails 7 data-confirm not working for User deletion


odinsride:
## Environment

- Ruby **3.0.3p157**
- Rails **7.0.0**
- Devise **4.8.1**

## Current behavior

When clicking the default "Cancel my account" link, the data-confirm is not triggering, and instead it is deleting the user account right away. I am using Rails 7 with turbo / stimulus / importmap, although I have `data-turbo` set to `false` for my edit user form. Here is the form code:

```
<%= form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put, 'data-turbo': false }) do |f| %>
  <%= render DeviseErrorsComponent.new(resource: resource) %>

  <div class="md:flex md:space-x-4">
    <div class="flex-1">
      <div class="o-control-group">
        <%= f.label :first_name, class: "o-label" %>
        <%= f.text_field :first_name, autofocus: true, autocomplete: 'disabled', class: "o-input" %>
      </div>
    </div>

    <div class="flex-1">
      <div class="o-control-group">
        <%= f.label :last_name, class: "o-label" %>
        <%= f.text_field :last_name, autocomplete: 'disabled', class: "o-input" %>
      </div>
    </div>
  </div>

  <div class="o-control-group">
    <%= f.label :timezone, class: "o-label" %>
    <%= f.time_zone_select :timezone, ActiveSupport::TimeZone.us_zones, { default: "Eastern Time (US & Canada)" }, { class: "o-input" } %>
  </div>
  
  <div class="o-control-group">
    <%= f.label :email, class: "o-label" %>
    <%= f.email_field :email, placeholder: "john@gmail.com", autocomplete: "email", class: "o-input" %>
  </div>

  <% if devise_mapping.confirmable? && resource.pending_reconfirmation? %>
    <div>Currently waiting confirmation for: <%= resource.unconfirmed_email %></div>
  <% end %>

  <div class="o-control-group">
    <%= f.label :current_password, class: "o-label" %>
    <%= f.password_field :current_password, autocomplete: "current-password", placeholder: "●●●●●●●●", class: "o-input" %>
    <div class="ml-2 mt-2 text-sm font-semibold text-gray-400">Your current password is required to confirm your changes</div>
  </div>

  <div class="text-center">
    <%= f.submit "Update", class: "o-btn o-btn--primary w-1/3" %>
[...]

PR #5548, opened 2023-01-31
Integrate with Hotwire/Turbo by configuring error and response statuses

carlosantoniodasilva:
Treat `:turbo_stream` request format as a navigational format, much like HTML, so Devise/responders can work properly.

Allow configuring the `error_status` and `redirect_status` using the latest responders features, via a new custom Devise responder, so we can customize the both responses to match Hotwire/Turbo behavior, for example with `422 Unprocessable Entity` and `303 See Other`, respectively. The defaults aren't changing in Devise itself (yet), so it still responds on errors cases with `200 OK`, and redirects on non-GET requests with `302 Found`, but new apps are generated with the new statuses and existing apps can opt-in. Please note that these defaults might change in a future release of Devise.

PRs/Issues references:

Closes https://github.com/heartcombo/devise/pull/5545
Closes https://github.com/heartcombo/devise/pull/5529
Closes https://github.com/heartcombo/devise/pull/5516
Closes https://github.com/heartcombo/devise/pull/5499
Closes https://github.com/heartcombo/devise/pull/5487
Closes https://github.com/heartcombo/devise/pull/5467
Closes https://github.com/heartcombo/devise/pull/5440
Closes https://github.com/heartcombo/devise/pull/5410
Closes https://github.com/heartcombo/devise/pull/5340

Closes https://github.com/heartcombo/devise/issues/5542
Closes https://github.com/heartcombo/devise/issues/5530
Closes https://github.com/heartcombo/devise/issues/5519
Closes https://github.com/heartcombo/devise/issues/5513
Closes https://github.com/heartcombo/devise/issues/5478
Closes https://github.com/heartcombo/devise/issues/5468
Closes https://github.com/heartcombo/devise/issues/5463
Closes https://github.com/heartcombo/devise/issues/5458
Closes https://github.com/heartcombo/devise/issues/5448
Closes https://github.com/heartcombo/devise/issues/5446
Closes https://github.com/heartcombo/devise/issues/5439

### TODO

- [x] Release responders and point to it.
- [x] Changelog note about Devise `responder` and how to override it back if needed.
- [x] Post install message
- [x] Upgrade guide / wiki
  - https://github.com/heartcombo/devise/wiki/How-To:-Upgrade-to-Devise-4.9.0-(Hotwire---Turbo-integration)

### Testing this branch on an existing app

In order to test this branch for turbo compatibility on an existing app, update your Gemfile to point Devise to this branch, and Responders to version 3.1+:

```ruby
gem "devise", github: "heartcombo/devise", branch: "ca-turbo"
gem "responders", "~> 3.1"
```

Then, add the following config to your Devise initializer (located in `config/initializers/devise.rb`):

```ruby
  # ==> Hotwire/Turbo configuration
  config.responder.error_status = :unprocessable_entity
  config.responder.redirect_status = :see_other
```

_Note: these are going to be the new defaults for newly generated apps, and will likely become the defaults for Devise in a future release as well, but for now they're configurable and default to the current values for backwards compatibility._

Check the upgrade wiki (WIP) for more info: https://github.com/heartcombo/devise/wiki/How-To:-Upgrade-to-Devise-4.9.0-(Hotwire---Turbo-integration)


CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
...
+  * Add support for Hotwire + Turbo, default in Rails 7+.
+    * `:turbo_stream` is now treated as a navigational format, so it works like HTML navigation when using Turbo. Note: if you relied on `:turbo_stream` to be treated as a non-navigational format before, you can reconfigure your `navigational_formats` in the Devise initializer file to exclude it.
+    * Devise requires the latest `responders` version, which allows configuring the status used for validation error responses (`error_status`) and for redirects after POST/PUT/PATCH/DELETE requests (`redirect_status`). For backwards compatibility, Devise keeps `error_status` as `:ok` which returns a `200 OK` response, and `redirect_status` to `:found` which returns a `302 Found` response, but you can configure it to return `422 Unprocessable Entity` and `303 See Other` to match the behavior expected by Hotwire/Turbo:
+
+     ```ruby
+     # config/initializers/devise.rb
+     Devise.setup do |config|
+       # ...
+       config.responder.error_status = :unprocessable_entity
+       config.responder.redirect_status = :see_other
```
excid3, line 14: What are your thoughts on setting these automatically `if defined?(Turbo)`, that way it's automatic for new users and they don't have to modify the config at all?

carlosantoniodasilva: [New users should get the devise initializer setup like that](https://github.com/heartcombo/devise/pull/5548/files#diff-1a65f2d7525a89107811cc734b24f8c92d7a7e04d0265b49c20f97fd928c4f68R299-R306), so they wouldn't have to do any extra step other than running the install generator; however existing users/apps will need to opt-in via the config, so upgrading Devise won't cause any behavior change until they add that config explicitly.

Were you thinking about that use-case? I guess we could do it based on some `defined?(Turbo)` check of sorts, but it _could_ change some unwanted behavior for someone upgrading in that case (and I still want to make sure new users always get the new behavior, whether or not they have Turbo) -- trying to think it through here if there's any scenario, for example they have turbo in their app but have explicitly disabled turbo in Devise forms (because it doesn't work yet), now upgrading Devise could change the responses behavior (because `defined?(Turbo)` is true) and possibly break some client expectation?

excid3: Ah! I missed the initializer change. I think this is the safer approach.

Some existing apps might use a SPA frontend and expect certain response codes, so changing that would be bad for existing apps.

ghiculescu: Yeah I agree this approach is sensible.

Given this is a fairly big change and it's not automatic (so people targeting https://github.com/heartcombo/devise/pull/5340 will need to make this change when they update), what do you think adding a `post_install_message` to the Gemspec to draw people's attention to it?

excid3:
@carlosantoniodasilva Just tested this out on our Hotwire app and it works like a charm! 👍

Haven't tested with a fresh Rails 7 app or an old UJS app yet, but will do shortly.

JunichiIto:
@carlosantoniodasilva Thank you for your great work! I tested this version against both Rails 6 (rails-ujs) and Rails 7 (Turbo) apps and confirmed they worked well 👍 

My concern is sign-in link for OmniAuth. You might need to change it to button and disable Turbo as I discussed [here](https://github.com/heartcombo/devise/pull/5545#discussion_r1093033194).

carlosantoniodasilva:
@excid3 that's awesome to hear, thanks!

@JunichiIto awesome too! Thanks for testing it out on both apps.

> My concern is sign-in link for OmniAuth. You might need to change it to button and disable Turbo as I discussed https://github.com/heartcombo/devise/pull/5545#discussion_r1093033194.

Yeah, I hadn't checked on that yet, but you're right about the CORS issue, I'll update this branch to use `button_to` to work around that for both cases. :+1:

CHANGELOG.md
carlosantoniodasilva: > what do you think adding a post_install_message to the Gemspec to draw people's attention to it?

I think we can add a temporary message there for this new release. I haven't been much a fan of those messages popping up on every gem install, but this one might be worth an exception.

carlosantoniodasilva: Ok, I just pushed a post install message, and added a [wiki](https://github.com/heartcombo/devise/wiki/How-To:-Upgrade-to-Devise-4.9.0-[Hotwire-Turbo-integration]) with some additional information for upgrading (basically expanding what's already on the changelog, but it makes it easier to update / tweak if needed). It's a little annoying because I can't enforce the responders version just yet, but hopefully that's clear enough.

Let me know if you can think of anything else, otherwise I think we're mostly done. Thanks!

JunichiIto:
In Changelog or wiki, it might be worth mentioning that sign-out link requires `data: { turbo_method: :delete }` for Turbo. Developers might not notice this change because sign-out link or button isn't generated by `devise install`.

```erb
<%# works with rails-ujs but does not work with Turbo %>
<%= link_to 'Sign out', destroy_user_session_path, method: :delete %>

<%# works with rails-ujs and Turbo %>
<%= link_to 'Sign out', destroy_user_session_path, method: :delete, data: { turbo_method: :delete } %>
<%= button_to 'Sign out', destroy_user_session_path, method: :delete %>
```

app/views/devise/registrations/edit.html.erb
```diff
@@ -38,6 +38,6 @@
 
 <h3>Cancel my account</h3>
 
-<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?" }, method: :delete %></p>
+<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?", turbo_confirm: "Are you sure?" }, method: :delete %></p>
```
JunichiIto, line 41: In Changelog or Wiki, it might be worth mentioning this change like OmniAuth "Sign in with" links.

app/views/devise/shared/_error_messages.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if resource.errors.any? %>
-  <div id="error_explanation">
+  <div id="error_explanation" data-turbo-cache="false">
```
JunichiIto, line 2: Sorry, I'm not sure why this change is required. I couldn't find any strange behavior even if I remove this option. Could you explain a bad scenario when we remove it?

JunichiIto:
https://github.com/heartcombo/devise/wiki/How-To:-Upgrade-to-Devise-4.9.0-%5BHotwire-Turbo-integration%5D

> OmniAuth "Sign in with" links where implemented using method: :post in order to send POST requests to the server that would then redirect to the OmniAuth provider

I feel "OmniAuth provider" should be "OAuth provider" in this context. Correct?

app/views/devise/registrations/edit.html.erb
kwtuku, line 41: It may need to be changed.
If @hotwired/turbo version is lower than 7.2.0-beta.1, confirmation does not work.

```suggestion
<p>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?" }, form: { data: { turbo_confirm: "Are you sure?" } }, method: :delete %></p>
```

https://github.com/hotwired/turbo-rails/issues/302
> I've tried this :
> 
> <%= button_to "Sign out", logout_path, method: :delete, data: { turbo_confirm: "Are you sure?" } %>
>
> but nothing works.

This issue fixed at hotwired/turbo v7.2.0-beta.1
https://github.com/hotwired/turbo/releases/tag/v7.2.0-beta.1
> Fix data-turbo-confirm to work with multiple submitters inside a form by @feliperaul in https://github.com/hotwired/turbo/pull/564

JunichiIto: Turbo-rails uses Turbo 7.2.0 since [v1.3.0](https://github.com/hotwired/turbo-rails/releases/tag/v1.3.0).

So I think this compatibility issue could be solved by adding this line.

```ruby
gem "turbo-rails", "~> 1.3"
```

app/views/devise/shared/_error_messages.html.erb
carlosantoniodasilva: It is not strictly required for the functionality. I was noticing an issue when going to edit the user, submitting a couple of times, then using the `Back` button (which is a javascript history.back() call), it was flashing the errors from the cached page before getting the response back from the server and displaying it:

https://user-images.githubusercontent.com/26328/217253701-6fa82752-f0b7-4503-b940-e061e087d526.mp4

So skipping the cache of this errors section seems to have "fixed" that for me. Makes sense?

app/views/devise/registrations/edit.html.erb
carlosantoniodasilva: Thanks for the info, glad to know this is no longer an issue. I guess we can keep it in the `data` attribute then, instead of the `form data`, and make a note about the compatibility in the wiki for example. Any new apps will be generated with the latest turbo so they shouldn't run into any issue.

carlosantoniodasilva:
@JunichiIto thanks for the feedback. I believe I have covered it all with the last commit, updating the changelog/readme (and the wiki) with some information about confirm/method changes for Turbo.

> In Changelog or wiki, it might be worth mentioning that sign-out link requires data: { turbo_method: :delete } for Turbo. Developers might not notice this change because sign-out link or button isn't generated by devise install.

Good call, I added some info about the sign out link/button. 

> I feel "OmniAuth provider" should be "OAuth provider" in this context. Correct?

I am removing `OmniAuth` and just leaving `provider` there for simplicity.

app/views/devise/registrations/edit.html.erb
carlosantoniodasilva: I added a note to the wiki: https://github.com/heartcombo/devise/wiki/How-To:-Upgrade-to-Devise-4.9.0-%5BHotwire-Turbo-integration%5D#data-confirm-vs-data-turbo-confirm (at the bottom of that section):
Screenshot 2023-02-07 at 11 08 24

app/views/devise/shared/_error_messages.html.erb
JunichiIto: > it was flashing the errors

Thank you for the explanation. It's hard to notice but I understand🙂

JunichiIto:
> updating the changelog/readme (and the wiki) with some information about confirm/method changes for Turbo.

Thank you for the update. It's perfect!

JunichiIto:
Just FYI, I'm translating and updating the upgrade guide in Japanese.

https://gist.github.com/JunichiIto/1398835b9825a9ce3958456041f59d27

BilalBudhani:
Just tried this out in a project. Works fine 👍

excid3:
🎉 Thanks @carlosantoniodasilva! 🎉

app/views/devise/registrations/edit.html.erb
lapser: This seems do not solve #5448. See here: https://github.com/heartcombo/devise/issues/5448#issuecomment-1457338283

PR #5554, opened 2023-02-11
Remove p tag since p tags cannot contain other block elements

JunichiIto:
Fix https://github.com/heartcombo/devise/issues/5553

HTML and UI will be below:
fixed


carlosantoniodasilva (commented):
Ah right, good catch, I always forget the whole `<p> tags can't contain other block elements` thing, and the fact that `button_to` generates a `form`.

I do wonder if we should at least keep it with a wrapper like a `<div>`... not that I feel super strong about it, but it feels a bit weird to have loose text in there for some reason... basically what you suggested as alternative in the issue:

> So it should be fixed like this:
> 
> ```erb
> Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?", turbo_confirm: "Are you sure?" }, method: :delete %>
> ```
> 
> or
> 
> ```erb
> <div>Unhappy? <%= button_to "Cancel my account", registration_path(resource_name), data: { confirm: "Are you sure?", turbo_confirm: "Are you sure?" }, method: :delete %></div>
> ```

I think we should do that. Would you mind amending the commit swapping the `<p>` with a `<div>`? Thanks!

JunichiIto:
@carlosantoniodasilva Thank you for your reply. I agree with you. I replaced p with div.

carlosantoniodasilva (approved):
Awesome, thanks!

Issue #4823, opened 2018-03-30
Wrong locale (I18n) in failure_app 


Yegorov:
## Environment

- Ruby **2.4.1**
- Rails **5.1.6**
- Devise **4.4.3**

## Current behavior

I set locale in my ApplicationController, as this:
```ruby
class ApplicationController < ActionController::Base
  before_action :set_locale
  # other code
  def set_locale
    I18n.locale = :ru
  end
end
```
And if i submit wrong email with password to Users::SessionsController#create, flash mesage has :en locale.

For example, i monkey patched add `binding.pry` to [this](https://github.com/plataformatec/devise/blob/master/lib/devise/failure_app.rb#L109), and i get in console:

```ruby
    376:     def i18n_message(default = nil)
    377:       message = warden_message || default || :unauthenticated
    378: 
    379:       if message.is_a?(Symbol)
    380:         options = {}
    381:         options[:resource_name] = scope
    382:         options[:scope] = "devise.failure"
    383:         options[:default] = [message]
    384:         auth_keys = scope_class.authentication_keys
    385:         keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class.human_attribute_name(key) }
    386:         options[:authentication_keys] = keys.join(I18n.translate(:"support.array.words_connector"))
    387:         options = i18n_options(options)
 => 388: binding.pry
    389:         I18n.t(:"#{scope}.#{message}", options)
    390:       else
    391:         message.to_s
    392:       end
    393:     end

[1] pry(#<Devise::FailureApp>)> I18n.locale
=> :en

```
What could be wrong?

Other translate flash message in devise work ok!

## Expected behavior

I18n.locale must be one that is set to ApplicationController#set_locale

Yegorov:
Hi @tegon. Thanks for your response!
If this matters, then I use [Rack::Locale middleware](https://github.com/rack/rack-contrib/blob/master/lib/rack/contrib/locale.rb) to determine the current user locale.
I modified your [test application (i18n-failure-app_2.zip)](https://github.com/plataformatec/devise/files/1886285/i18n-failure-app_2.zip).
To reproduce the error you need repeatedly change the locale on the main page, go to the login page and enter the wrong data.
Some screenshots for proof:
![_478](https://user-images.githubusercontent.com/2566462/38454089-f20d76fc-3a68-11e8-8797-19a6c20ff7ea.png)
[...]

Issue #3052, opened 2014-06-04
Weird i18n issues


thomasdarde:
I've been searching for a solution.
Sometimes the flash messages is displayed in another locale then I18n.locale.

Any suggestion ?

gingerlime:
Looks like changing the middleware position for dealing with locales can resolve this. See https://github.com/plataformatec/devise/issues/4823#issuecomment-388112897

PR #5567, opened 2023-03-07
Respect locale set by controller in the failure app

### LINKED ISSUE #5246: Devise translations not working.
## Environment

- Ruby **[2.1.2]**
- Rails **[4.2.7.1]**
- Devise **[3.5.10]**

## Current behavior

So i finished with the translations of my project with i18n. For the devise messages, i just created a new file called, devise.es.yml, and placed there all the translations in spanish. The rest of my project is working, but, with the devise i have some weird behaviour, because sometimes i get messages in english when im in spanish and viceversa. So, i have some view that renders the errors messages in the login page:

```
<% flash.each do |name, msg| %>
  <% if msg.is_a?(String)%>
    <div class="alert alert-<%= name.to_sym == :notice ? 'success' : 'danger' %>">
      <a class="close" data-dismiss="alert">&#215;</a>
      <%= content_tag :div, content_tag(:ul, msg.html_safe, class: "clear"), :id => "flash_#{name}" %>
    </div>
  <% elsif msg.is_a?(Array) %>
    <div class="alert alert-<%= name.to_sym == :notice ? 'success' : 'danger' %>">
      <a class="close" data-dismiss="alert">&#215;</a>
      <%= content_tag :div, content_tag(:ul, "<li>#{msg.join('</li><li>')}</li>".html_safe, class: "clear"), :id => "flash_#{name}" %>
    </div>
  <% end %>
<% end %>
```

So, even if the language in that view is spanish, sometimes, i already have messages in english inside the flash variable. The thing is, i imagined there would be somewhere in the project where that variable is being setted, with messages like t('devise.failure.unauthenticated'), but i saw this is nowhere in the project. So i dont know what is the problem. I have messages that i dont know where they are being assigned to the flash variable, and from what i read, to translate things from devise, it is just creating the other file devise.es.yml.

image

There, everything is correct except for the message error, which is in english, and im having a hard time figuring out why this happens.

Sorry if my problem is not clear, english is not my first language. Thanks to whoever can help me.

### LINKED ISSUE #5247: Devise uses the wrong locale for some flash messages
## Environment

- Ruby 2.6
- Rails 6.0
- Devise 4.7

## Current behavior

Currently I use this code as suggested by the Rails guide:

```
  # application_controller.rb

  around_action :switch_locale

  def switch_locale(&action)
    locale = params[:locale] || I18n.default_locale
    I18n.with_locale(locale, &action)
  end
```

This works perfectly for the application... but not for all Devise messages! Some are translated, others are not. Devise keeps using the default language for some messages (even if the translated YML files are present).

Possibly related:
#145
https://github.com/heartcombo/devise/issues/5246

In order to reproduce the issue:

1. try login with an invalid password
2. you get an error message (flash) in the wrong language (i.e. Devise seems to use the `default_locale` set in the environment files, not the current one set in application controller)

If you log in successfully, instead, the flash message is in the correct language (the language set in application controller).

I have also tried to copy the official YML file from Devise and translate it, but I have the same issue: some translations appear, other do not.

## Expected behavior

Devise should always use the current locale.

carlosantoniodasilva:
A common usage of I18n with different locales is to create some around callback in the application controller that sets the locale for the entire action, via params/url/user/etc., which ensure the locale is respected for the duration of that action, and resets at the end.

Devise was not respecting the locale when the authenticate failed and triggered the failure app, because that happens in a warden middleware right up in the change, by that time the controller around callback had already reset the locale back to its default, and the failure app would just translate flash messages using the default locale.

Now we are passing the current locale down to the failure app via warden options, and wrapping it with an around callback, which makes the failure app respect the set I18n locale by the controller at the time the authentication failure is triggered, working as expected. (much more like a normal controller would.)

I chose to introduce a callback in the failure app so we could wrap the whole `respond` action processing rather than adding individual `locale` options to the `I18n.t` calls, because that should ensure other possible `I18n.t` calls from overridden failure apps would respect the set locale as well, and makes it more like one would implement in a controller. I don't recommend people using callbacks in their own failure apps though, as this is not going to be documented as a "feature" of failures apps, it's considered "internal" and could be refactored at any point.

It is possible to override the locale with the new `i18n_locale` method, which simply defaults to the passed locale from the controller.

Closes #5247
Closes #5246

Related to: #3052, #4823, and possible others already closed. Related to warden: (may be closed there afterwards) https://github.com/wardencommunity/warden/issues/180 https://github.com/wardencommunity/warden/issues/170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Unreleased
+
+* bug fixes
+  * Respect locale set by controller in failure app. Devise will carry over the current I18n.locale option when triggering authentication, and will wrap the failure app call with it. [#5567](https://github.com/heartcombo/devise/pull/5567)
+
 ### 4.9.3 - 2023-10-11
 
 * enhancements
--- a/app/controllers/devise/sessions_controller.rb
+++ b/app/controllers/devise/sessions_controller.rb
@@ -45,7 +45,7 @@ class Devise::SessionsController < DeviseController
   end
 
   def auth_options
-    { scope: resource_name, recall: "#{controller_path}#new" }
+    { scope: resource_name, recall: "#{controller_path}#new", locale: I18n.locale }
   end
 
   def translation_scope
--- a/lib/devise/controllers/helpers.rb
+++ b/lib/devise/controllers/helpers.rb
@@ -46,6 +46,7 @@ module Devise
                 mappings.unshift mappings.delete(favorite.to_sym) if favorite
                 mappings.each do |mapping|
                   opts[:scope] = mapping
+                  opts[:locale] = I18n.locale
                   warden.authenticate!(opts) if !devise_controller? || opts.delete(:force)
                 end
               end
@@ -115,6 +116,7 @@ module Devise
         class_eval <<-METHODS, __FILE__, __LINE__ + 1
           def authenticate_#{mapping}!(opts = {})
             opts[:scope] = :#{mapping}
+            opts[:locale] = I18n.locale
             warden.authenticate!(opts) if !devise_controller? || opts.delete(:force)
           end
 
--- a/lib/devise/failure_app.rb
+++ b/lib/devise/failure_app.rb
@@ -18,6 +18,11 @@ module Devise
 
     delegate :flash, to: :request
 
+    include AbstractController::Callbacks
+    around_action do |failure_app, action|
+      I18n.with_locale(failure_app.i18n_locale, &action)
+    end
+
     def self.call(env)
       @respond ||= action(:respond)
       @respond.call(env)
@@ -107,13 +112,17 @@ module Devise
         options[:default] = [message]
         auth_keys = scope_class.authentication_keys
         keys = (auth_keys.respond_to?(:keys) ? auth_keys.keys : auth_keys).map { |key| scope_class.human_attribute_name(key) }
-        options[:authentication_keys] = keys.join(I18n.translate(:"support.array.words_connector"))
+        options[:authentication_keys] = keys.join(I18n.t(:"support.array.words_connector"))
         options = i18n_options(options)
 
         I18n.t(:"#{scope}.#{message}", **options)
       else
         message.to_s
       end
+    end
+
+    def i18n_locale
+      warden_options[:locale]
     end
 
     def redirect_url
--- a/test/controllers/helpers_test.rb
+++ b/test/controllers/helpers_test.rb
@@ -64,30 +64,30 @@ class ControllerAuthenticatableTest < Devise::ControllerTestCase
   end
 
   test 'proxy authenticate_user! to authenticate with user scope' do
-    @mock_warden.expects(:authenticate!).with({ scope: :user })
+    @mock_warden.expects(:authenticate!).with({ scope: :user, locale: :en })
     @controller.authenticate_user!
   end
 
   test 'proxy authenticate_user! options to authenticate with user scope' do
-    @mock_warden.expects(:authenticate!).with({ scope: :user, recall: "foo" })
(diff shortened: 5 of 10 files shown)


lapser:
Please commit/merge this asap if it works.

carlosantoniodasilva:
@lapser thanks. I plan to get it merged soon, just leaving it sit here for a bit longer in case someone from the issues I pinged previously wanted to give it a try first. If you do test it out, let me know how it goes.

ejohnson6:
I was having the same issue as https://github.com/heartcombo/devise/issues/5246 and tried out this branch and confirmed that it fixes the issue for me

lillieatwork:
This would be wildly helpful!! ❤️ 🚢 :shipit: ❤️

hientranea:
Any updates on this PR? I had the same problem and the PR fixes the issue for me 🚀

mark-young-atg:
For me this works when an incorrect password is provided, however it does not work when I logout of the application. What I am seeing is that when `redirect` is called in failure_app.rb:86 the i18n_message method is returning 'You need to sign in or sign up before continuing.'. 

The step before that when the destroy method in sessions_controller.rb:27 is being run, I can see that I18n.locale = :de and the message produced for signing out is "Erfolgreich abgemeldet.". Then it redirects to /, at which point the locale set in the controller is lost.

Is this a scenario you expect to cover with this PR?

mark-young-atg:
Hi @carlosantoniodasilva 
I have created a demonstration of the problem I'm referring [in github](https://github.com/mark-young-atg/devise_test)
I hope it clarifies my previous question.
I would like the language used during the sign out processing to be also used when the new login page is displayed.
All the best,
Mark

mark-young-atg:
I have removed my previous comments. This branch does work as expected. The problem I saw is elsewhere.

leonardow-unep-wcmc:
Having the same issue and this PR works for me.
Rails 7.0.5.1 API only app
Devise 4.2.9

ejohnson6:
Hey just checking on this again, hopping to see it get in soon! Any progress updates would be appreciated! 😃

carlosantoniodasilva:
Sorry folks, I know it's been taking some time, but I had to step away from things for personal reasons in the past couple of months. I'm slowly catching up now, and have this as one of the main things on my radar to look at. :soon:

mark-young-atg:
Hi @carlosantoniodasilva 
I trust you are well. I see you are getting active again with devise. Is there any chance of getting this PR merged in as part of v4.9.3?
All the best,
Mark

carlosantoniodasilva:
@mark-young-atg :wave: thanks.

v4.9.3 already landed, but I'll get this merged and released as part of the next version.

carlosantoniodasilva:
This is in 4-stable and main now.

lillieatwork:
Thank you for all your work on this project!! 💖 💖 💖

yuhiyone:
@carlosantoniodasilva 
Thank you for your work. I am also facing the same problem with my app. I'm looking forward to a new issue.

mark-young-atg:
Now this work has been merged into `main` is it possible to produce a new release of devise please?

RavWar:
Maybe it's worth noting, if current_user is called anywhere in before action then this fix is not working, sign_in goes straight to warden and sessions_controller#create action along with auth_options are not called. Combining with this fix: https://github.com/heartcombo/devise/issues/5602#issuecomment-1876164084 - everything works as expected

carlosantoniodasilva:
@RavWar the ideal solution is to make sure you're always setting the locale first, in a prepended before action / around action. That other change you pointed out can be used, but it has more side-effects that are worth being aware of.

@mark-young-atg I'll look into cutting a new release, thanks.

carlosantoniodasilva:
For folks waiting on a release: I just released v4.9.4 including this fix.

PR #5568, opened 2023-03-08
Bring back Mongoid official support

carlosantoniodasilva:
Devise hasn't been tested with Mongoid since Rails version 5, only 4.x was still running those tests.

This enables the tests again on all currently supported Rails versions, with their respective mongoid supported versions. There were a couple of minor tweaks to make it happen, namely:

* The way we were dropping the session before doesn't work in later versions so I changed back to calling `purge!` which appears to work fine. We used to call `Mongoid.purge!` but that changed in #4686.
* Some of the configs in the Rails test app were setting Active Record values when outside of the AR ORM tests, updated those to make sure they are not set when running mongoid ORM tests.
* The validations added to the shared admin code in tests were only checking for Rails version 5.1, but we need to use the same check for AR 5.1 that is used in code, otherwise it will try to use methods not available in mongoid there.

I also had to lock the sqlite3 gem on Rails 6 versions to 1.5.x, due to it dropping support to Ruby 2.6 on v1.6+ now, and those Rails 6 versions need to support that Ruby version.

Finally, I locked mongoid to v7.5+ but not allowing v8+ yet, there seems to be some change to dirty attributes that I need to look into which is making a handful of tests fail. For now, this should hopefully get us to green as step 1.


jclusso:
> Devise hasn't been tested with Mongoid since Rails version 5, only 4.x was still running those tests.

I'm floored that it still works based on this statement. Glad to see this PR getting updated. I would've never known Mongoid was not officially supported at this point since the README has it mentioned all over the place.

Issue #5579, opened 2023-03-30
Argon2 encryptor


RobertDober:
Hello there

**This would be better as a PR**, but as it is based on a tag (which we need) that does not work, so please forgive me to discuss thus here:

https://github.com/heartcombo/devise/compare/v4.8.1...RobertDober:argon2-4.8.1-draft?expand=1

This is (yet) more a question then an implementation, although preliminary tests with our application are promising. 

AAMOF my organization is pushed by one of our clients to not use bcrypt but argon2 for password hashing. Although we are perfectly happy with bcrypt we have a fork running with Argon2 replacing Bcrypt.

But we have no intention to maintain a fork of devise, so we thought, that maybe you would be interested in integrating argon2 and I have implemented it as can be seen above.

Of course the argon2 dependency could be removed from the PR by making the encryptor API public which is just the two module methods digest and compare.
OTOH I wanted to share the Argon code too if you were interested.

I would appreciate any feedback in oreder to avoid any unnecessary waste of time, if however you are interested our organization would be most happy to give back to devise which is my second preferred José Valim project, (sorry Ruby) you can guess which one is the first ;).

The PR would then be against `main` or any preferred branch, we would love to have it in an update of 4.8.x too of course because not yet Rails 7 :sob:

Thank you in advance and for devise

Robert

carlosantoniodasilva:
@RobertDober sorry about the delay, I'm replaying here on the issue to keep the discussion going.

Appreciate your work on implementing an Argon2 compatible encrypting with Devise. I myself am not much familiar with Argon2 yet, have to do some catching up there.

The way we generally setup different encryptors with Devise is via a separate [devise-encryptable](https://github.com/heartcombo/devise-encryptable) library. Up until now all the encryption libs have been older encryptors like SHA & other compatibility things, so it might need some tweaks for better compatibility with something like Argon2 (for example it sort of requires a separate "password salt" since all of them currently do, but bcrypt doesn't, and as far as I can tell argon doesn't either.)

In any case, ideally any new encryption mechanism that's not the default bcrypt should live there. If we implement something in Devise again it will conflict with that infrastructure so it'd have to be taken into account, which is why it might be preferable to keep those there for now.

RobertDober:
@carlosantoniodasilva thank you for your reply. I had been looking into devise-encryptable but did not understand it at first.
I will be crafting out a PR today.

Até logo

RobertDober:
[...]

PR #5581, opened 2023-04-03
Make encryptor pluggable

RobertDober:
This is an alternative solution to #5579 where I sketched out an Argon2 Encryptor inside Devise.

If that would interest you I can gladly make another  PR. However here I only make the encryptor pluggable which would be of great help for my organisation as we are under heavy pressure (which I fail to understand) form the customer to use Argon2.


RobertDober:
@carlosantoniodasilva so shall we close this, as it replicates the devise-encryptable gem?

carlosantoniodasilva:
Yes, let's close it for now. Feel free to reference it from devise-encryptable if you'd like. Thanks.

PR #5583, opened 2023-04-06
Use a dedicated ActiveSupport::Deprecation

etiennebarrie:
Rails 7.1 will deprecate using the singleton ActiveSupport::Deprecation instance (https://github.com/rails/rails/pull/47354). This defines one for the gem and uses it.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -521,8 +521,12 @@ module Devise
     res == 0
   end
 
+  def self.deprecator
+    @deprecator ||= ActiveSupport::Deprecation.new("5.0", "Devise")
+  end
+
   def self.activerecord51? # :nodoc:
-    ActiveSupport::Deprecation.warn <<-DEPRECATION.strip_heredoc
+    deprecator.warn <<-DEPRECATION.strip_heredoc
       [Devise] `Devise.activerecord51?` is deprecated and will be removed in the next major version.
       It is a non-public method that's no longer used internally, but that other libraries have been relying on.
     DEPRECATION
--- a/app/helpers/devise_helper.rb
+++ b/app/helpers/devise_helper.rb
@@ -4,7 +4,7 @@ module DeviseHelper
   # Retain this method for backwards compatibility, deprecated in favor of modifying the
   # devise/shared/error_messages partial.
   def devise_error_messages!
-    ActiveSupport::Deprecation.warn <<-DEPRECATION.strip_heredoc
+    Devise.deprecator.warn <<-DEPRECATION.strip_heredoc
       [Devise] `DeviseHelper#devise_error_messages!` is deprecated and will be
       removed in the next major version.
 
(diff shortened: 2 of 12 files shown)


RailsCod3rFuture:
Hope to see these depreciation issues fixed before Rails 7.1 official release. I was playing around with MRSK. Doesn't seem plausible just yet.

carlosantoniodasilva:
> I was playing around with MRSK. Doesn't seem plausible just yet.

@RailsCod3rFuture I'm not sure how this relates to the changes here? You mean the deprecations are causing you trouble with MRSK?

RailsCod3rFuture:
> > I was playing around with MRSK. Doesn't seem plausible just yet.
> 
> @RailsCod3rFuture I'm not sure how this relates to the changes here? You mean the deprecations are causing you trouble with MRSK?

The rails server doesn't start on rails 7.1 as a result of the ActiveSupport::Deprecation instance

RailsCod3rFuture:
> @RailsCod3rFuture so I'd guess this is not a Devise problem, but an MRSK problem? I mean, I see no reason why the deprecations would prevent Devise from working. (I'm not saying we aren't gonna fix it, just mentioning it shouldn't break stuff like that imo.)

This may be different - but I just started receiving it.
```
/bin/zsh -c "bash -c 'env RBENV_VERSION=3.1.2 /usr/local/Cellar/rbenv/1.2.0/libexec/rbenv exec ruby /Users/dwood/RubymineProjects/hairola_app/bin/rails server -b 0.0.0.0 -p 3000 -e development'"
DEPRECATION WARNING: DeprecatedConstantAccessor.deprecate_constant without a deprecator is deprecated (called from <module:Authenticatable> at /Users/dwood/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/devise-4.8.1/lib/devise/models/authenticatable.rb:65)
/Users/dwood/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/bootsnap-1.10.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:15:in `require': cannot load such file -- polyamorous/activerecord_7.1_ruby_2/join_association (LoadError)
```

RailsCod3rFuture:
> I think you're just seeing the deprecation just above your real issue with [polyamorous](https://rubygems.org/gems/polyamorous).

Good catch. I'll probably have to reach out to the ransack team on that note....see if they have any solutions or knowledge of this issue.

lib/devise.rb
```diff
@@ -521,8 +521,12 @@ def self.secure_compare(a, b)
     res == 0
   end
 
+  def self.deprecator
```
rafaelfranca, line 524: We need to register the deprecator in the application deprecator. We want users to be able to silence the deprecations in their apps, or change the behavior to use notifications for example. Can you do that?

RailsCod3rFuture: I'll look into over the weekend when have some more free time.

etiennebarrie: Done!

PR #5584, opened 2023-04-09
Allow `password_length: nil (or false)` to skip password length validation

johnnyshields:
I have custom validation logic for password length, and as such I'd like to skip the default validation from Devise. Please let me know if this patch would be accepted, and if so, I'll add tests.


colinross:
If you are implementing your own validations, why are you including `:validatable` in the devise config?

PR #5598, opened 2023-05-26
Removed deprecations warning output for `Devise::Models::Authenticatable::BLACKLIST_FOR_SERIALIZATION`

soartec-lab:
This deprecation warning worked for us, but it has served its purpose.
It was added over 2 years ago now, isn't that long enough to trigger a deprecation warning?

https://github.com/heartcombo/devise/blob/main/CHANGELOG.md#480---2021-04-29

Now that I've given it enough time, I've removed this as I think it's noise for many people who don't see `Devise::Models::Authenticatable::BLACKLIST_FOR_SERIALIZATION`.

However, `Devise::DeprecatedConstantAccessor`, which was added in the same commit, is not referenced anywhere else, but has not been removed as it may be used in the future. Please let me know so I can delete it if necessary.

These source codes were added in the PR below:

https://github.com/heartcombo/devise/commit/0c2cab7c946e0796c673a36aebba7c0352e5fec8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Unreleased
 
-
+* enhancements
+  * Removed deprecations warning output for `Devise::Models::Authenticatable::BLACKLIST_FOR_SERIALIZATION` (@soartec-lab)
 
 ### 4.9.2 - 2023-04-03
 
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -61,9 +61,6 @@ module Devise
         :last_sign_in_ip, :password_salt, :confirmation_token, :confirmed_at, :confirmation_sent_at,
         :remember_token, :unconfirmed_email, :failed_attempts, :unlock_token, :locked_at]
 
-      include Devise::DeprecatedConstantAccessor
-      deprecate_constant "BLACKLIST_FOR_SERIALIZATION", "Devise::Models::Authenticatable::UNSAFE_ATTRIBUTES_FOR_SERIALIZATION", deprecator: Devise.deprecator
-
       included do
         class_attribute :devise_modules, instance_writer: false
         self.devise_modules ||= []
--- a/test/models/serializable_test.rb
+++ b/test/models/serializable_test.rb
@@ -31,12 +31,6 @@ class SerializableTest < ActiveSupport::TestCase
     assert_key "username", @user.as_json({ only: :username, except: [:email].freeze }.freeze)["user"]
   end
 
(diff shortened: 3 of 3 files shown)


ghiculescu:
Could we get a new release for this?

ghiculescu:
It means you get a deprecation warning when you boot an app running Rails 7.1 (alpha). So not a problem, just a little nuisance.

ghiculescu:
This is the warning:

```
DEPRECATION WARNING: DeprecatedConstantAccessor.deprecate_constant without a deprecator is deprecated (called from <module:Authenticatable> at /home/ruby/.rvm/gems/ruby-2.7.7/gems/devise-4.9.2/lib/devise/models/authenticatable.rb:65)
```

So I think it's actually the absence of https://github.com/heartcombo/devise/pull/5583 that is causing it.

soartec-lab:
@ghiculescu @carlosantoniodasilva 

Hi.

This happens by calling `ActiveSupport::Deprecation.instance` with the code below:

https://github.com/soartec-lab/devise/blob/master/lib/devise/rails/deprecated_constant_accessor.rb#L29

This is because from `rails 7.1`, `ActiveSupport::Deprecation.instance` has been deprecated and changed to issue a warning. Please check below:

https://github.com/rails/rails/pull/47354

The solution is to change `lib/devise/rails/deprecated_constant_accessor.rb` to the latest source code.
However, this file is a process to maintain backward compatibility, and backward compatibility support for older `rails` versions will be removed in the following PR.

https://github.com/heartcombo/devise/pull/5600

I think it will probably be removed by the next major update.

carlosantoniodasilva:
@ghiculescu gotcha, makes sense. That may or may not be easier to handle (in terms of a new release that's backwards compatible)

@soartec-lab thanks.

soartec-lab:
@carlosantoniodasilva 

OK, If there is no prospect of responding to the PR below, I will create a PR to update [this file](https://github.com/soartec-lab/devise/blob/master/lib/devise/rails/deprecated_constant_accessor.rb#L29) to the latest version. What do you think?

https://github.com/heartcombo/devise/pull/5600

carlosantoniodasilva:
@soartec-lab if all we need is to update `ActiveSupport::Deprecation.instance` to `Devise.deprecator` there, based on https://github.com/heartcombo/devise/pull/5583, then yeah that sounds good to me, sounds like it was just missed there.

soartec-lab:
@carlosantoniodasilva 

I have the same opinion. It looks like there was an update missing, so I'll create a new PR and fix it.

soartec-lab:
@carlosantoniodasilva 

I corrected this issue with the PR below. Could you review?

https://github.com/heartcombo/devise/pull/5628

PR #5600, opened 2023-06-09
Drop support to EOL rubies and Rails

rafaelfranca:
@carlosantoniodasilva I'm not sure if you were keeping support to those old versions by design, but maybe it is time for us to let it go. All tests on Ruby < 2.7 were failing.


RailsCod3rFuture (approved):

carlosantoniodasilva:
Thanks @rafaelfranca! I was hesitant to drop support and have to go to a 5x release, but it's probably time, there's a few other potential breaking changes we can then include to move things forward.

I've branched out a `4-stable` to release any bug fixes as needed, so let's make main target a 5x release and drop these old versions. I added a couple extra commits, but mainly I wanted to keep support to Ruby 2.7, even though it's EOL I think it makes sense to support it for a while longer to keep better compatibility with Rails. We can drop it at a later date.

PR #5621, opened 2023-08-17
Inherit from only base.

santu-essence:


carlosantoniodasilva:
Thanks. There's some issues with the build that I'm looking into, but they're unrelated.

PR #5623, opened 2023-08-22
update readme file for generate devise model usage

junminhong:
### env
`rails version: 6.1.6.1`
`ruby version: 2.7.4`

I tried using the readme file init, but when I do `rails generated devise MODEL`.
I get this problem
截圖 2023-08-22 下午4 05 22

so this PR is to update the README and help others not to have the same problem


junminhong:
oh... I got mixed up

PR #5628, opened 2023-09-12
Fixed missing migration to dedicated deprecator

soartec-lab:
This is a follow-up to the PR below:

- https://github.com/heartcombo/devise/pull/5583
- https://github.com/heartcombo/devise/pull/5598

`rails` has deprecated the use of `ActiveSupport::Deprecation` in singleton in [this PR](https://github.com/rails/rails/pull/47354/files), so `devise` has defined its own `Deprecator` in response to this, but there was an omission in the correction. So I fixed it.

Please note that a PR has been created that will discontinue support for EOL rails versions, and the files modified this time may also be deleted.

https://github.com/heartcombo/devise/pull/5600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * enhancements
   * Removed deprecations warning output for `Devise::Models::Authenticatable::BLACKLIST_FOR_SERIALIZATION` (@soartec-lab)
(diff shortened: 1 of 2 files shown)


soartec-lab:
I think the failing tests will be fixed in https://github.com/heartcombo/devise/pull/5622

carlosantoniodasilva:
Thanks!

PR #5629, opened 2023-09-14
Bangla translation

ajpalok:
Translation added for Bengali Language.


carlosantoniodasilva:
Thank you, but we don't maintain translations in devise itself. Please check the wiki for other options: https://github.com/heartcombo/devise/wiki/I18n

PR #5638, opened 2023-10-09
Avoid using `with_options` for Rails 7.1.0 support

r7kamura:
I encountered the problem reported in the following Issue:

- Fix https://github.com/heartcombo/devise/issues/5637

As gagoit mentioned in the comments of that issue, I think the problem is caused by a change in the behavior of `Hash#deep_merge` used inside `Object#with_options` in Rails 7.1.0.

Simply avoiding the use of `Object#with_options` looks a good way to solve the problem, so I suggest changing it that way to support Rails 7.1.0.

I tried this change on a medium sized app that I am upgrading to Rails 7.1.0 and found it to work fine.

### DIFF
--- a/lib/devise/modules.rb
+++ b/lib/devise/modules.rb
@@ -1,30 +1,24 @@
 # frozen_string_literal: true
 
-require 'active_support/core_ext/object/with_options'
+# Strategies first
+routes = [nil, :new, :destroy]
+Devise.add_module :database_authenticatable, model: true, strategy: true, controller: :sessions, route: { session: routes }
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
@r7kamura thanks for the updates. It looks like the culprit might be something else (potentially the `config` gem?), [see the conversation there](https://github.com/heartcombo/devise/issues/5637#issuecomment-1755358970). Can you double check if that's the case for you as well?

If there's such a fundamental behavior change in `with_options` / `deep_merge`, I'd rather not change Devise to accommodate it but try to look for the potential underlying problem. If there's no other way around it, fine, but otherwise that seems the best option.

PR #5645, opened 2023-10-20
Removed now obsolete SecretKeyFinder

BroiSatse:
SecretKeyFinder was required to handle rails configuration pre 6.0 which is no longer supported. Secret key can (and should!) be now read directly from rails application.

Fixes: https://github.com/heartcombo/devise/issues/5644
Probably surpasses: https://github.com/heartcombo/devise/pull/5604


dgm:
The best bugfix is one that removes code. :)

issei-m:
Any news?

grk:
For those looking for a workaround, you can set the secret key in your `devise.rb` initializer:
```ruby
```
and you won't see the deprecation warning anymore.

jordan-brough:
Would be nice to have this published. cc @carlosantoniodasilva? 🙏

jcoyne:
Yes! I was just about to submit a similar change, but you went a step further. 👍

albus522:
It is probably worth noting somewhere that this is a potentially breaking change. As I noted in #5634, Rails and Devise use a different priority order in what they choose. For certain old app configurations this could result in the key unintentionally changing.

dan-jensen:
@albus522 good call. And we could even consider this a bugfix, because Devise shouldn't have been choosing a different key than the application. Seems like bumping the version to 4.10 and noting this breaking change in [CHANGELOG.md](https://github.com/heartcombo/devise/blob/main/CHANGELOG.md) would be sufficient.

bbuchalter:
It looks like this PR lost momentum and still addresses an open issue. What's needed to move it forward?

nashby:
Thanks everyone!

PR #5664, opened 2023-12-18
Replace [data-turbo-cache=false] with [data-turbo-temporary]

eduardoyutaka:
Following the suggestion from issue https://github.com/heartcombo/devise/issues/5662

The turbo drive's attribute `[data-turbo-cache=false]` was deprecated in favor of `data-turbo-temporary`

### DIFF
--- a/app/views/devise/shared/_error_messages.html.erb
+++ b/app/views/devise/shared/_error_messages.html.erb
@@ -1,5 +1,5 @@
 <% if resource.errors.any? %>
-  <div id="error_explanation" data-turbo-cache="false">
+  <div id="error_explanation" data-turbo-temporary>
     <h2>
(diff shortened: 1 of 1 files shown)


timdiggins:
This should work with Turbo >= 7.3.0 and won't work with Turbo < 7.3.0

Could just support latest version of turbo? (break in old Turbo)

Or stick with old turbo? (deprecation warnings on new Turbo)

Could have a turbo compatibility layer:

```erb
  <div id="error_explanation" <%= Devise::TurboCompat.gte_730? &&  'data-turbo-temporary' || 'data-turbo-cache="false"' %>>
```

Given that Turbo is a Js library, it seems tricky to find an automatic way to support different versions of turbo.  Would then have to be configured by either an env var or a devise configuration option?

PR #5668, opened 2024-01-02
Adds Ruby 3.3 to CI

berkos:
This PR adds Ruby 3.3 in the CI matrix.

It excludes rubies that are incompatible with Rails 7.2

```
Because every version of rails depends on Ruby >= 3.1.0
  and Gemfile-rails-main depends on rails >= 0,
  Ruby >= 3.1.0 is required.
  ```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,7 @@ jobs:
           - gemfiles/Gemfile-rails-6-1
           - gemfiles/Gemfile-rails-6-0
         ruby:
+          - '3.3'
           - '3.2'
           - '3.1'
           - '3.0'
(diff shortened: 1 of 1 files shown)


capripot:
Yes please!

gemfiles/Gemfile-rails-7-1
carlosantoniodasilva, line 1: 7.1 is tested from our main Gemfile, we generally don't add a new / custom one until 7.2 is released. (that's when the Gemfile moves to test the current 7.2, and we add the previous as a custom Gemfile to continue testing it)

carlosantoniodasilva (commented):
Thanks. We can add the Ruby version to the build, but need to wait a bit until 7.2 is released to add a custom 7.1 gemfile to test against.

gemfiles/Gemfile-rails-7-1
berkos: Hi, thanks for the review!
ah yes, I missed that main Gemfile is also in the matrix.
The updated commit should address it https://github.com/heartcombo/devise/pull/5668/commits/c80026d95583401a78c07428e58e50b0cf0d933b

PR #5669, opened 2024-01-19
Simplify SecretKeyFinder

jcoyne:
Rails back to 6.0 has Rails.application.secret_key_base.  We don't need all these conditionals anymore.


jcoyne:
Duplicate of #5645

PR #5679, opened 2024-03-29
Use locales on mailer views

JuanVqz:
The other day I was trying to internationalize mailers on my application. I ended up copying the views to translate the mail content, which could be done by changing the content in the local (en.yml)file.

So, that is why I think it would be nice to have the I18n keys instead of the plain text in the mailers


nashby:
@JuanVqz thank for the PR but we don't maintain translations in Devise. You can use https://github.com/devise-i18n/devise-i18n for that or just copy these mailer views into your app and do it there.

PR #5685, opened 2024-04-16
increase default password length validation to minimum 12 characters

gregmolnar:


salzig:
I like this change. It's simple and increases the security for default installations. I expect all developers to already use password managers that generate passwords that are mostly longer than 20chars. So this is only helping people to choose a more secure password not already using a password manager. 

We need to remember that there are a lot of people not using a password manager, which would really use 6 chars long passwords when possible.

fthobe:
@gregmolnar @salzig Can I say that I agree with both of you on this.

kykyi (approved):
Not a maintainer but want to show my support 😄

salzig (approved):

gregmolnar:
I just realized that this would be a breaking change, so I will rework it. Ideally when a password is updated the new length would be required, I will look into how to make that happen.

fthobe:
And just like that he almost broke the entire rails based internet 😂😂😂

gregmolnar:
> And just like that he almost broke the entire rails based internet 😂😂😂

Wouldn't be the first time, nor the last time :) 
If it would be merged to a major release it would be fine I think, but we can do better, I just need to find some time to do some changes.

fthobe:
@gregmolnar i start to have the Impression that no one is merging anything here TBH

gregmolnar:
I changed this to set the new minimum length for newly generated configs. Let's see it maintenance picks up and then I will work on rolling this out on password updates too.

Issue #5438, opened 2021-12-18
user_signed_in? in integration tests


sudoMoshe:
hey there , 

there is no helper method inside integration tests that can tell if the user is logged in or not, I tried searching for a solution in stack-overflow but the only thing I came up with is checking for changes in the layout 

I didn't give up on trying to find a better solution that checks the actual session cookie so I won't have to rely on something in the views to verify the user is logged in 

I ended up writing my own helper method inside test_helper.rb that uses the warden.authenticated? method but I really think this is 
something that should be inside the IntegrationHelpers 

anyway this is the method i ended up writing inside my test_helper.rb in case anyone would want it:

` def is_logged_in? `
  `  request.env['warden'].authenticated?(:user)`
 `end`

nashby:
Hey!

I don't see much value in having such a method for integration tests, since you're supposed to check if the user actually signed in (e.g., by checking the page content, as you mentioned). If you really need to check the request object, feel free to do it in your own helper method. Thanks!

PR #5703, opened 2024-07-31
Add helper method

CuddlyBunion341:
Added the integration test helper, as defined in #5438.


nashby:
Hey!

I don't see much value in having such a method for integration tests, since you're supposed to check if the user actually signed in (e.g., by checking the page content, as you mentioned). If you really need to check the request object, feel free to do it in your own helper method. Thanks!

PR #5727, opened 2024-11-22
Introduce config to allow for password complexity

kykyi:
In relation to https://github.com/heartcombo/devise/issues/5591

This PR introduces application config to allow for password complexity to granularly managed with new validation options for:
* presence of a lower case letter
* presence of a upper case letter
* presence of a number
* presence of a special character (from a list of special characters that is configurable)

These are all `false` by default, and configurable like:
```ruby
Devise.setup do |config|
      config.password_length = 8..128 # (existing)
      config.password_requires_lowercase = true
      config.password_requires_uppercase = true
      config.password_requires_number = true
      config.password_requires_special_character = true
      config.password_special_characters = ":)"
end
```

**Note** 
* I haven't run any linting, I couldn't find instructions on what config was used for that 😄 
* I haven't updated any docs, please advise where I need to update if at all 🙏

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -120,6 +120,26 @@ module Devise
   mattr_accessor :password_length
   @@password_length = 6..128
 
+  # Validate presence of lower case letter in password
+  mattr_accessor :password_requires_lowercase
+  @@password_requires_lowercase = false
+
+  # Validate presence of upper case letter in password
+  mattr_accessor :password_requires_uppercase
+  @@password_requires_uppercase = false
+
+  # Validate presence of special character in password
+  mattr_accessor :password_requires_special_character
+  @@password_requires_special_character = false
+
+  # Special character options
+  mattr_accessor :password_special_characters
+  @@password_special_characters = "!?@#$%^&*()_+-=[]{}|:;<>,./"
+
+  # Validate presence of a number in password
+  mattr_accessor :password_requires_number
+  @@password_requires_number = false
+
   # The time the user will be remembered without asking for credentials again.
   mattr_accessor :remember_for
   @@remember_for = 2.weeks
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -120,25 +120,15 @@ module Devise
   mattr_accessor :password_length
   @@password_length = 6..128
 
-  # Validate presence of lower case letter in password
-  mattr_accessor :password_requires_lowercase
-  @@password_requires_lowercase = false
-
-  # Validate presence of upper case letter in password
-  mattr_accessor :password_requires_uppercase
-  @@password_requires_uppercase = false
-
-  # Validate presence of special character in password
-  mattr_accessor :password_requires_special_character
-  @@password_requires_special_character = false
-
-  # Special character options
-  mattr_accessor :password_special_characters
-  @@password_special_characters = "!?@#$%^&*()_+-=[]{}|:;<>,./"
-
-  # Validate presence of a number in password
-  mattr_accessor :password_requires_number
-  @@password_requires_number = false
+  # Password complexity configuration
+  mattr_accessor :password_complexity
+  @@password_complexity = {
+    require_upper: false,    
+    require_lower: false,    
+    require_digit: false,   
+    require_special: false,  
+    special_characters: "!?@#$%^&*()_+-=[]{}|:;<>,./"
+  }
 
   # The time the user will be remembered without asking for credentials again.
   mattr_accessor :remember_for
--- a/lib/devise/models/validatable.rb
+++ b/lib/devise/models/validatable.rb
@@ -13,11 +13,16 @@ module Devise
     #
     #   * +email_regexp+: the regular expression used to validate e-mails;
     #   * +password_length+: a range expressing password length. Defaults to 6..128.
-    #
+    #   * +password_requires_lowercase+: a boolean to require a lower case letter in the password. Defaults to false.
+    #   * +password_requires_uppercase+: a boolean to require an upper case letter in the password. Defaults to false.
+    #   * +password_requires_special_character+: a boolean to require a special character in the password. Defaults to false.
+    #   * +password_requires_number+: a boolean to require a number in the password. Defaults to false.
+    #   * +password_special_characters+: a string with the special characters that are allowed in the password. Defaults to nil.
+    
     module Validatable
       # All validations used by this module.
       VALIDATIONS = [:validates_presence_of, :validates_uniqueness_of, :validates_format_of,
-                     :validates_confirmation_of, :validates_length_of].freeze
+                     :validates_confirmation_of, :validates_length_of, :validate].freeze
 
       def self.required_fields(klass)
         []
@@ -35,6 +40,13 @@ module Devise
           validates_presence_of     :password, if: :password_required?
           validates_confirmation_of :password, if: :password_required?
           validates_length_of       :password, within: password_length, allow_blank: true
+
+          validates_format_of :password, with: /\p{Lower}/, if: -> { password_requires_lowercase }, message: :must_contain_lowercase
+          validates_format_of :password, with: /\p{Upper}/, if: -> { password_requires_uppercase }, message: :must_contain_uppercase
+          validates_format_of :password, with: /\d/, if: -> { password_requires_number }, message: :must_contain_number
+
+          # Run as special character check as a custom validation to ensure password_special_characters is evaluated at runtime
+          validate :password_must_contain_special_character, if: -> { password_requires_special_character }
         end
       end
 
@@ -60,8 +72,47 @@ module Devise
         true
       end
 
+      # Make these instance methods so the default Devise.password_requires_<> 
+      #can be overridden
+      def password_requires_lowercase
+        self.class.password_requires_lowercase
+      end
+
+      def password_requires_uppercase
+        self.class.password_requires_uppercase
+      end
+      
+      def password_requires_number
+        self.class.password_requires_number 
+      end
+      
+      def password_requires_special_character
+        self.class.password_requires_special_character
+      end
+
+      def password_special_characters
+        self.class.password_special_characters
+      end
+
+      def password_must_contain_special_character
+        special_character_regex = /[#{Regexp.escape(password_special_characters)}]/
+      
+        unless password =~ special_character_regex
+          errors.add(:password, :must_contain_special_character)
+        end
+      end
+      
       module ClassMethods
-        Devise::Models.config(self, :email_regexp, :password_length)
+        Devise::Models.config(
+          self,
+          :email_regexp,
+          :password_length,
+          :password_requires_lowercase,
+          :password_requires_uppercase,
+          :password_requires_special_character,
+          :password_requires_number,
+          :password_special_characters
+        )
       end
     end
   end
--- a/lib/devise/models/validatable.rb
+++ b/lib/devise/models/validatable.rb
@@ -16,7 +16,7 @@ module Devise
     #   * +password_requires_lowercase+: a boolean to require a lower case letter in the password. Defaults to false.
     #   * +password_requires_uppercase+: a boolean to require an upper case letter in the password. Defaults to false.
     #   * +password_requires_special_character+: a boolean to require a special character in the password. Defaults to false.
-    #   * +password_requires_number+: a boolean to require a number in the password. Defaults to false.
+    #   * +password_requires_digit+: a boolean to require a digit in the password. Defaults to false.
     #   * +password_special_characters+: a string with the special characters that are allowed in the password. Defaults to nil.
     #
     # Since +password_length+ is applied in a proc within `validates_length_of` it can be overridden
@@ -45,7 +45,7 @@ module Devise
 
           validates_format_of :password, with: /\p{Lower}/, if: -> { password_requires_lowercase }, message: :must_contain_lowercase
           validates_format_of :password, with: /\p{Upper}/, if: -> { password_requires_uppercase }, message: :must_contain_uppercase
-          validates_format_of :password, with: /\d/, if: -> { password_requires_number }, message: :must_contain_number
+          validates_format_of :password, with: /\d/, if: -> { password_requires_digit }, message: :must_contain_digit
 
           # Run as special character check as a custom validation to ensure password_special_characters is evaluated at runtime
           validate :password_must_contain_special_character, if: -> { password_requires_special_character }
@@ -76,24 +76,28 @@ module Devise
 
       # Make these instance methods so the default Devise.password_requires_<> 
       #can be overridden
+      def password_complexity
+        self.class.password_complexity
+      end
+
       def password_requires_lowercase
-        self.class.password_requires_lowercase
+        password_complexity[:require_lower]
       end
 
       def password_requires_uppercase
-        self.class.password_requires_uppercase
+        password_complexity[:require_upper]
       end
       
-      def password_requires_number
-        self.class.password_requires_number 
+      def password_requires_digit
+        password_complexity[:require_digit]
       end
       
       def password_requires_special_character
-        self.class.password_requires_special_character
+        password_complexity[:require_special]
       end
 
       def password_special_characters
-        self.class.password_special_characters
+        password_complexity[:special_characters]
       end
 
       def password_must_contain_special_character
@@ -109,11 +113,7 @@ module Devise
           self,
           :email_regexp,
           :password_length,
-          :password_requires_lowercase,
-          :password_requires_uppercase,
-          :password_requires_special_character,
-          :password_requires_number,
-          :password_special_characters
+          :password_complexity
         )
       end
     end
--- a/lib/devise/models/validatable.rb
+++ b/lib/devise/models/validatable.rb
@@ -13,11 +13,12 @@ module Devise
     #
     #   * +email_regexp+: the regular expression used to validate e-mails;
     #   * +password_length+: a range expressing password length. Defaults to 6..128.
-    #   * +password_requires_lowercase+: a boolean to require a lower case letter in the password. Defaults to false.
-    #   * +password_requires_uppercase+: a boolean to require an upper case letter in the password. Defaults to false.
-    #   * +password_requires_special_character+: a boolean to require a special character in the password. Defaults to false.
-    #   * +password_requires_digit+: a boolean to require a digit in the password. Defaults to false.
-    #   * +password_special_characters+: a string with the special characters that are allowed in the password. Defaults to nil.
+    #   * +password_complexity+: a hash with the password complexity requirements, with the following keys:
+    #       - +require_lower+: a boolean to require a lower case letter in the password. Defaults to false.
+    #       - +require_upper+: a boolean to require an upper case letter in the password. Defaults to false.
+    #       - +require_special+: a boolean to require a special character in the password. Defaults to false.
+    #       - +require_digit+: a boolean to require a digit in the password. Defaults to false.
+    #       - +special_characters+: a string with the special characters that are allowed in the password. Defaults to nil.
     #
     # Since +password_length+ is applied in a proc within `validates_length_of` it can be overridden
     # at runtime.
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -185,20 +185,14 @@ Devise.setup do |config|
   # to give user feedback and not to assert the e-mail validity.
   config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
 
-  # Password requires a lower case letter
-  # config.password_requires_lowercase = false
-
-  # Password requires an upper case letter
-  # config.password_requires_uppercase = false
-
-  # Password requires a number
-  # config.password_requires_number = false
-
-  # Password requires a special character
-  # config.password_requires_special_character = false
-
-  # Special characters that are allowed in the password
-  # config.password_special_characters = "!?@#$%^&*()_+-=[]{}|:;<>,./"
+  # Password complexity configuration
+  # config.password_complexity = {
+  #   require_upper: false,    
+  #   require_lower: false,    
+  #   require_digit: false,   
+  #   require_special: false,  
+  #   special_characters: "!?@#$%^&*()_+-=[]{}|:;<>,./"
+  # }
 
   # ==> Configuration for :timeoutable
   # The time you want to timeout the user session without activity. After this
--- a/lib/generators/templates/devise.rb
+++ b/lib/generators/templates/devise.rb
@@ -185,6 +185,21 @@ Devise.setup do |config|
   # to give user feedback and not to assert the e-mail validity.
   config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
 
+  # Password requires a lower case letter
+  # config.password_requires_lowercase = false
+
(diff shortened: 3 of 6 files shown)


kykyi:
Hey @nashby if I could please request a review 😄

lib/devise.rb
```diff
@@ -120,6 +120,26 @@ module Test
...
+  @@password_requires_special_character = false
+
+  # Special character options
+  mattr_accessor :password_special_characters
+  @@password_special_characters = "!?@#$%^&*()_+-=[]{}|:;<>,./"
+
+  # Validate presence of a number in password
+  mattr_accessor :password_requires_number
+  @@password_requires_number = false
+
```
kykyi, line 142: I wonder if there could be some more general config like:
```ruby
  @@required_complex_password = false
```
Which if `true` would require all of these individual pieces?
So the validations could become:
```ruby
validates_format_of :password, with: /\p{Upper}/, if: -> { password_requires_uppercase || require_complex_password }, message: :must_contain_uppercase

**Comment by kykyi** [comment]
Polite bump @nashby @carlosantoniodasilva 😇

**Comment by datpmt** [comment]
how about modify the following configurations in the initializer file as below?
```ruby
config.password_complexity = {
  upper: 1,    # At least 1 uppercase letter
  lower: 2,    # At least 2 lowercase letters
  digit: 3,    # At least 3 digits
  special: 4,  # At least 4 special characters
}
```

kykyi:
> how about modify the following configurations in the initializer file as below?
> 
> ```ruby
> config.password_complexity = {
>   upper: 1,    # At least 1 uppercase letter
>   lower: 2,    # At least 2 lowercase letters
>   digit: 3,    # At least 3 digits
>   special: 4,  # At least 4 special characters
> }
> ```

Thanks @datpmt seems like an elegant solution ✅ One issue which I could see arise however could be a clash between this and password length minimums? For ex, if you set the above, but stuck with the default 8 character minimum, you couldn't satisfy all the configured preferences. I think something like this could use your nicer syntax but also be more ergonomic with the wider validation system:
```ruby
config.password_complexity = {
  upper: true,    # require upper
  lower: false,    # don't require lower
  digit: true,    # require digit
  special: true,  # require special character
  special_characters: ["!", "?", "@", "\"]
}
```
What do you think?

datpmt:
> stuck with the default 8 character minimum

@kykyi Ah I see. Cool! Let do it! :thumbsup:

kykyi:
@datpmt updated to use your dict style ✅

test/models/validatable_test.rb
```diff
@@ -1,10 +1,9 @@
 # encoding: UTF-8
 # frozen_string_literal: true
-
-require 'test_helper'
+require "test_helper"
```
datpmt, line 3:
```suggestion
require 'test_helper'
```

Prefer single-quoted strings when you don't need string interpolation or special symbols.
References: [rubocop](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/QuotedSymbols)

test/models/validatable_test.rb
```diff
@@ -121,4 +120,85 @@ class ValidatableTest < ActiveSupport::TestCase
...
+        assert user.invalid?  
+        assert_equal 'must include at least one special character', user.errors[:password].join
+      end
+    end
+  end 
+
+  def with_password_requirement(requirement, value)
+    # Change the password requirement and restore it after the block is executed
+    original_password_complexity= User.public_send("password_complexity")
+    original_value = original_password_complexity[requirement]
```
datpmt, line 194: Useless assignment to variable - `original_value`.

test/models/validatable_test.rb
```diff
@@ -1,10 +1,9 @@
 # encoding: UTF-8
 # frozen_string_literal: true
-
-require 'test_helper'
+require "test_helper"
 
 class ValidatableTest < ActiveSupport::TestCase
-  test 'should require email to be set' do
+  test 'should require email to be set' do
```
datpmt, line 6:
```suggestion
  test 'should require email to be set' do
```
remove redundant space

nvasilevski:
Sorry for stirring the pot but I wanted to cross-post an opinion that presence of upper-cased letters, special characters and numbers has very little to do with password strength and what really contributes to the password strength is the length of the password. I'm genuinely worried that enforcing these password requirements from default will only contribute to poor user experience and potentially less secure passwords overall

More context - https://github.com/rails/rails/pull/53984#issuecomment-2551842959

kykyi:
Agreed @nvasilevski I think whilst this change pushes users to increase the entropy of their passwords, setting these and forgetting could lead devs into a false sense of security ➕

timdiggins:
I agree with @nvasilevski - here's a specific argument against complexity requirements from the UK's National Cyber Security Centre: https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#PasswordGuidance:UpdatingYourApproach-Donotusecomplexityrequirements

Recommend closure of the issue for Devise

fthobe:
@nvasilevski @timdiggins Mates, I would like to break some glass here. 

I do not believe that what you write is wrong, the contrary, I believe some very valid points have been made. I am, as probably also you, sure that we are going towards a mixture of MFA and / or passwordless login and that's overall a good thing, but the environments around us are frequently not there yet. 

The problem is the world: 
- password complexity / length policies are frequenty requirements to pass audits;
- password complexity / length should have a reference implementation to avoid errors in implementation.

I think @kykyi made a great PR covering everything from configurability to effective resolution of the issue, maybe you could give it a look with different eyes (fearing weak integrations instead of fearing false sense of security) and we add some lines into documentation to make this a better PR covering also that the entire UTF-8 character set is only as great as the password length. 

I'd really appreciate a feedback from you guys on this.

kykyi:
Thanks for the perspective and for weighing in @fthobe 🙏 . I'll reopen and wait for a maintainer to merge/comment/close just so the issue can be resolved 😄

fthobe:
> Thanks for the perspective and for weighing in @fthobe 🙏 . I'll reopen and wait for a maintainer to merge/comment/close just so the issue can be resolved 😄 

Actually @nvasilevski did not obligate you to close it, sometimes things need some time to get traction and sometimes topics move in waves. 

Be patient, open source is neither fast nor democratic 😅

fthobe:
@timdiggins Hey, I would be super interested in your oppinion on my comment :)

fthobe:
Ping @nvasilevski @datpmt @timdiggins 

Could you retake a look at this PR and the comments made above.

datpmt:
> Ping @nvasilevski @datpmt @timdiggins
> 
> Could you retake a look at this PR and the comments made above.


First of all, I completely agree that the length of a password significantly impacts its strength.

In fact, many websites and company security policies require users to apply complexity rules to their passwords. This means that web applications using `Devise` have a valid reason to implement password complexity as a feature. Whether to enable it or not is up to the owner’s discretion.

I believe `Devise` should offer this option.

fthobe:
> I believe Devise should offer this option.

Is this PR for you acceptable or does it need additional work?

timdiggins:
@fthobe im. It a maintainer/committee here so my 10c isn't worth so much 😀 

A pragmatic response: My sense is that there isn't much maintainer involvement and devise is going into decline - there are CI-fix and bug fix PRs that have had no maintainer involvement so I don't think there's feature development will gain any traction.

fthobe:
@timdiggins a lot of stuff still relies on it so we need to work with what we have. 

I honestly think it's a very mature product. 

Anyhow I'd really appreciate your opinion. 

@carlosantoniodasilva does this PR have a shot to be integrated?

datpmt:
> > I believe Devise should offer this option.
> 
> @datpmt Is this PR for you acceptable or does it need additional work?

I will take the time to run it locally and provide a review as soon as possible.

fthobe:
> roposal is different, it's an option that isn't enabled by default so I have nothing of a substance to say against. I just hope that applications won't reach for this feature without absolute necessity and let me use my long, secure, easy to remember and easy to type passwords.

Man, to be honest, it's a mess. 
[PCI Compliancy ](https://blog.hypr.com/pci-dss-4-password-mfa-requirements) just to name one standard (probably the most used one) is directly asking for complexity + MFA. This practically means by law as all ecommerce companies over a certain volume need to adhere. 
NIST is saying the opposite. 

It should be everybody's choice guided by strong conventions how to do it (which means here) and that's why I am pushing for it.

fthobe:
@kykyi I love your humor, but can we replace 
`config.password_special_characters = ":)"`
with 
`config.password_special_characters = "'-!"#$%&()*,./:;?@[]^_`{|}~+<=>"`
And add a comment that " " (space) is not an accepted character.

fthobe:
@datpmt i have some time on Wednesday to test this with spree and solidus. Is there anything I should be particularly attentive off?

datpmt:
> @datpmt i have some time on Wednesday to test this with spree and solidus. Is there anything I should be particularly attentive off?

@fthobe You need to pay attention to the `require_special` and `special_characters` params. I think require_special is somewhat redundant. If I do it, I will only use `special_characters`. If it's empty, then no special character is required, and vice versa.

fthobe:
@kykyi can you implement a change? @datpmt point of uniting special and special_characters is very valid to keep everything streamlined. I can test it out on Wednesday with some gems that use devise.

fthobe:
@kykyi thank!

@datpmt this should be as you wanted it.

fthobe:
@kykyi also given the small amount of files involved you could consider to squash the commits as "merge..." is not a great commit Name for clarity.

fthobe:
@datpmt worked for me, did you encounter any issues?

fthobe:
@datpmt Baden (@kykyi ) was kind enough to make the changes. Could we carry this over the finish line?

kykyi:
@fthobe for sure, thanks @datpmt, we need a maintainer to approve though?

fthobe:
@kykyi the reality is @datpmt is not much of a password complexity lover, I think his password is test123, I bet he gets hacked frequently :bowtie: .

Jokes apart, @datpmt @kykyi put a lot of effort in this and as outlined above it has its merits. Can we get this over the finish line?

JustShah:
This looks great! 🎉 , it helps to keep things secure while giving developers flexibility. I really support this change.

leesmith:
This is a welcomed feature to Devise! Thank you, @kykyi!

If for no other reason, implement password complexity in your apps so that people might be more likely to use a password manager once they grow tired of trying to remember complex passwords.

datpmt:
@kykyi @fthobe
I am just someone who supports this pull request and has given some reviews from my personal perspective. I am not the person who can merge this pull request. You know, I am still struggling with my busy life, but I am still keeping an eye on this pull request.

fthobe:
@nashby what can be done to bring this over the finish line?

codepilotsf:
Yes please!!!

lib/generators/templates/devise.rb
```diff
@@ -185,6 +185,14 @@
   # to give user feedback and not to assert the e-mail validity.
   config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
 
+  # Password complexity configuration
+  # config.password_complexity = {
+  #   require_upper: false,    
+  #   require_lower: false,    
+  #   require_digit: false,   
+  #   special_characters: "!?@#$%^&*()_+-=[]{}|:;<>,./"
```
salzig, line 193: IMHO: We shouldn't limit the list of allowed special characters, even if this is only a suggestion because it isn't enabled by default. If the user is able remember and write any of this letters `«∑€®†Ω¨⁄øπ•±å‚∂ƒ©ªº∆@œæ‘≤¥≈ç√∫~µ∞…»„‰¸˝ˇÁÛØ∏°ÅÍ™ÏÌÓıˆﬂŒÆ’≥‡ÙÇ◊‹›˘˛÷—` (just examples), we should make sure they can be used with suggested default. 

What's wrong with asking for any non word character? Something like `\W` 

> /\W/ - A non-word character ([^a-zA-Z0-9_]).  – https://ruby-doc.org/core-2.5.8/Regexp.html#class-Regexp-label-Character+Classes

salzig:
I think it's good to question the default of 6 to 128 characters, especially since it is configurable it should maybe suggest a more "demanding" default of at least 12 characters. 

But I don't like this implementation. 

There are multiple ways to write secure passwords, enforcing one way would reduce the number of possible passwords that any user could use. Suggesting a naive at least one character of 4 character-classes isn't considered the single best approach, so we shouldn't use it as suggestion. 

If there is any specific requirement you have to comply with, use a specific password validator. Maybe it would be good to have gems like `pci_dss_password_validator`, `password_validator_bsi`, `password_rules_ncsc`, `nist_password_validator`, … and suggest with a comment that those could be used to comply with local regulations. But choosing a global default (devise is used everywhere) should (IMHO) be considered out of scope. 

To describe it more dystopian: Attackers would be happy to know that with a devise application there is a good chance the password is following a specific pattern. If we want to improve the security of users using applications build with devise, we shouldn't make it easier for attackers. **Which means making it harder for attackers should be our top priority.** 

Want to make it harder for attackers? Rate-Limit the login per user. -> have a look into [Rack::Attack](https://github.com/rack/rack-attack) or similar.

fthobe:
@salzig thank for your joining the conversation. I believe all your points are valid and that this PR does not represent the ultimate solution, but simply throws something most of us do anyway in a standardised procedure. 

Right now most devise installations (especially those using spree or solidus) and plenty of other apps using transaction relevant data or even store cc tokens will have one or another way to raise the complexity scheme of passwords. In my school of thought, creating outcome variability in the way that this done does not ease maintenance or security and having this standardised might not be perfect but a step forward.

lib/generators/templates/devise.rb
kykyi: Thanks for the suggestion, updated the code, looks much cleaner now 🙏

kykyi:
Hey all, I've updated the README and included an updated for `\W` based on your suggestion @salzig 🙏

I agree this PR won't make passwords on Devise impenetrable. But given how Devise is an out-of-the-box, batteries-included solution for auth, I think these changes set a minimum level of password complexity which (along with an increase in min password length) pushes users of Devise applications to use better passwords.

kykyi:
Hey all, I've updated the README and included an updated for \W based on your suggestion @salzig 🙏

I agree this PR won't make passwords on Devise impenetrable. But given how Devise is an out-of-the-box, batteries-included solution for auth, I think these changes set a minimum level of password complexity which (along with an increase in min password length) pushes users of Devise applications to use better passwords.

lib/generators/templates/devise.rb
fthobe: Hey @salzig I really appreciate this, but we do not all have identical systems. Special character support is not universal and some characters simply do not work everywhere.

salzig: Sure, not all work everywhere, but the default should not reduce security in unnecessary manner. Special characters is still configurable.

fthobe:
@nashby and @carlosantoniodasilva 
There has been no issue having had this much traction on devise in years (look at the reactions to the PR). It's a sought after feature, can we get this out of door somehow?

salzig:
If the real intend is to increase the default security without arguing about what way of enforcing patterns would be the best, than you should support https://github.com/heartcombo/devise/pull/5685. That really helps the default a lot. 

And we don't have to stop argue about how we can improve even further, but that one is really the base to having a more modern default.

lib/generators/templates/devise.rb
kykyi: Ok I can revert to allow a configurable list of special chars

fthobe:
> If the real intend is to increase the default security without arguing about what way of enforcing patterns would be the best, than you should support #5685. That really helps the default a lot.

Man, you wouldn't believe how much I agree with you, I know this seems like a crusade. But I feel like I can be for long passwords and at the same time believe that many people due to external requirements need password policies and that we should find a default implementation for this.

lib/devise/models/validatable.rb
```diff
@@ -62,8 +76,53 @@ def email_required?
...
+
+      def password_requires_special_character
+        password_complexity[:require_special_character]
+      end
+
+      def allowed_special_characters
+        password_complexity[:allowed_special_characters]
+      end
+
+      def password_contains_special_character
```
kykyi, line 105: @salzig @fthobe this allows both approaches: `\W` and check based on configured special chars 👍

fthobe (approved):
Looks great to me @kykyi 
But I'd really prefer also @salzig and @gregmolnar to look over this.

fthobe:
hey @kuahyeow , thank you for participating in the discussion. As outlined above >10m Downloads of gems adding this functionality clearly indicate that the need of 30% of Devise installations is exactly this feature. 

I fail to see yet one topic related argument why we can’t reduce implementation variability by adding exactly this.  All I read is arguing how longer passwords are better completely missing the Problem: having Code fragmentation for such a basic Feature in one of the most sensitive parts of the rails ecosystem.

salzig:
> As outlined above >10m Downloads of gems adding this functionality clearly indicate that the need of 30% of Devise installations is exactly this feature.

30% is an argument for a separate Gem, not for inclusion. IMHO: only something used by ~80-90% of the users should be considered for inclusion. It's just more source code the maintainers have to work with, and there is already a lot.

kykyi:
Throwing this into the ring, not by any means peer reviewed or backed by data, but since devise is kind of a roll-your-own Auth0 in some respects, I wanted to share that they define complexity like this:

- **None** (default): at least 1 character of any type.
- **Low**: at least 6 characters.
- **Fair**: at least 8 characters including a lower-case letter, an upper-case letter, and a number.
- **Good**: at least 8 characters including at least 3 of the following 4 types of characters: a lower-case letter, an upper-case letter, a number, a special character (such as !@#$%^&*).
- **Excellent**: at least 10 characters including at least 3 of the following 4 types of characters: a lower-case letter, an upper-case letter, a number, a special character (such as !@#$%^&*). Not more than 2 identical characters in a row (for example, 111 is not allowed).

So these changes would give devise similar configurability which would be Good at minimum by their standard. 

https://auth0.com/docs/authenticate/database-connections/password-strength

crespire:
Putting a +1 to this, as it would have come in handy for our app, I will have to implement by hand for now.

fthobe:
@josevalim @tegon @nashby There's no single response to any PR, is anybody of you still maintaining the project? 

At this point wouldn't it be better to maybe hand over the reigns to more volunteers?

gregmolnar:
I think there is a better way to address this problem. Having these validation options wouldn't necessary yield better passwords, I would rather suggest to use the combination of `zxcvbn` and `pwned`. I wrote a short blogpost about how they can be easily used in a Rails apps:

I think Devise is complicated enough on its own and doesn't need any new other dependencies, so my suggestion is to add a recommendation of those tools to the Readme, so folks can pull in the dependencies should they want.

fthobe:
> It seems a bit non-sensical to have to pull in additional dependencies if I'm using Devise validatable already.

Fully agree.

gregmolnar:
> Just my two cents, but if Devise `validatable` is already doing length validation, it seems to make sense to also allow the option to configure complexity, as it is part of what would determine a valid password in the context of a Devise based authentication system.

Length matters way more than the "complexity" these options would offer.


> It seems a bit non-sensical to have to pull in additional dependencies if I'm using Devise `validatable` already.

Fair point, but I still wouldn't recommend using these validators, because they give you a false sense of security.

> To me, the logical outcome from your stance is to actually fully drop `validatable` all together, and if people want to validate passwords, they should roll it by hand or pull in another gem to fully take that responsibility. Why do half measures either way?

I think then it would be better to accept a new dependency(two actually) and add a new validator that uses `zxcvbn` and another one that uses the `pwnd` gem so folks could choose the ones they want to use.

kykyi:
It feels completely within the remit of a Devise - an authentication solution - to provide password validations, wether through the means proposed in this PR or through some other means.

gregmolnar:
> It feels completely within the remit of a Devise - an authentication solution - to provide password validations, wether through the means proposed in this PR or through some other means.

It already does that, although it could do even more. But, (don't take it wrong) your solution gives a false sense of security. 
Let's give the maintainers some time to sort out maintenance of the gem and then introduce 2 new validators. One based on `zxcvbn` and one based on `pwnd`. Ideally both would be used in tandem, but even if one of them used besides the current length validator, that would enforce pretty good passwords.

fthobe:
@gregmolnar 
> It already does that, although it could do even more. But, (don't take it wrong) your solution gives a false sense of security.

How often do you want to repeat a minority oppinion?

kuahyeow:
@fthobe What are you really after ? If it's [PCI-DSS 4.0](https://blog.hypr.com/pci-dss-4-password-mfa-requirements), then this is only part of the picture, as it requires:

- Minimum 12 characters, and use special characters, uppercase, and lowercase letters.

Otherwise, implementing half of this (adding password complexity without fixing the `6` default) does not really help. 

Providing a configurable thing which allows for developer to potentially select `require_upper: true, require_lower: false, require_special_character: true, require_digit: false, allowed_special_characters: '!@#$%'` is also just implementing half of PCI-DSS. 

Side-note: Interesting https://docs-prv.pcisecuritystandards.org/PCI%20DSS/Standard/PCI-DSS-v4_0_1.pdf says something that is more relaxed:

> 
> 8.3.6 If passwords/passphrases are used as
> authentication factors to meet Requirement 8.3.1,
> they meet the following minimum level of
> complexity:
> • A minimum length of 12 characters (or IF the
> system does not support 12 characters, a
> minimum length of eight characters).
> • Contain both numeric and alphabetic characters.

----

> It seems a bit non-sensical to have to pull in additional dependencies if I'm using Devise validatable already.

This is a bit like saying Rails provides validations so therefore the developer does not have to write their own validations anymore.

gregmolnar:
> @gregmolnar
> 
> > It already does that, although it could do even more. But, (don't take it wrong) your solution gives a false sense of security.
> 
> How often do you want to repeat a minority oppinion?

As often as it needs for people to understand that "Admin123!" is a weak password so any validation rule that permits that is bad.
As I suggested above, once the maintenance is sorted, it would be better to add validation helpers offers better results.

fthobe:
1) Convention over configuration: this is needed by such a vast amount of people that keeping it separate doesn't make sense. 

2) it's not about PCI DSS per se, there are tons of standards that require password complexity. 

Till now i haven't seen a single convincing argument against it. The only compelling argument was "false sense of security", but if people still have to deal with Standards i prefer a sanitized comunity approved standard environment instead of everybody coming their own broth. 

Please advise: what sense does it have not to do it?

Do we want to force custom development enabling all the risks coming with it?
Do We want to Force people into extensions with all the risks coming with it?

Let's slap a big disclaimer on it and get this done, it might not make sense for everybody, but it's better than the alternatives.oä

kuahyeow:
> Convention over configuration: this is needed by such a vast amount of people that keeping it separate doesn't make sense.

Where is the convention ? I see 5 separate configs instead. Convention would be more something like:

```
config.password_complexity_requirements = :pci_dss
```

> it's not about PCI DSS per se, there are tons of standards that require password complexity.

I am curious, what other standards can we form another convention on ?

> The only compelling argument was "false sense of security", but if people still have to deal with Standards i prefer a sanitized comunity approved standard environment instead of everybody coming their own broth.

Feel free to use https://github.com/devise-security/devise-security/blob/main/lib/devise-security/validators/password_complexity_validator.rb. It seems to have its own community around it.

PR #5729, opened 2024-11-24
Fix and cleanup CI

jeromedalbert:
## Problem

- CI is currently failing for `Gemfile-rails-main` because Rails 8+ requires Ruby 3.2+
- Ruby 2.7 and Ruby 3.0 are EOL. Ref: https://endoflife.date/ruby
- Rails 6.x is EOL. Ref: https://endoflife.date/rails

## Solution

- Tweak CI config to ignore Rails 3.1 for Rails main
- Add a `Gemfile-rails-8-0` while I was at it
- Remove EOL Ruby and Rails versions from CI config

After this, CI should be green!


nashby:
@jeromedalbert hey, thanks for PR! Could we please start with just fixing CI without removing Support for older rubies and Rails 6.1?

PR #5730, opened 2024-11-25
Fix Rails main CI

jeromedalbert:
CI is currently failing for Gemfile-rails-main because Rails 8+ requires Ruby 3.2+.

This PR tweaks the CI config to ignore Rails 3.1 for Rails main.


nashby:
thanks!

PR #5731, opened 2024-11-25
Fix Zeitwerk autoloading when ActionMailer is not present.

### LINKED ISSUE #5140: Rails6 without ActionMailer won't boot with zeitwerk eager-loading

nashby:
When `ActionMailer` is not defined we have empty `app/mailers/devise/mailer.rb` file and Zeitwerk doesn't like that and errors with

`expected file app/mailers/devise/mailer.rb to define constant Devise::Mailer`

The fix is to tell Zeitwerk to ignore that file if `ActionMailer` constant if not defined. 

I tried to write a spec for it but since specs are run in the same process it's hard to have two Rails applications where one of them has `ActionMailer` define and the seconds one doesn't.

closes #5140, #5485

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,5 +19,8 @@
   * Add Rails 8 support.
     - Routes are lazy-loaded by default in test and development environments now so Devise loads them before `Devise.mappings` call.
 
+* bug fixes
+  * Make `Devise` work without `ActionMailer` when `Zeitwerk` autoloader is used.
+
 Please check [4-stable](https://github.com/heartcombo/devise/blob/4-stable/CHANGELOG.md)
 for previous changes.
--- a/lib/devise/rails.rb
+++ b/lib/devise/rails.rb
@@ -47,5 +47,11 @@ module Devise
           )
         end
     end
+
+    initializer "devise.configure_zeitwerk" do
+      if Rails.autoloaders.zeitwerk_enabled? && !defined?(ActionMailer)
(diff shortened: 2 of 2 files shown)


Faq:
@nashby with what Rails version this was tested?
with: 
```
gem 'devise', git: 'https://github.com/heartcombo/devise.git', ref: '7eccc91f615fd9bbb8a347ed05a9fb34b1664bf3'
```

in devise initializer:
[4] pry(main)> config.mailer
```
NameError: uninitialized constant Devise::Mailer
from /Users/myuser/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/activesupport-7.0.4.3/lib/active_support/inflector/methods.rb:280:in `const_get'
```

Issue #2469, opened 2013-06-14
Default registerable not working in 3.0.0.rc with Rails 3.2.13 and strong_parameters 0.2.1


filipegiusti:
These 3 doesn't seems to be working fine together.

`Devise::RegistrationsController#create` isn't working for the default case and I get in the logs `Unpermitted parameters: email, password, password_confirmation`.

I didn't have time to pull devise and run the tests, but what I could find was that

```
$ bundle exec rails c
Loading development environment (Rails 3.2.13)
irb(main):001:0> class User < ActiveRecord::Base; devise :database_authenticatable; end
=> {}
irb(main):002:0> params = ActionController::Parameters.new(user: { "email" => "jose", "password" => "passpass",  "role" => "invalid" })
=> {"user"=>{"email"=>"jose", "password"=>"passpass", "role"=>"invalid"}}
irb(main):003:0> sanitizer = Devise::ParameterSanitizer.new(User, :user, params)
=> #<Devise::ParameterSanitizer:0x007f5487b3f730 @resource_class=User(Table doesn't exist), @resource_name=:user, @params={"user"=>{"email"=>"jose", "password"=>"passpass", "role"=>"invalid"}}, @blocks={}>
irb(main):004:0> sanitizer.for(:sign_up)
=> {}
```

This is almost the same code run in a test case in devise's test suite, but with a different end  result.

This may be the same problem of https://github.com/plataformatec/devise/pull/2452.

I found that #permit was receiving an Array instead of multiple parameters here https://github.com/plataformatec/devise/blob/master/lib/devise/parameter_sanitizer.rb#L48 and this may be the root cause. See the implementation of permit here https://github.com/rails/strong_parameters/blob/master/lib/action_controller/parameters.rb#L64

Anyway, this is an incomplete report, I'll try to look into it and improve this report in the following days.

PR #5745, opened 2024-12-17
Lock Nokogiri version for Webrat monkey-patch

c960657:
The monkey-patch ~#2469~ in [test/support/webrat/matchers.rb](https://github.com/heartcombo/devise/blob/main/test/support/webrat/matchers.rb) is not compatible with Nokogiri ≥ 1.17, so many tests are currently failing.

Nokogiri [dropped support for Ruby 2.7 in 1.16](https://github.com/sparklemotion/nokogiri/pull/3040) which is still supported by Devise, so locking Nokogiri to < 1.17 seems like the easiest fix.

### DIFF
--- a/gemfiles/Gemfile-rails-6-0
+++ b/gemfiles/Gemfile-rails-6-0
@@ -16,6 +16,7 @@ group :test do
   gem "omniauth-openid"
   gem "rexml"
   gem "timecop"
+  gem "nokogiri", "< 1.17.0", require: false
   gem "webrat", "0.7.3", require: false
   gem "mocha", "~> 2.1", require: false
 end
--- a/gemfiles/Gemfile-rails-6-1
+++ b/gemfiles/Gemfile-rails-6-1
@@ -22,6 +22,7 @@ group :test do
   gem "omniauth-openid"
   gem "rexml"
   gem "timecop"
+  gem "nokogiri", "< 1.17.0", require: false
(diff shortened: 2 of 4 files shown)


c960657:
@jrichy1 I don't understand what you mean. Where exactly are you tagged?

timdiggins:
I'm not a committer, but I don't think this PR is mergeable. 

Locking devise to an older version of Nokogiri, a gem that gets a lot of security updates will lock out a lot of users for long ago compatibility support. 

Better to drop ruby 2.7 (which is not compatible with latest Rails versions and went EOL in 2023: https://endoflife.date/ruby)

Also FYI #2469 is an issue not a PR, and has no attached code. 

And as an aside, webrat (last update 10 years ago) needs to be dropped - not  most likely with capybara - which is the logical upgrade path. But we'd need to see a lot of commiter enthusiasm before anyone would do a PR on that.

c960657:
Sorry, the link to #2469 was wrong. Should have been https://github.com/sparklemotion/nokogiri/issues/2469.

Nokogiri is only used in tests, so there is no real security issue. But I agree that it is time to drop Ruby 2.7 support.

timdiggins:
> Nokogiri is only used in tests, so there is no real security issue. 

Oops - yes missed that it was in Gemfile not gemspec... Sorry

> But I agree that it is time to drop Ruby 2.7 support

👍

PR #5746, opened 2024-12-17
Update argument name for config.warden

c960657:
The argument for the block passed to `config.warden` is no a `Warden::Manager` instance but a `Warden::Config` instance, but it is confusingly named `manager` in the generated file.

I suggest renaming this to `warden_config` for clarity.


carlosantoniodasilva:
Thanks!

PR #5769, opened 2025-03-03
Reconfigure warden when the middleware stack is rebuilt:

Edouard-chin:
Reconfigure warden when the middleware stack is rebuilt:

- ### Problem

  Starting from Rails 7.2, in tests environment, it is possible that Rails will rebuild the middleware stack, which for consequence will create a new Warden middleware that will not be configured by Devise.

  ### Context

  Since this commit https://github.com/rails/rails/commit/79fd5b45a27 upstream, when an application uses the `with_routing` helper to create a temporary route set and defines a route that maps to a controller requiring authentication, an error will be raised on request: "RuntimeError: No Failure App provided"

  The issue is due to a memoization on devise.

  ### Solution

  When the `Devise.warden_config` object is modified, reset the memoization.

### DIFF
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -293,9 +293,13 @@ module Devise
   @@helpers << Devise::Controllers::Helpers
 
   # Private methods to interface with Warden.
-  mattr_accessor :warden_config
+  mattr_reader :warden_config
   @@warden_config = nil
   @@warden_config_blocks = []
+  def self.warden_config=(config)
+    @@warden_configured = nil
+    @@warden_config = config
+  end
 
   # When true, enter in paranoid mode to avoid user enumeration.
   mattr_accessor :paranoid
--- a/test/integration/authenticatable_test.rb
+++ b/test/integration/authenticatable_test.rb
@@ -457,6 +457,23 @@ class AuthenticationOthersTest < Devise::IntegrationTest
     end
   end
 
+  test 'reconfigure warden when the middleware stack is rebuilt' do
+    get "/unauthenticated"
+    assert_equal 401, response.status
+
+    with_routing do |set|
+      set.draw do
+        get "/unauthenticated", to: "streaming#index"
+      end
+
+      refute_nil(Devise.warden_config.failure_app)
+
+      assert_nothing_raised do
+        get "/unauthenticated"
+      end
+    end
+  end
+
   test 'does not intercept Rails 401 responses' do
     get '/unauthenticated'
     assert_equal 401, response.status


quixoten:
@Edouard-chin, I tried your branch on my rails project, but I was still getting the `No Failure App provided` error. I was able to get the failing test to pass by adding `Devise.configure_warden!`:

```ruby
  test "should not allow portal users to access pipelines" do
    Devise.configure_warden! # << added line
    sign_in portal_users(:one)
    get pipelines_url
    assert_response :redirect
    assert_redirected_to new_customer_session_path
  end
```

I also tried removing the [`@devise_finalized ||= begin`](https://github.com/Edouard-chin/devise/blob/ec-reconfigure-warden/lib/devise/rails/routes.rb#L10) conditional so that `Devise.configure_warden!` would be called every time `finalize!` was called, but I still got the error. So at some point a new `Devise::RouteSet` is getting created but `finalize!` is not called on it before the next test runs.

I also tried replacing `Devise.configure_warden!` in the test with a some other potential solutions, but none worked:

- `app.routes.finalize!` (NOTE: `app.routes` is a `Rails::Engine::LazyRouteSet`)
- `app.reload_routes!`

---

**UPDATE**: adding `app.routes.finalize!` to the test (instead of `Devise.configure_warden!`) _does_ work if I also remove the `@devise_finalized ||= begin` conditional from the `Devise::RouteSet#finalize!` method. In this case, the `finalize!` method gets called twice. The first call is happening somewhere outside of the test, and the second call is from adding `app.routes.finalize!` directly to the test. I added the following debug statements to the `Devise::RouteSet#finalize!` method:

```ruby
      pp @devise_finalized
      pp Devise.class_variable_get(:@@warden_configured)
      pp Devise.warden_config
```

On the _first_ call (outside of the test), I see:

```ruby
nil
true
{:default_scope=>:user,
 :scope_defaults=>{},
 :default_strategies=>{:customer=>[:magic_link_authenticatable], :employee=>[], :portal_user=>[:database_authenticatable, :two_factor_authenticatable]},
 :intercept_401=>false,
 :failure_app=>#<Devise::Delegator:0x0000755912314da0>}
```

and on the second call (inside the test), I see:

```ruby
true
nil
{:default_scope=>:default, :scope_defaults=>{}, :default_strategies=>{}, :intercept_401=>true}
```

---

I also tried removing both the `@@warden_configured ||= begin` conditional and the `@devise_finalized ||= begin` conditional, but the test still failed without the explicit `app.routes.finalize!` call

quixoten:
Here's a combination of changes in rails and devise that got the test working without needing to change the test itself.

```diff
diff --git a/actionpack/lib/action_dispatch/testing/assertions/routing.rb b/actionpack/lib/action_dispatch/testing/assertions/routing.rb
index 7ed155ab1a..5cec99dae3 100644
--- a/actionpack/lib/action_dispatch/testing/assertions/routing.rb
+++ b/actionpack/lib/action_dispatch/testing/assertions/routing.rb
@@ -69,7 +69,7 @@ def reset_routes(old_routes, old_integration_session)
             app.instance_variable_set(:@routes, old_routes)
             app.instance_variable_set(:@app, old_rack_app)
             @integration_session = old_integration_session
-            @routes = old_routes
+            @routes = old_routes.tap(&:finalize!)
           end
       end
```

and the second change is in devise:

```diff
diff --git a/lib/devise.rb b/lib/devise.rb
index 86d5c395..f0ecc6e4 100644
--- a/lib/devise.rb
+++ b/lib/devise.rb
@@ -294,6 +294,7 @@ module Devise
 
   # Private methods to interface with Warden.
   mattr_reader :warden_config
+  mattr_reader :warden_configured
   @@warden_config = nil
   @@warden_config_blocks = []
   def self.warden_config=(config)
diff --git a/lib/devise/rails/routes.rb b/lib/devise/rails/routes.rb
index f43e62fe..637d3774 100644
--- a/lib/devise/rails/routes.rb
+++ b/lib/devise/rails/routes.rb
@@ -7,8 +7,8 @@ module Devise
   module RouteSet
     def finalize!
       result = super
-      @devise_finalized ||= begin
-        if Devise.router_name.nil? && defined?(@devise_finalized) && self != Rails.application.try(:routes)
+      Devise.warden_configured || begin
+        if Devise.router_name.nil? && Devise.warden_configured && self != Rails.application.try(:routes)
           warn "[DEVISE] We have detected that you are using devise_for inside engine routes. " \
             "In this case, you probably want to set Devise.router_name = MOUNT_POINT, where "   \
             "MOUNT_POINT is a symbol representing where this engine will be mounted at. For "   \
@@ -18,7 +18,6 @@ module Devise
 
         Devise.configure_warden!
         Devise.regenerate_helpers!
-        true
       end
       result
     end
```

Edouard-chin:
Thanks, I didn't see that Rails' `reset_routes` also rebuild the middleware stack with the previous route set that is already finalized (and therefore won't trigger devise's monkeypatch). 

I'll try to figure out a solution.

Edouard-chin:
I opened a fix in Rails https://github.com/rails/rails/pull/54705, as I believe the `with_routing` implementation has some caveats that needs to be fixed. If it gets accepted, this PR is not necessary. I'll close for now

PR #5785, opened 2025-07-08
Tell the recipient software when a password link expires.

arnt:
https://datatracker.ietf.org/doc/draft-ietf-mailmaint-expires/

That draft documents the (rare) existing usage of the Expires field, specifies a tighter syntax, and clarifies that "expires" doesn't mean having to delete anything. "Automatically moving the message from inbox to archive" is sensible.

The [IETF mailmaint WG](https://datatracker.ietf.org/wg/mailmaint/about/) requires two implementations before issuing an RFC. I did this one today, to see how hard it would be.

(Or maybe the true reason is: I searched my mail today for a particular domain and and saw several old password reset messages among the top search results. That's junk, those links are expired, let's tell the mailbox indexer when those messages become poor search results.)

It seems reasonable to extend it to cover confirmation messages too.

### DIFF
--- a/lib/devise/mailers/helpers.rb
+++ b/lib/devise/mailers/helpers.rb
@@ -41,6 +41,15 @@ module Devise
         headers.delete(:from) if default_params[:from]
         headers.delete(:reply_to) if default_params[:reply_to]
 
+        # If this message expires, then specify the time Devise thinks it
+        # was sent and also when it expires. This may specify a second or
+        # two too late, not a problem.
+        expiry = expires_for(action)
+        if expiry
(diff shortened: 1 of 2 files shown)


arnt:
Huh, I didn't see any checklist when I originally made the PR? But it ① follows the code style (and what a nice style it is, too), ② adds tests and ③ contains no breaking changes.

arnt:
The internet-draft has passed review and is now in the RFC editor's queue waiting to be published. This means that the the substance of the document is now fixed but there still isn't a number.

I extended this a little and rebased it.

PR #5793, opened 2025-08-16
feat: allow computed method for To header in mailers

boypie510:
## Description

This PR adds support for computed `To` header in Devise mailers, allowing users to customize the recipient field using `default :to` declarations, similar to how `from` and `reply_to` already work.

## Problem

Currently, Devise hardcodes the `to` field to `resource.email` and doesn't respect custom `default :to` values set in mailer classes. This prevents users from:

- Formatting recipient as "Name <email>" for better display
- Using dynamic recipient logic based on user attributes
- Customizing the `to` field like other mailer headers

## Solution

Add a single line to check `default_params[:to]` in the `headers_for` method:

```ruby
headers.delete(:to) if default_params[:to]
```

This follows the same pattern already used for `from` and `reply_to` fields.

## Changes Made

- **Modified**: `lib/devise/mailers/helpers.rb` - Added support for `default :to`
- **Added**: Test coverage in `test/mailers/mailer_test.rb` for computed `to` field
- **Updated**: `.ruby-version` to use Ruby 3.2.2 for compatibility

## Testing

- ✅ Added comprehensive test for `default :to` functionality
- ✅ Test verifies that custom `to` defaults are respected
- ✅ All existing tests continue to pass
- ✅ Follows existing testing patterns for `from` and `reply_to`

## Example Usage

```ruby
class ApplicationDeviseMailer < Devise::Mailer
  default to: -> { email_address_with_name(resource.email, resource.name) }
end
```

## Breaking Changes

None. This is a backward-compatible enhancement.

## Related Issues

Fixes the issue described in: [Allow providing computed method for To header](link-to-issue)

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (if applicable)
- [x] Code follows existing style guidelines
- [x] No breaking changes introduced

### DIFF
--- a/test/mailers/mailer_test.rb
+++ b/test/mailers/mailer_test.rb
@@ -22,6 +22,7 @@ class MailerTest < ActionMailer::TestCase
     class TestMailerWithDefault < Devise::Mailer
       default from: -> { computed_from }
       default reply_to: ->(_) { computed_reply_to }
+      default to: -> { computed_to }
 
       def confirmation_instructions(record, token, opts = {})
         @token = token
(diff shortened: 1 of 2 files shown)


test/mailers/mailer_test.rb
```diff
@@ -37,10 +38,17 @@ def computed_from
...
       end
+
+      def computed_to
+        "to@example.com"
+      end
     end
 
     mail = TestMailerWithDefault.confirmation_instructions(create_user, "confirmation-token")
-    assert mail.from, "from@example.com"
-    assert mail.reply_to, "reply_to@example.com"
```
boypie510, line 44: https://docs.ruby-lang.org/en/2.1.0/Test/Unit/Assertions.html

These assertions will ALWAYS pass

assert mail.from - checks if mail.from is truthy (which it always is)
"from@example.com" - this second argument is completely ignored

PR #5802, opened 2025-10-30
Fixed inconsistency between description and example in 'Running tests' section

Fuyutsubaki:
I corrected the “Running tests” section because there was an inconsistency between the description and the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ bin/test test/models/trackable_test.rb
 * Running a specific test given a line number or a regex:
 ```bash
 bin/test test/models/trackable_test.rb:16
-bin/test test/models/trackable_test.rb -n /update.*record/ 
+bin/test test/models/trackable_test.rb -n '/update.*record/'
 ```
 
 ## Starting with Rails?


README.md
```diff
@@ -163,9 +163,10 @@ bin/test
 bin/test test/models/trackable_test.rb
 ```
 
-* Running a specific test given a regex:
+* Running a specific test given a line number or a regex:
 ```bash
 bin/test test/models/trackable_test.rb:16
+bin/test test/models/trackable_test.rb -n /update.*record/
```
carlosantoniodasilva: I had to quote this to run in zsh, not sure it's necessary on bash though:

```
% bin/test test/models/trackable_test.rb -n /update.*record/
zsh: no matches found: /update.*record/
```

```
% bin/test test/models/trackable_test.rb -n '/update.*record/'

Run options: -n /update.*record/ --seed 49067

# Running:

..

Finished in 0.103160s, 19.3874 runs/s, 164.7926 assertions/s.
2 runs, 17 assertions, 0 failures, 0 errors, 0 skips
```

are you using bash?

Fuyutsubaki: I’m using bash.
You’re right 
 it’s better to enclose this in quotes. I’ll fix it.

PR #5807, opened 2025-11-05
sha256 password before passing to bcrypt to avoid issues with 72 bytes truncation for passwords

le0pard:
More info: https://github.com/bcrypt-ruby/bcrypt-ruby/issues/283

Reproduction:

```ruby
BCrypt::Password.new(BCrypt::Password.create('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1')) == 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa2'
BCrypt::Password.new(BCrypt::Password.create('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1')) == 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa222333'
BCrypt::Password.new(BCrypt::Password.create('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1')) == 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa222333234234324'
```

All return `true`, so

Password 1: `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1`
Password 2: `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa2`

These two users can login to each other's accounts because brcypt caps hashing to the first 72 bytes. 

```ruby
> hash = Devise::Encryptor.digest(Devise, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1')
=> "$2a$13$XxwZStO7/NHTDjJsnGQhIOSb8ZO12PTL1/.Lze6OIT.qOAfBrqBHS"
> Devise::Encryptor.compare(Devise, hash, 'password')
=> false
> Devise::Encryptor.compare(Devise, hash, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa1')
=> true
> Devise::Encryptor.compare(Devise, hash, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa2')
=> true
```

As solution - hash password to sha256, so it always will be smaller, than 72 bytes. Added fallback for old passwords.

In this case we can reject https://github.com/heartcombo/devise/pull/5806

### DIFF
--- a/test/encryptor_test.rb
+++ b/test/encryptor_test.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'bcrypt'
+
+class EncryptorTest < ActiveSupport::TestCase
+  test 'digest/compare passwords' do
+    hashed_password = Devise::Encryptor.digest(Devise, 'example')
+    assert Devise::Encryptor.compare(Devise, hashed_password, 'example')
+    assert_not Devise::Encryptor.compare(Devise, hashed_password, 'example1')
+  end
+
+  test 'false for incorrect bcrypt string' do
+    assert_not Devise::Encryptor.compare(Devise, 'incorrect_bcrypt_string', 'example')
+  end
+
+  test 'digest/compare support passwords longer 72 bytes' do
+    hashed_password = Devise::Encryptor.digest(Devise, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa123')
+    assert Devise::Encryptor.compare(Devise, hashed_password, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa123')
+    assert_not Devise::Encryptor.compare(Devise, hashed_password, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa125')
+  end
+
+  test 'digest/compare support old bcrypt only passwords' do
+    password = 'example'
+    password_with_pepper = "#{password}#{Devise.pepper}"
+    old_hashed_password = ::BCrypt::Password.create(password_with_pepper, cost: Devise.stretches)
+
+    assert Devise::Encryptor.compare(Devise, old_hashed_password, password)
+    assert_not Devise::Encryptor.compare(Devise, old_hashed_password, 'examplo')
+  end
+end
--- a/lib/devise/encryptor.rb
+++ b/lib/devise/encryptor.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'bcrypt'
+require 'digest'
 
 module Devise
   module Encryptor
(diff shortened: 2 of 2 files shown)


test/encryptor_test.rb
```diff
@@ -0,0 +1,31 @@
...
+  test 'digest/compare support passwords longer 72 bytes' do
+    hashed_password = Devise::Encryptor.digest(Devise, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa123')
+    assert Devise::Encryptor.compare(Devise, hashed_password, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa123')
+    assert_not Devise::Encryptor.compare(Devise, hashed_password, 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa125')
+  end
+
+  test 'digest/compare support old bcrypt only passwords' do
+    password = 'example'
+    password_with_pepper = "#{password}#{Devise.pepper}"
+    old_hashed_password =::BCrypt::Password.create(password_with_pepper, cost: Devise.stretches)
```
kossy0701: Excuse me for interrupting, I thought a half space was needed.

### suggest

```
old_hashed_password = ::BCrypt::Password.create(password_with_pepper, cost: Devise.stretches)
```

le0pard:
You can read about this here - https://security.stackexchange.com/questions/92175/what-are-the-pros-and-cons-of-using-sha256-to-hash-a-password-before-passing-it

TLDR: SHA-256 allows for avoiding length constraints where entropy would otherwise be lost

Passlib in python do the same - https://passlib.readthedocs.io/en/stable/lib/passlib.hash.bcrypt.html#bcrypt-password-truncation
Django framework have such implementation (BCryptSHA256PasswordHasher)

Possible alternative, move to Argon2, which not have such limitations, but this will require to add another dependency and leave bcrypt for backward compatibility

le0pard:
ok, so we can have something like this validation:

https://github.com/rails/rails/blob/ffcbf6f205363f8c2fb3e9834bc86690dd59f1cb/activemodel/lib/active_model/secure_password.rb#L138-L140

But again, not sure it will be not broken changes (maybe not critical, but still)

PR #5816, opened 2026-01-05
Drop obsolete Devise::Models::Authenticatable#inspect method

### LINKED ISSUE #5274: Models with Devise::Models::Authenticatable have degraded pretty-printing
## Environment

- Ruby **2.5.8**
- Rails **6.0.3.1**
- Devise **4.7.1**

## Current behavior

```ruby
class AuthenticatableUser < ActiveRecord::Base
  devise :database_authenticatable
end

pp AuthenticatableUser.first
=begin
#<AuthenticatableUser id: 1, name: "Greg Grimaldus", email: "greg@example.com", nested_field: 
=end
```

Additionally, the output is all one color in a REPL like Pry.

## Expected behavior
```ruby
class NonAuthenticatableUser < ActiveRecord::Base
end

pp NonAuthenticatableUser.first
=begin
#<NonAuthenticatableUser
 id: 1,
 name: "Greg Grimaldus",
 email: "greg@example.com",
 nested_field:
  #<NestedField:0x0ab88383223
   @nested_field_attr=1>,
 ...MANY_MORE_FIELDS...>
=end 
```

Additionally, in a REPL like Pry, each attribute has color highlighting based on the value type.

## Cause

The reason this occurs is because `Devise::Models::Authenticatable` overrides `inspect` and if the method is overriden, then `ActiveRecord::Core#pretty_print` defers to `super`. This is probably `PP::ObjectMixin#pretty_print`, which doesn't have the niceties described in the "Expected behavior" section.

OutlawAndy:
Rails has handled [parameter filtering](https://api.rubyonrails.org/classes/ActiveSupport/ParameterFilter.html) for a number of years now.  So this PR just follows through on the suggestion made in [this comment](https://github.com/heartcombo/devise/pull/4957#issuecomment-450851530)

Fix #5274

Thank you!

### DIFF
--- a/lib/devise/models/authenticatable.rb
+++ b/lib/devise/models/authenticatable.rb
@@ -115,15 +115,6 @@ module Devise
         super(options)
       end
 
-      # Redefine inspect using serializable_hash, to ensure we don't accidentally
-      # leak passwords into exceptions.
-      def inspect
-        inspection = serializable_hash.collect do |k,v|
-          "#{k}: #{respond_to?(:attribute_for_inspect) ? attribute_for_inspect(k) : v.inspect}"
-        end
-        "#<#{self.class} #{inspection.join(", ")}>"
-      end
-
       protected
 
       def devise_mailer


carlosantoniodasilva (commented):
I think this makes sense, but I want to mention that there'll be a slight change in the behavior of `#inspect` because it relies on `serializable_hash` which uses `UNSAFE_ATTRIBUTES_FOR_SERIALIZATION`, and that means any of those attributes that don't match `filter_parameters` values will now show up in `#inspect`.

Either way, I agree it's something worth delegating back to Rails, mentioning it in the changelog.

Would it make sense to ensure Devise includes things like `password` and `token` in the Rails params filter automatically? (even though it's normally part of the `rails new` setup.)

PR #5849, opened 2026-06-10
Update commented email_regexp example to match the current default

### LINKED ISSUE #5721: Potential Update for Rails7 with Ruby3.1
https://github.com/heartcombo/devise/blob/12c796e4994c9fbdc12eb0d6b70450cf82fcec62/test/rails_app/config/initializers/devise.rb#L102

I ran into a failing test with Rspec on the valid email check

The original regex here;
```ruby
config.email_regexp =/^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})$/i
```ruby
was failing on a new line error

I corrected to
```ruby
/\A([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})\z/i
```ruby

Basically the same but changing the start and end removed the multiline anchors error for me.
OLD uses ^ and $: Matches start/end of any line
NEW uses \A and \z: Matches start/end of entire string

Not sure if this is correct as this is my first contribution but I came across and thought I would drop it in.

Cheers

augustocbx:
The commented `email_regexp` example in the test app initializer still showed the old `/^...$/i` regex with multiline anchors, which no longer matches the actual default in `lib/devise.rb` or the generator template. This updates the example to the current default `/\A[^@\s]+@[^@\s]+\z/`. Unlike #5723, it syncs the example with the real default rather than patching the old pattern.

Fixes #5721

### DIFF
--- a/test/rails_app/config/initializers/devise.rb
+++ b/test/rails_app/config/initializers/devise.rb
@@ -99,7 +99,7 @@ Devise.setup do |config|
   # config.password_length = 8..72
 
   # Regex to use to validate the email address
-  # config.email_regexp = /^([\w\.%\+\-]+)@([\w\-]+\.)+([\w]{2,})$/i
+  # config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
 
   # ==> Configuration for :timeoutable
   # The time you want to timeout the user session without activity. After this


carlosantoniodasilva:
Thanks!

Issue #5752, opened 2025-01-07
Devise in test environment: mapping.to.serialize_from_session gives error "wrong number of arguments (given 10, expected 2)"


dianedouglas-thrive:
## Environment

- Ruby **[3.3.6]**
- Rails **[8.0.0]**
- Devise **[4.9.4]**

## Current behavior

I just updated my app to rails 8 with the newest version of Devise and hotwire turbo. When running tests with Capybara/Rspec I'm seeing certain form submissions fail with a 500 error from inside of Devise. This behavior seems to be only in the test environment.

lib/devise.rb:496

```
  # A method used internally to complete the setup of warden manager after routes are loaded.
  # See lib/devise/rails/routes.rb - ActionDispatch::Routing::RouteSet#finalize_with_devise!
  def self.configure_warden! #:nodoc:
    @@warden_configured ||= begin
      warden_config.failure_app   = Devise::Delegator.new
      warden_config.default_scope = Devise.default_scope
      warden_config.intercept_401 = false

      Devise.mappings.each_value do |mapping|
        warden_config.scope_defaults mapping.name, strategies: mapping.strategies

        warden_config.serialize_into_session(mapping.name) do |record|
          mapping.to.serialize_into_session(record)
        end

        warden_config.serialize_from_session(mapping.name) do |args|
          mapping.to.serialize_from_session(*args)
        end
      end

      @@warden_config_blocks.map { |block| block.call Devise.warden_config }
      true
    end
  end
```

When `mapping.to.serialize_from_session(*args)` is called I get the error "wrong number of arguments (given 10, expected 2)". 

I have not been able to find a pattern between tests that pass and tests that cause this behavior except that I can see the data is different. 

On a passing test args looks like this:

```
[1] pry(#<Warden::SessionSerializer>)>  args
=> [[1], "$2a$04$yPg9CBmwVeIZc3VYqdmL3O"]
```

But on a failing test args looks like this:

```
[1] pry(#<Warden::SessionSerializer>)> args
=> {"id"=>27,
 "email"=>"user12.name@email.com",
 "created_at"=>"2000-01-01T05:00:03.000Z",
 "updated_at"=>"2000-01-01T05:00:19.000Z",
 "client_id"=>18,
 "role"=>"coordinator",
 "authentication_token"=>"pxcFTApxeyCo1evMNd9X",
 "first_name"=>"User12",
 "last_name"=>"Name",
 "analytics_enabled"=>false}
```

[...]

PR #5853, opened 2026-06-25
Allow early configuration of devise models

kaisq:
As of Rails 8, which enables lazy-loaded routes by default, Devise configuration particularly in test environments consistently breaks because tests begin running before routes have been loaded. This causes Devise (and therefore Warden) to operate with only partial configuration, particularly, a stale default_scope among other things.

This change includes a new method to create a skeleton Mapping object which can be used to indicate a Devise-enabled model class that will be used for authentication before the route loading step has been completed. This object will have its configuration finalized once routes have been completed, and the Devise routes for that object are properly drawn.

This is backwards compatible with the current route-centric configuration.


kaisq:
This would fix #5752 (I have tested this locally with code that I believe to be hitting the same error) and likely also fixes #5844 (though I do not currently have a broken example of this)

clarissalimab:
Thank you for working on this!

I've tested this solution against a mini app that reproduces a scenario I faced with a project where we built a custom Warden strategy for a header-based authentication: https://github.com/thoughtbot/devise-eager-load-issue-demo (see [this branch]( https://github.com/thoughtbot/devise-eager-load-issue-demo/tree/testing-devise-pr-5853)).

It works well, but without the Devise model mapping configuration requirement that this introduces, it would introduce a regression on the next version since all requests would fail to authenticate going forward, which we may be hesitant to add.

I wonder if there's a way to have this solution where people can opt-in to use it without introducing the regression. Do you have any ideas? I'd be happy to help digging.

kaisq:
@clarissalimab thanks for testing this out, that's a great point. I've fixed that by updating the `configure_warden!` phase to be called both after the initial registration and after route loading. Would you mind testing again?

For the maintainers: to clarify my goals here, what I mean by "backwards compatible" is that it should be possible to update to a version with this change without making any configuration changes at all. I'm currently achieving this using a "last config wins" strategy if non-route configurations are declared in both locations, but there are a few other options to do this more intelligently and I'm happy to discuss that if we deem this a strategy worth pursuing. The "early config" stage should be entirely optional for projects where the existing route-based configuration is still working.

yourfinancedomain-star:
Re: Identity verified and then I can get more info on this platform so I can use it for enterprise workflows and the 3rd edition of th'e business

PR #5855, opened 2026-07-21
Add absolute (max) session timeout to Timeoutable

codepawpaw:
Timeoutable only supported an inactivity timeout via config.timeout_in. Some applications need to log a user out after a fixed session length for security reasons, regardless of activity.

timeout_in now also accepts a Hash to opt into an absolute session timeout alongside the existing inactivity timeout:

  config.timeout_in = { inactivity: 30.minutes, max: 8.hours }

Passing an Integer/String/duration keeps the current inactivity only behavior, so this is fully backward compatible. The session start time is stored in the warden session and the user is signed out once the maximum length is reached.


codepawpaw:
Issue: https://github.com/heartcombo/devise/issues/5753

